Issue #195, opened 2023-03-01
Roadmap, short- to medium-term


jherland:
A high-level collection of tasks/issues that we should be able to tackle fairly soon, and that would significantly improve the user experience of FawltyDeps:

### Improving our matching of dependencies vs imports

- [ ] Finding and fixing various corner cases where our mapping from dependency names to import names using the local environment is incorrect.
  - Related issues: #142, #173, #176, #187
- [x] Teach FawltyDeps to look at a different environment than what it itself is running in. This would allow you to e.g. `pipx install fawltydeps` to have it available across all your projects instead of having to install it in every project's own virtualenv.
  - Related issues: #161
- [ ] Look at fully decoupling ourselves from the local environment: pull information about packages and their import names directly from PyPI? Maybe include a hardcoded registry of common cases inside FawltyDeps itself?
    - Related issues: #32
    - See also the [design document](https://github.com/tweag/FawltyDeps/blob/main/docs/DesignDoc.md#4-proposed-solution)

### Parsing imports and dependencies

- [ ] A generic and easy-to-use directive for _ignoring_ files.
  - [x] Along the same lines as when we added the `basepath` positional argument (which is a shortcut for passing `--code` and `--deps`) we should have a `--ignore`(?) option for easily ignoring parts of the directory tree, both with respect to parsing code and deps.
  - [ ] Do we want to ignore some things by default? `build/`? `dist/`? virtualenvs? `__pycache__`? Maybe safer to reuse stuff from `.gitignore`?
  - Related issue: #226
- [ ] Improve our imports parsing:
  - [ ] Try to recognize and parse some common uses of `importlib` for importing modules.
    - Related issues: #46
  - [ ] At least warn the user when we see that a dynamic import is likely happening, but we're unable to fully parse it.
  - [ ] Improve our detection of `sys.path` and similar environment manipulation, and warn the user of how this likely affects our classification of 1st- vs 3rd-party packages.
  - [ ] Add knobs to allow the user to tell us where more 1st-party packages can be found.
  - [ ] Recognize the _context_ in which something is imported, and how that should affect our reporting (see issue #171 for inspiration). Recognize e.g. the following patterns:
    -  `try: import ... except ImportError: ...`
    -  `if sys.version_info >= ...: import ... else: import ...` 
  - [ ] Deal with various corner cases of parsing notebooks, e.g. #132
- [ ] Improve our dependency parsing:
  - Are there more dependency formats that we need to support?
    - What about Conda?
[...]

Issue #187, opened 2023-02-24
Packages that are installed from local (editable?) folders are not reported by `packages_distributions()`?


jherland:
This is visible inside the FawltyDeps dev environment: `packages_distributions()` does not return `fawltydeps` among its import names or packages:
```sh
python -c "from importlib_metadata import packages_distributions; print(packages_distributions())" | grep fawl
# returns nothing.
```

Also very common in monorepo setups:
```toml
[tool.poetry.dev-dependencies]
pytest = "^7.1.2"
xxx-geometry = {path = "../geometry", develop = false}
```
This installs `xxx-geometry` from elsewhere in the monorepo (by relative path), and `packages_distributions()` fail to report this as well (even if the `develop = false` indicates that the install is not editable, it is still certainly _local_).

Here is another example that installs fawltydeps by local path a new venv:

First in a non-editable manner, in which `packages_distributions()` _is_ able to find it:
```sh
❯ python3 -m venv foobar
❯ foobar/bin/pip install .  # install by relative path, but non-editable
Processing /home/jherland/code/fawltydeps
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting importlib_metadata<6.0.0,>=5.0.0
  Using cached importlib_metadata-5.2.0-py3-none-any.whl (21 kB)
Collecting tomli<3.0.0,>=2.0.1
  Using cached tomli-2.0.1-py3-none-any.whl (12 kB)
Collecting isort<6.0,>=5.10
  Using cached isort-5.12.0-py3-none-any.whl (91 kB)
Collecting pydantic<2.0.0,>=1.10.4
  Using cached pydantic-1.10.5-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (3.1 MB)
Collecting zipp>=0.5
  Using cached zipp-3.14.0-py3-none-any.whl (6.7 kB)
Collecting typing-extensions>=4.2.0
  Using cached typing_extensions-4.5.0-py3-none-any.whl (27 kB)
Building wheels for collected packages: fawltydeps
  Building wheel for fawltydeps (pyproject.toml) ... done
  Created wheel for fawltydeps: filename=fawltydeps-0.4.0-py3-none-any.whl size=28588 sha256=78ccc5870a632b7e299b0fbccef7f1550ec4bd6b1bf202b11c4ca9dbc983df67
[...]

Issue #161, opened 2023-02-17
Add the possibility to point towards the path of the environment to use for inferring package distribution information.


mknorps:
To be able to test the behaviour of FawltyDeps on real projects and to see the improvement, we need to have a way to define and use an environment different that the FD test environment.

This issue was mentioned by @jherland, @Nour-Mws  and @mknorps on many occasions:

- in `tests_map_dep_name_to_import_names.py`:
    ```
    # TODO: These tests are not fully isolated, i.e. they do not control the
    # virtualenv in which they run. For now, we assume that we are running in an
    # environment where at least these packages are available:
    # - setuptools (exposes multiple import names, including pkg_resources)
    # - pip (exposes a single import name: pip)
    # - isort (exposes no top_level.txt, but 'isort' import name can be inferred)
    ```
- in #150 
- in slack communication:
  @Nour-Mws 
  > Thinking out loud: with our current local-dep-to-import mapping, even when #150 is merged, it will still be impossible to run fawltydeps in the real projects tests, without installing all the deps defined by a project in the env in which fawltydeps is running?
  > In other terms, if we want to separate project x run from project y run, we can't use the current test setup, as they all currently run in the same environment.
  > Another thing to note is that since fawltydeps needs to be in the env of the project that it tests, it may be the case that some of the project's undeclared deps end up installed by fawltydeps, and  so while for a project declaring pyml, you might have ended with: "I couldn't find pyml in the env, mapping to pyml", you actually find the given package in the env.
  > Thoughts?
  
  @jherland 
  > True. Also PR #147 did change real_projects to run fawlytdeps in a subprocess, and with that we could run it inside a different environment, but we still have to add the functionality to setup those environments and then make fawltydeps run inside them.
  > In short, I think there are three different scenarios we could look at here:
  > A. FD can only query packages inside the environment it is running.  This is where we are, and we really haven't planned on changing that all, yet. (The real_projects work would "simply" be about installing and running FD inside the various test environments.)
  > B. Make FD somehow able to switch its execution context from one environment to another. I think this is impossible in general (without switching processes), but it might be possible to e.g. call importlib_metadata.packages_distributions() in a subprocess (than runs inside the target environment). However, this still relies on importlib_metadata being installed in that environment.
   >  C. It might be possible to configure/customize importlib_metadata to examine a different environment than the current environment. Worst case, we could even write our own version of it. I have not looked deeply into this yet.

Nour-Mws:
Closed via #219.

Issue #142, opened 2023-02-14
Dependency to import name(s) mapping from local environment: a roadmap


Nour-Mws:
### Dependency to import mapping
- [x] Python stdlib machinery to map a package name to its import name(s) using distribution information of packages installed in a given environment (covered by #139).
- [x] #144
- [ ] #161

### Comparison module 
- [x] Agree on what to do when packages corresponding to parsed dependencies are not found in the environment (we're not contacting PyPI for the immediate next release): INTERTWINED WITH 2ND
     - use a fallback strategy (e.g. to identity mapping)?
     - declare dependencies as missing and tell the user they were not found and they have to be installed?
     - other?
-  ~~Keep the possibility to do a simple identity mapping in the comparison module?~~

### CLI
- ~~Related to the last point above, if identity mapping is not completely replaced, we should add the option to choose the type of mapping and default to dep->import mapping from local installation.~~
- ~~Add optional arg to point towards the environment path (if not defined, what to default to?)~~

PR #215, opened 2023-03-13
Teach `packages.LocalPackageLookup` to look at other venvs

jherland:
This is the most basic building block towards decoupling the venv that FawltyDeps is installed in, from the venv that FawltyDeps is run against. From this we can work towards several goals in future PRs:
- No longer require `pip install fawltydeps` into your project's venv, but instead e.g. `pipx install fawltydeps` to make FD available to all your projects.
- Pass `--venv` as a command line option
- Passing more than one venv
- Auto-detecting and adding venvs under `basepath` (i.e. make `--venv` somewhat similar to `--code` and `--deps`)
- Supporting a first approximation to using PyPI, by downloading/installing packages from PyPI into a temporary venv, and then pointing FD at that venv.
- Changing our `real_projects` tests to no longer have to install/uninstall FD in the experiment's venv.

This PR is directly relevant to #161, and as such is also related to task 2 in the roadmap: #195. As far as #142 is still relevant, I suspect this might go a way towards fixing that as well. Furthermore, revisiting https://github.com/tweag/FawltyDeps/issues/187 is probably warranted after this PR is merged.

Commits:
- `pyproject.toml`: Use latest `importlib_metadata` across all Python versions
- Use `importlib_metadata` instead of `importlib.metadata` everywhere
- `packages`: Teach `LocalPackageLookup` to start looking at other venvs
- `test_real_projects`: Verify that all requirements are present
- `packages.LocalPackageLookup`: Remove redundant sanity check

### DIFF
--- a/fawltydeps/main.py
+++ b/fawltydeps/main.py
@@ -20,6 +20,7 @@ from operator import attrgetter
 from pathlib import Path
 from typing import Dict, List, Optional, TextIO, no_type_check
 
+import importlib_metadata
 from pydantic.json import custom_pydantic_encoder  # pylint: disable=no-name-in-module
 
 from fawltydeps import extract_imports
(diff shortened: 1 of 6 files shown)


Nour-Mws (approved):
Looks good. I haven't found anything to complain about :)

PR #219, opened 2023-03-14
Add `--venv` option to CLI

jherland:
Depends on #215.

- `packages`: Add `venv_path` argument to `resolve_dependencies()`
- Add `venv` setting and `--venv` command-line option
- `test_real_projects`: Use `--venv` instead of editable install of FD

### DIFF
--- a/fawltydeps/settings.py
+++ b/fawltydeps/settings.py
@@ -120,6 +120,7 @@ class Settings(BaseSettings):  # type: ignore
     actions: Set[Action] = {Action.REPORT_UNDECLARED, Action.REPORT_UNUSED}
     code: Set[PathOrSpecial] = {Path(".")}
     deps: Set[Path] = {Path(".")}
+    venv: Optional[Path] = None
     output_format: OutputFormat = OutputFormat.HUMAN_SUMMARY
     ignore_undeclared: Set[str] = set()
     ignore_unused: Set[str] = set()
@@ -323,6 +324,16 @@ def populate_parser_options(parser: argparse._ActionsContainer) -> None:
         help=(
             "Where to find dependency declarations (file or directory, defaults"
             " to looking for supported files in the current directory)"
+        ),
+    )
+    parser.add_argument(
+        "--venv",
+        type=Path,
+        metavar="VENV_DIR",
+        help=(
+            "Where to find a virtualenv that has the project dependencies"
+            " installed, defaults to the Python environment where FawltyDeps is"
+            " installed."
         ),
     )
     parser.add_argument(
--- a/noxfile.py
+++ b/noxfile.py
@@ -81,7 +81,7 @@ def lint(session):
     session.run("pylint", "fawltydeps")
     session.run(
         "pylint",
-        "--disable=missing-function-docstring,invalid-name,redefined-outer-name",
+        "--disable=missing-function-docstring,invalid-name,redefined-outer-name,too-many-lines",
         "tests",
     )
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,10 @@
 """Fixtures for tests"""
+import sys
+import venv
 from pathlib import Path
+from tempfile import mkdtemp
 from textwrap import dedent
-from typing import Dict, Iterable, Union
+from typing import Dict, Iterable, Set, Union
 
 import pytest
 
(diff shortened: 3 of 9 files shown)


noxfile.py
```diff
@@ -81,7 +81,7 @@ def lint(session):
     session.run("pylint", "fawltydeps")
     session.run(
         "pylint",
-        "--disable=missing-function-docstring,invalid-name,redefined-outer-name",
+        "--disable=missing-function-docstring,invalid-name,redefined-outer-name,too-many-lines",
```
Nour-Mws, line 84: What's causing this?

fawltydeps/settings.py
```diff
@@ -325,6 +326,16 @@ def populate_parser_options(parser: argparse._ActionsContainer) -> None:
             " to looking for supported files in the current directory)"
         ),
     )
+    parser.add_argument(
+        "--venv",
+        type=Path,
+        metavar="VENV_DIR",
+        help=(
+            "Where to find a virtualenv that has the project dependencies"
```
Nour-Mws, line 334: "where to find" makes it seem like we will go looking for the venv while the reality is that we expect the user to provide the path.
```suggestion
            "Path to a virtualenv that has the project dependencies"
```

tests/test_real_projects.py
```diff
@@ -53,11 +52,11 @@ def verify_requirements(venv_path: Path, requirements: List[str]) -> None:
 def run_fawltydeps_json(
     *args: str, venv_dir: Optional[Path], cwd: Optional[Path] = None
 ) -> JsonData:
-    cmd = ["fawltydeps"]
-    if venv_dir:
-        cmd = [f"{venv_dir}/bin/fawltydeps"]
+    argv = ["fawltydeps", "--config-file=/dev/null", "--json"]
```
Nour-Mws, line 55: Is this supposed to be `args`?

noxfile.py
jherland: `test_cmdline.py` has grown past 1000 lines 😱

Related to what I mentioned in today's meeting. Also excited to see what @mknorps comes up with in her test PR.

tests/conftest.py
```diff
@@ -21,6 +24,34 @@ def _inner(file_contents: Dict[str, str]) -> Path:
...
+            # into believing these are genuine packages
+            dist_info_dir = site_dir / f"{package_name}-1.2.3.dist-info"
+            dist_info_dir.mkdir()
+            (dist_info_dir / "METADATA").write_text(
+                f"Name: {package_name}\nVersion: 1.2.3\n"
+            )
+            top_level = dist_info_dir / "top_level.txt"
+            top_level.write_text("".join(f"{name}\n" for name in sorted(import_names)))
+            for name in import_names:
+                (site_dir / f"{name}.py").touch()
```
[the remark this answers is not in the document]
jherland: So, I'm not sure if there's an "official" definition of what constitutes a minimal Python package (and looking at wheels vs. eggs vs. source archives would certainly complicate that definition).

However after digging around inside `importlib_metadata` for a few days now, _this_ is what I found to be the minimal needed, specifically for `importlib_metadata`, yes:
- The `.dist-info` subdir is how wheels are installed, so I chose that as the more relevant thing to fake (as opposed to the deprecated-but-still-in-use `.egg-info` format).
- The `METADATA` files has a _lot_ of fields for most packages, but the `Name` is what `importlib_metadata` uses to identify the package name.
- I'm actually not strictly sure that the `Version` is necessary, but it's also part of the `*.dist-info` subdir name, so I thought it best to not rock the boat too much.
- The `top_level.txt` is the _preferred_ way to enumerate what import names a package provides.
- The `.py` files are dummy modules, but in real life they are most often _packages_ instead, i.e. subdirs with `__init__.py` files and more. _Technically_, with the current version of `importlib_metadata` I don't think these are necessary, since the `top_level.txt` is followed blindly. But my upstream PR at https://github.com/python/importlib_metadata/pull/437 might end up checking that these packages/modules actually exist (to work around bugs in `_top_level_inferred()`).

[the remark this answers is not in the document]
Nour-Mws: Thanks. This makes sense now.

tests/test_real_projects.py
Nour-Mws: It's fine either way so feel free to do either and resolve. It's the first time that I see `argv` so I'll let you judge whether that's known in the Python community.

PR #223, opened 2023-03-14
Explain the origin of the tool's name in the README

Nour-Mws:


README.md
```diff
@@ -2,9 +2,7 @@
 
 # FawltyDeps
 
-A dependency checker for Python.
-
-Find _undeclared_ and/or _unused_ 3rd-party dependencies in your Python project.
+FawltyDeps is a dependency checker for Python that finds _undeclared_ and/or _unused_ 3rd-party dependencies in your Python project. The name is inspired by the Monty Python's [Fawlty Towers](https://en.wikipedia.org/wiki/Fawlty_Towers) sitcom.
```
jherland, line 5: But Fawlty Towers was _not_ technically a Monty Python project, rather it was created by one of its members (and others) as a separate project: https://en.wikipedia.org/wiki/Fawlty_Towers#Origins

```suggestion
FawltyDeps is a dependency checker for Python that finds _undeclared_ and/or
_unused_ 3rd-party dependencies in your Python project.
The name is inspired by the Monty Python-adjacent
[Fawlty Towers](https://en.wikipedia.org/wiki/Fawlty_Towers) sitcom.
```

jherland (changes requested):
Agree to mention this in the README, but let's get the facts straight 😅

jherland (approved):

PR #224, opened 2023-03-14
main: Write a friendly success message when we find no issues

### LINKED ISSUE #222: Success message

When/after the project has no problems FawltyDeps detects, the output is:

```
$ poetry run fawltydeps        

```

This is slightly confusing as I would expect _either_:
* A hardcore behavior of only setting the exit code and not printing anything (cf. `rm important_data` or `grep hello /dev/null`)
* A status message like "No unused or undeclared dependencies detected."

When there are no problems, output "No unused or undeclared dependencies detected." before suggesting `--detailed`.

jherland:
When checking a project in the (default) summary mode, and no problems
were found, we would still print the following, confusing message to our
users:

Only print that message when there is indeed a preceding report.

Otherwise, when we're checking for undeclared/unused, and no problems
were found, print this friendlier message instead:

  No unused or undeclared dependencies detected.

Suggested by @koterpillar in #222.
This fixes #222.

### DIFF
--- a/fawltydeps/main.py
+++ b/fawltydeps/main.py
@@ -44,6 +44,7 @@ from fawltydeps.types import (
 
 logger = logging.getLogger(__name__)
 
+SUCCESS_MESSAGE = "No unused or undeclared dependencies detected."
 VERBOSE_PROMPT = "For a more verbose report re-run with the `--detailed` option."
 UNUSED_DEPS_OUTPUT_PREFIX = "These dependencies appear to be unused (i.e. not imported)"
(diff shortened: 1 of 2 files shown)


koterpillar:
Does this report appropriately when only unused or only undeclared dependencies are being checked?

koterpillar:
Sorry for my unclear question. What I meant is, the message should not mention _undeclared_ dependencies when checking for _unused_, and vice versa. Do you want a separate bug for that?

PR #227, opened 2023-03-14
Leverage existing requirements.txt parsing logic to close #225

### LINKED ISSUE #225: Line continuations aren't fully supported for per-requirement options in requirements.txt
**Describe the bug**
Parsing of requirements.txt appears to not handle line continuations.

**To Reproduce**
Create a `requirements.txt` file with these contents:

```
black==23.1.0 \
    --hash=sha256:0052dba51dec07ed029ed61b18183942043e00008ec65d5028814afaab9a22fd
```

Run fawltydeps with this switch: `--check-unused`

Observe the output:
```
WARNING:fawltydeps.extract_declared_dependencies:Could not parse requirements.txt line 'black==23.1.0 \\': not enough values to unpack (expected 1, got 0)
```

**Expected behavior**
Correct parsing of above `requirements.txt`.

**Environment**
 - OS name + version: macOS 13.1
 - Version of the code: 0.6.1

**Additional context**
None

vreuter:
Once https://github.com/nexB/pip-requirements-parser/pull/17 is merged, we could more directly use `pip-requirements-parser` (namely, the `RequirementsParser.from_string` method). Until then, this works around a `NameError` that comes up when using the less common `from_string` method in the `RequirementsFile` class in that project.

### DIFF
--- a/fawltydeps/extract_declared_dependencies.py
+++ b/fawltydeps/extract_declared_dependencies.py
@@ -6,12 +6,11 @@ import logging
 import re
 import sys
 from dataclasses import replace
-from functools import partial
-from itertools import takewhile
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Callable, Iterable, Iterator, NamedTuple, Optional, Set, Tuple
 
+from pip_requirements_parser import RequirementsFile  # type: ignore
 from pkg_resources import Requirement
 
 from fawltydeps.limited_eval import CannotResolve, VariableTracker
@@ -32,13 +31,6 @@ else:
 logger = logging.getLogger(__name__)
 
 ERROR_MESSAGE_TEMPLATE = "Failed to %s %s %s dependencies in %s: %s"
-# https://pip.pypa.io/en/stable/reference/requirements-file-format/#per-requirement-options
-PER_REQUIREMENT_OPTIONS = [
-    "--install-option",
-    "--global-option",
-    "--config-setting",
-    "--hash",
-]
 
 NamedLocations = Iterator[Tuple[str, Location]]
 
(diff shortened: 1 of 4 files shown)


jherland:
Hmm. I took a closer look at  `pip-requirements-parser`'s own code, and AFAICS it boils down to a single Python script of almost 3000 lines (😱), the vast majority of which is copied directly from `pip` itself. After reading through the project's own `README`, my initial WTF!? reaction is somewhat softened, and the project argues well for why things are done this way.

Still, what I don't like about this integration here is having to write out a temporary file to parse some requirements text, when this text (with the sole exception of parsing `setup.cfg`) _already_ comes directly from a file!

However, fixing this requires firrst some moderate refactoring in our `extract_declared_dependencies.py` to adjust our various parsers to work on `Path`s directly rather than `str` + `Location` arguments.

Coincidentally, this also frees up our parser to decide for themselves exactly how to read the given file (i.e. they may choose something more appropriate than `Path.read_text()`, depending on their context. I think this refactoring might be a net win overall, and worth doing in any case...

After this refactoring `parse_requirements_contents()` would be reduced to a single call to `RequirementsFile.from_file(filename=str(path), include_nested=False).requirements`.

Then there is the special case of `parse_setup_cfg_contents()` which also leans on `parse_requirements_contents()` to parse its requirements values, but given that `requirements.txt` probably outnumber `setup.cfg` files 1000:1, I'd rather put the clumsy workaround (writing a temp file) inside `parse_setup_cfg_contents()` rather than inside `parse_requirements_contents()`.

I started looking at this refactoring, and I'd rather you rebuild this PR on top of that, instead of having to write temporary files for every single `requirements.txt` file we will ever parse... Will post a branch as soon as I got something that passes tests...

jherland:
> I started looking at this refactoring, and I'd rather you rebuild this PR on top of that, instead of having to write temporary files for every single `requirements.txt` file we will ever parse... Will post a branch as soon as I got something that passes tests...

https://github.com/tweag/FawltyDeps/pull/229

jherland (approved):
Looks great! Two things missing:

- An added test (demonstrating #225) that is fixed by this PR.
- A comment/consideration on how to possibly make the `setup.cfg` situation better.

For the second one, I've been looking at pip-requirements-parser to see if there is _anything_ in there we could benefit from[^1].

[^1]: Even if we were to call into its internal functions, this could still be considered acceptable as we can _pin_ our use of pip-requirements-parser (and control when we upgrade it) as opposed to `pip` itself where we are only given whatever is available on the user's machine.

AFAICS, we have to dig all the way in to `RequirementsFileParser._parse_file()` before we've actually opened and read the file. so a cleaner implementation of `RequirementsFile.from_string()` would have to copy all but the first line of `._parse_file()`...

vreuter:
> Looks great! Two things missing:
> 
>     * An added test (demonstrating #225) that is fixed by this PR.
> 
>     * A comment/consideration on how to possibly make the `setup.cfg` situation better.
> 

The test I add here: 0dac8e8a262ff341b7142d014d2a719e5a26ef04
And the comment I add here: dab88d4fca0c647974b9a1722d7bed2bd586520b

PR #229, opened 2023-03-14
Refactor `extract_declared_dependencies` parser APIs in preparation for #227

jherland:
This should enable @vreuter's rewrite of our requirements.txt parser to avoid creating a temporary copy of each requirements.txt file while parsing it.

We do however need to find a better way to parse our setup.cfg files...

- `test_extract_declared_dependencies_succes`: Harmless test refactoring
- `extract_declared_dependencies`: Refactor API of parser functions
- `test_extract_declared_dependencies_pyproject_toml`: Remove unnecessary `dedent()`
- `extract_declared_dependencies`: Rename parser functions

### DIFF
--- a/fawltydeps/extract_declared_dependencies.py
+++ b/fawltydeps/extract_declared_dependencies.py
@@ -5,9 +5,11 @@ import configparser
 import logging
 import re
 import sys
+from dataclasses import replace
 from functools import partial
 from itertools import takewhile
 from pathlib import Path
+from tempfile import NamedTemporaryFile
 from typing import Callable, Iterable, Iterator, NamedTuple, Optional, Set, Tuple
 
 from pkg_resources import Requirement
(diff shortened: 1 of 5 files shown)


fawltydeps/extract_declared_dependencies.py
```diff
@@ -5,9 +5,11 @@
 import logging
 import re
 import sys
+from dataclasses import replace
```
vreuter, line 8: Nit:

When 2 conditions are true, I especially favor just importing the library and then having use sites be qualified. Those would be...
a) The imported name is a relatively common word / name (`replace`)
b) It's not as typical / standard (i.e., `from functools import partial` violates a, but I'd argue that since it's been around quite some time it's more natural / common.

tests/test_extract_declared_dependencies_pyproject_toml.py
```diff
@@ -176,118 +158,104 @@
...
+    write_tmp_files, pyproject_toml, expected_deps
 ):
-    source = Location(Path("pyproject.toml"))
-    result = list(parse_pyproject_contents(pyproject_toml, source))
-    expected = [DeclaredDependency(dep, source) for dep in expected_deps]
+    tmp_path = write_tmp_files({"pyproject.toml": pyproject_toml})
+    path = tmp_path / "pyproject.toml"
+
+    result = list(parse_pyproject_toml(path))
+    expected = [DeclaredDependency(dep, Location(path)) for dep in expected_deps]
```
vreuter, line 168: The lists are compared for full equivalence rather than just contents; why does order matter here?

vreuter (approved):

fawltydeps/extract_declared_dependencies.py
jherland: This makes sense, and I would say that my threshold is only slightly lower:
- When you only use one (or very few) names from a module, it is nice to communicate this up front in the import statement.
- Even if the name is a common word, if it is _only_ used as that imported name across the whole file, I don't have a problem with it.
- But if the name introduces _any_ kind of (perceived) ambiguity, e.g. you have a class with a `replace()` method in the same file, then I would definitely argue for using it with qualification everywhere.

In summary (and in this case): grepping for `replace` in the file yields only relevant hits, so I find any extra qualification unnecessary.

PR #230, opened 2023-03-15
Adapt success message to the performed checks

Nour-Mws:
A follow up to #224.

A function in the Analysis class now generates the success message.
This function is static and does not directly rely on the checks (actions) currently activated in an Analysis instance.
I opted for this to avoid creating an Analysis instance in tests just to generate the success message.

### DIFF
--- a/fawltydeps/main.py
+++ b/fawltydeps/main.py
@@ -44,7 +44,6 @@ from fawltydeps.types import (
 
 logger = logging.getLogger(__name__)
 
-SUCCESS_MESSAGE = "No unused or undeclared dependencies detected."
 VERBOSE_PROMPT = "For a more verbose report re-run with the `--detailed` option."
 UNUSED_DEPS_OUTPUT_PREFIX = "These dependencies appear to be unused (i.e. not imported)"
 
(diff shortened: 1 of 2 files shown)


fawltydeps/main.py
```diff
@@ -76,6 +75,18 @@ def is_enabled(self, *args: Action) -> bool:
...
+        """Returns the message to print when the analysis finds no errors."""
+        checking = []
+        if check_undeclared:
+            checking.append("undeclared")
+        if check_unused:
+            checking.append("unused")
+        if checking:
+            return f"No {' or '.join(checking)} dependencies detected."
+        return ""
+
```
jherland, line 89: Nit: Does it make sense to return `Optional[str]` and `return None` in the fallthrough case, to more clearly identify the no-message case?

mknorps: Nit is is, but I agree :)

Nour-Mws: Done in [3ea141c](https://github.com/tweag/FawltyDeps/pull/230/commits/3ea141cb794a71c33d1cc205e1c100255c8d8e91).

Issue #198, opened 2023-03-02
Discussion: possibility to add multiple positional arguments to CLI


mknorps:
Started by @mknorps in a discussion: https://github.com/tweag/FawltyDeps/pull/194#discussion_r1122082040
```
    parser.add_argument(
        "basepath",
        type=lambda p: None if p == argparse.SUPPRESS else Path(p),
        nargs="?",
        help="(Optional) directory in which to search for code (imports) "
        "and/or dependency declarations",
    )
```
> [discussion] Checking options, '*' also seems interesting to use here:
> '*'. All command-line arguments present are gathered into a list.
> 
>  Then we could add multiple folders to search in a single command. The downside is that it requires agreeing on a default behaviour of such combination.
> 
>  I checked how it works in case of `ls` and we have following scenarios:
>  - `ls` will list content of the current folder
>  - `ls dir` will list contents of the `dir`
>  - `ls dir1 dir2` will give sth like:
>      ```
>         dir1:
>         file1 file2
> 
>         dir2:
>         file3 file4
>      ```
> 
> FD could use the above, or more intuitively
> ```
> fawltydeps dir1 dir2
> ```
> will run FD on those folders combined.

This issue is to consider how the positional arguments should behave, keeping in mind the user who is used to some other command-line tools' behaviors.
This should be discussed jointly with #156.

PR #238, opened 2023-03-15
Multiple positional args for base paths

vreuter:
This aims to address #198

### DIFF
--- a/fawltydeps/settings.py
+++ b/fawltydeps/settings.py
@@ -193,13 +193,16 @@ class Settings(BaseSettings):  # type: ignore
 
         args_dict = cmdline_args.__dict__
 
-        base_path = getattr(cmdline_args, "basepath", None)
-        if base_path is not None:
-            code_path = args_dict.setdefault("code", {base_path})
-            deps_paths = args_dict.setdefault("deps", {base_path})
-            paths = [base_path, *code_path, *deps_paths]
-            if len(set(paths)) == len(paths):
-                msg = f"Each path option has a different value: {paths}"
+        base_paths = set(getattr(cmdline_args, "basepaths", []))
+        if base_paths:
+            code_paths = args_dict.setdefault("code", base_paths)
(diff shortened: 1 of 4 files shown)


tests/test_cmdline.py
```diff
@@ -946,6 +946,35 @@ def test_cmdline_args_in_combination_with_config_file(
...
+    next(obs_lines_relevant)  # discard
+    unused_deps = [UnusedDependency(name, [Location(deps_path)]) for name in uniq_deps]
+    exp_lines = [
+        line
+        for dep in unused_deps
+        for line in f"- {dep.render(include_references=True)}".split("\n")
+    ]
+    assert list(obs_lines_relevant) == exp_lines
+
+
```
jherland, line 977: NIt: Why move these tests to the other side of `pyproject_toml_contents()`? Personally, I have a slight preference for defining the callee (`pyproject_toml_contents()`) above the callers (the tests).

fawltydeps/settings.py
```diff
@@ -193,13 +193,16 @@ def create(cls, cmdline_args: argparse.Namespace) -> "Settings":
...
 
-        base_path = getattr(cmdline_args, "basepath", None)
-        if base_path is not None:
-            code_path = args_dict.setdefault("code", {base_path})
-            deps_paths = args_dict.setdefault("deps", {base_path})
-            paths = [base_path, *code_path, *deps_paths]
-            if len(set(paths)) == len(paths):
-                msg = f"Each path option has a different value: {paths}"
+        base_paths = set(getattr(cmdline_args, "basepaths", []))
+        if base_paths:
```
mknorps, line 197: That is actually the best conflict resolution 🙌

tests/test_cmdline.py
vreuter: No _strong_ preference, but in general I like the structure...
1. fixtures
2. tests
3. helpers

(disregarding imports / constants at the top)

PR #237, opened 2023-03-15
Various doc updates

jherland:
- `README.md`: Add documentation about `venv` config file option
- `README.md`: Reword the section on contributing to our real tests
- `docs/CodeDesign.md`: Rewrite/update pass

PR #242, opened 2023-03-16
Rename `--venv` to `--pyenv`

### LINKED ISSUE #236: Different name for pypi package and exposed package
**Describe the bug**
Some packages have the same name, but are imported by a different name. For example python-dotenv

image

**To Reproduce**
Install python-dotenv and add the following example.

```py
from dotenv import load_dotenv

load_dotenv()  # take environment variables from .env.
```

**Expected behavior**
As a result, we don't see unused and undeclared dependencies

**Environment**
 - OS name + version: NixOS 23.05.20230301.a08d697 (Stoat) x86_64
 - Version of the code: 0201e74a158c5774db9150e1e70f01e76bad34e7

jherland:
This fixes #236, and recognizes that `--venv` is a poor option name when we also recognize more than just virtualenvs. Hence the rename to `--pyenv`.

Commits:
- `packages.LocalPackageLookup`: Accept more Python environments types
- Rename `--venv` option to `--pyenv`


README.md
```diff
@@ -231,6 +232,9 @@ Here is a complete list of configuration directives we support:
   Defaults to the current directory, i.e. like `code = ["."]`.
 - `deps`: Files or directories containing the declared dependencies.
   Defaults to the current directory, i.e. like `deps = ["."]`.
+- `pyenv`: The path to the Python environment to use for resolving project
+  dependencies into provided import names. Defaults to the Python environment
+  where FawltyDeps is installed).
```
jherland, line 237: Found by @vreuter in #237:

```suggestion
- `pyenv`: The path to the Python environment to use for resolving project
  dependencies into provided import names. Defaults to the Python environment
  where FawltyDeps is installed.
```

tests/test_local_env.py
```diff
@@ -1,13 +1,56 @@
...
+]
+
+
+@pytest.mark.parametrize(
+    "subdir", [pytest.param(d, id=f"venv:{d}") for d in env_subdirs]
+)
+def test_determine_package_dir__various_paths_in_venv(tmp_path, subdir):
+    venv.create(tmp_path, with_pip=False)
+    path = tmp_path / subdir
+    expect = tmp_path / f"lib/python{major}.{minor}/site-packages"
```
vreuter, line 33: Since it's used a few places, maybe a helper could be added that captures the logic of "give me the path to `site-packages` if i give you the base path, the major version, and the minor version"

vreuter (approved):
looks good, just a few suggestions

fawltydeps/settings.py
```diff
@@ -327,11 +327,11 @@ def populate_parser_options(parser: argparse._ActionsContainer) -> None:
         ),
     )
     parser.add_argument(
-        "--venv",
+        "--pyenv",
         type=Path,
-        metavar="VENV_DIR",
+        metavar="DIR",
```
Nour-Mws, line 332: Why `DIR` instead of `PYENV_DIR`?

fawltydeps/packages.py
```diff
@@ -102,46 +102,67 @@ def is_used(self, imported_names: Iterable[str]) -> bool:
...
+    @classmethod
+    def determine_package_dir(cls, path: Path) -> Optional[Path]:
+        """Return the site-packages directory corresponding to the given path.
+
+        The given 'path' is a user-provided directory path meant to point to
+        a Python environment (e.g. a virtualenv, a poetry2nix environment, or
+        similar). Deduce the appropriate site-packages directory from this path,
+        or return None if no environment could be found at the given path.
+        """
+        # We define an "valid Python environment" as a directory that contains
```
Nour-Mws, line 134:
```suggestion
        # We define a "valid Python environment" as a directory that contains
```

fawltydeps/settings.py
jherland: 🤷🏻‍♂️ I'm good with either option.

```suggestion
        metavar="PYENV_DIR",
```

Nour-Mws: Let's go with the `PYENV_DIR` then :D

Issue #211, opened 2023-03-08
Properly handling command-line options that take multiple values


jherland:
Tangentially related to [this slack thread](https://tweag.slack.com/archives/C04AAQP8C85/p1678287577829779), I'm noticing some weird (maybe not wrong) interaction with the command line interface when you want to pass multiple values. This applies to `--ignore-undeclared` and `--ignore-unused` (and soon `--code` and `--deps`, see #156 and #210). I'm gonna use `--ignore-unused` only in the following examples:

1. Not given: `fawltydeps --generate-toml-config` ==> `ignore_unused = []` ✅
2. Missing arg: `fawltydeps --generate-toml-config --ignore-unused` ==> argparse error ✅
3. Empty arg: `fawltydeps --generate-toml-config --ignore-unused ""` ==> `ignore_unused = ['']` acceptable corner case IMHO ✅
4. Empty arg 2: `fawltydeps --generate-toml-config --ignore-unused=` ==> `ignore_unused = ['']` acceptable corner case IMHO ✅
5. Single arg: `fawltydeps --generate-toml-config --ignore-unused foo` ==> `ignore_unused = ['foo']` ✅
6. Single arg 2: `fawltydeps --generate-toml-config --ignore-unused=foo` ==> `ignore_unused = ['foo']` ✅
7. Two args:  `fawltydeps --generate-toml-config --ignore-unused foo bar` ==> `ignore_unused = ['foo', 'bar']` ✅
8. Two args 2:  `fawltydeps --generate-toml-config --ignore-unused foo --ignore-unused bar` ==> `ignore_unused = ['bar']` (the second overrides the first) ❓
9. Two args 3:  `fawltydeps --generate-toml-config --ignore-unused=foo bar` ==> `ignore_unused = ['foo']` (the second arg is interpreted as basepath) ❗❓
10. Two args 4: `fawltydeps --generate-toml-config --ignore-unused=foo --ignore-unused=bar` ==> `ignore_unused = ['bar']` (the second overrides the first) ❓
11. Two args 5: `fawltydeps --generate-toml-config --ignore-unused=foo,bar` ==> `ignore_unused = ['foo,bar']` (interpreted as one arg with comma embedded, not really unexpected) ❓
12. Two args 6: `fawltydeps --generate-toml-config --ignore-unused="foo bar"` ==> `ignore_unused = ['foo bar']` (interpreted as one arg with space embedded, also not really unexpected) ❓

I'm sure there are more variations we could play around with...

[...]

PR #247, opened 2023-03-17
Collect multivalued CLI args rather than overwriting

vreuter:
This addresses cases 8 and 10 in #211


mknorps:
I agree that extension it the most intuitive behaviour here!

Could you add a test to the test suite (of the command line tool, I know it is long enough already 🙈 ) to document that?

tests/test_settings.py
```diff
@@ -126,6 +128,53 @@ def test_base_path_overrides_config_file_code_and_deps(
...
+    return ret2
+
+
+@pytest.mark.parametrize("optargs", multivalued_optargs_grid())
+def test_multivalued_options_are_aggregated_correctly(optargs):
+    settings = run_build_settings(optargs)
+    assert settings.code == to_path_set(CODE_VALUES)
+    assert settings.deps == to_path_set(DEPS_VALUES)
+    assert settings.ignore_undeclared == set(UNDECLARED_VALUES)
+    assert settings.ignore_unused == set(UNUSED_VALUES)
```
jherland: Can save a few lines by moving the `keyed_items` above the helper function.

Also, my brain is already too fried on a Friday afternoon to make sense of the logic in here. Can I ask for a docstring and some more descriptive variable names than `ret` and `ret2`? 😉 

```suggestion
OPTION_VALUES = {
    "code": ["a", "b", "c"],
    "deps": ["d", "e"],
    "ignore-undeclared": ["f", "g"],
    "ignore-unused": ["h", "i", "j"],
}


def multivalued_optargs_grid():
    """This could really use a docstring..."""
    def powerset(iterable):
        xs = list(iterable)
        return chain.from_iterable(combinations(xs, k) for k in range(len(xs) + 1))

    rev_keyed = {v: k for k, vs in OPTION_VALUES.items() for v in vs}
    ret = []  # 
    for items in OPTION_VALUES.values():
        split_2 = set(
            chain.from_iterable(
                [(x, y), (y, x)]
                for x, y in combinations(powerset(items), 2)
                if len(x + y) == len(items) and not set(x) & set(y) and (x and y)
            )
        )
        ret.append(split_2)
    ret = list(product(*ret))
    ret2 = []
    for param_grid in ret:
        xss = list(chain(*param_grid))
        random.shuffle(xss)
        optargs = list(chain(*[[f"--{rev_keyed[xs[0]]}"] + list(xs) for xs in xss]))
        ret2.append(optargs)
    return ret2


@pytest.mark.parametrize("optargs", multivalued_optargs_grid())
def test_multivalued_options_are_aggregated_correctly(optargs):
    settings = run_build_settings(optargs)
    assert settings.code == to_path_set(OPTION_VALUES["code"])
    assert settings.deps == to_path_set(OPTION_VALUES["deps"])
    assert settings.ignore_undeclared == set(OPTION_VALUES["ignore-undeclared"])
    assert settings.ignore_unused == set(OPTION_VALUES["ignore-unused"])
```

fawltydeps/settings.py
```diff
@@ -450,3 +468,12 @@ def print_toml_config(settings: Settings, out: TextIO) -> None:
...
     print("\n".join(lines), file=out)
+
+
+class ArgparseExtendAction(argparse.Action):
+    """Action for doing 'extend' pre-3.8"""
+
+    def __call__(self, parser, namespace, values, option_string=None):  # type: ignore
+        items = getattr(namespace, self.dest, [])
+        items.extend(values)
+        setattr(namespace, self.dest, items)
```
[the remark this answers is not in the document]
vreuter: Excellent question for which I should add a test!

fawltydeps/settings.py
```diff
@@ -206,7 +206,18 @@ def create(cls, cmdline_args: argparse.Namespace) -> "Settings":
...
+            try:
+                field = cls.__fields__[arg]
+            except KeyError:
+                continue
+            else:
+                ret[arg] = (
+                    set(value)
+                    if isinstance(value, list) and set == field.default
+                    else value
+                )
```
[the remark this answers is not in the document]
vreuter: Agree, I will attempt to make this change.

tests/test_settings.py
vreuter: Haha yeah absolutely, bad variable names and lack of comments by me. I pieced this together from tinkering around in a Jupyter REPL and neglected to do a pass back through for readability improvements. Will do.

vreuter: I saved some space here:
https://github.com/tweag/FawltyDeps/pull/247/commits/b4f3429e083e322707aec74c8c81072ebb181646
and eventually made some better names and added a docstring here:
https://github.com/tweag/FawltyDeps/pull/247/commits/1b3ad8ef8a60f693955f6497a966de6f22f2979d

fawltydeps/settings.py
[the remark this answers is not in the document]
vreuter: Union action is added here:
https://github.com/tweag/FawltyDeps/pull/247/commits/1b3ad8ef8a60f693955f6497a966de6f22f2979d

tests/test_settings.py
```diff
@@ -128,6 +130,62 @@ def test_base_path_overrides_config_file_code_and_deps(
...
+    parameter_grids = []
+    for items in OPTION_VALUES.values():
+        split_2 = set(
+            chain.from_iterable(
+                [(x, y), (y, x)]
+                for x, y in combinations(powerset_non_empties(items), 2)
+                if len(x + y) == len(items) and not set(x) & set(y)
+            )
+        )
+        parameter_grids.append(split_2)
```
jherland, line 159: If we look at the `split_2` set as all the pairs that show a different way to order + cover the chars in the list, then we can rephrase its construction as generating permutations of all the values, and then finding ways to slice those permutations into non-empty pairs. Consider this helper:

```
    parameter_grids = [
        set(chain.from_iterable(map(slice_in_two, permutations(items))))
        for items in OPTION_VALUES.values()
    ]
```
with this helper:
```python
    def slice_in_two(items: List[T]) -> Iterator[Tuple[List[T], List[T]]]:
        assert len(items) >= 2
        for i in range(1, len(items)):
            yield items[:i], items[i:]
```

or using comprehensions instead of `map` and `chain.from_iterable`:

```suggestion
    def slice_in_two(items: List[T]) -> Iterator[Tuple[List[T], List[T]]]:
        assert len(items) >= 2
        for i in range(1, len(items)):
            yield items[:i], items[i:]

    parameter_grids = [
        {
            pair
            for sequence in permutations(items)
            for pair in slice_in_two(sequence)
        }
        for items in OPTION_VALUES.values()
    ]
```

tests/test_settings.py
```diff
@@ -128,6 +130,62 @@ def test_base_path_overrides_config_file_code_and_deps(
...
+
+
+@pytest.mark.parametrize(
+    "optname",
+    [act.dest for act in build_parser()._actions if isinstance(act, ArgparseUnionAction)], # pylint: disable=protected-access
+)
+def test_multivalued_options_are_absent_if_not_provided_at_cli(optname):
+    parsed_cli = build_parser().parse_args([])
+    with pytest.raises(AttributeError):
+        getattr(parsed_cli, optname)
```
jherland: Love this, although I don't see why this need to be parametrized? The code we're _testing_ (`build_parser().parse_args([])`) is the same in all cases, we're just testing multiple (non-)attributes of the returned args. Suggestion:

```suggestion
def test_multivalued_options_are_absent_if_not_provided_at_cli():
    parsed_args = build_parser().parse_args([])
    suppressed_attrs = [
        act.dest
        for act in build_parser()._actions  # pylint: disable=protected-access
        if isinstance(act, ArgparseUnionAction)
    ]
    for attr in suppressed_attrs:
        assert not hasattr(parsed_args, attr)
```

Maybe even consider extending the test to also cover other `Settings` members that correspond directly to `Namespace` members returned from `parse_args()`. After all, the cascade of CLI -> environment -> config file -> hardcoded defaults _depend_ very much on CLI options _not_ being present in this `Namespace` when they're not passed at the command line.

jherland (changes requested):
Thanks for the improvements. With that I think I was able to get through the heavy part of `multivalued_optargs_grid()`. Most of my comments are notes I made to myself while trying to grok this. You can disregard those, unless they pinpoint where I've seriously misunderstood something 😅 . Other than that, there are only a few more things I think require some attention before this is ready to 🚀

fawltydeps/settings.py
```diff
@@ -450,3 +478,17 @@ def print_toml_config(settings: Settings, out: TextIO) -> None:
...
+class ArgparseUnionAction(argparse.Action):
+    """Action for doing taking union of multiple command-line arguments for a single option"""
+
+    def __call__(  # type: ignore
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: Sequence[Any],
+        option_string: Optional[str] = None,
+    ):
```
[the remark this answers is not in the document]
vreuter: That'll probably fix it, yes

tests/test_settings.py
```diff
@@ -128,6 +130,62 @@ def test_base_path_overrides_config_file_code_and_deps(
...
+        )
+        parameter_grids.append(split_2)
+    parameter_grids = list(product(*parameter_grids))
+    command_lines = []
+    for param_grid in parameter_grids:
+        xss = list(chain(*param_grid))
+        random.shuffle(xss)
+        optargs = list(chain(*[[f"--{opt_by_arg[xs[0]]}"] + list(xs) for xs in xss]))
+        command_lines.append(optargs)
+    return command_lines
```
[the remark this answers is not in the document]
vreuter: And the `random.shuffle` is to demonstrate / prove that the result should be invariant to the order of specification on the command-line (the options themselves, and the arguments for each one).

fawltydeps/settings.py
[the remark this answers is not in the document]
vreuter: Ah no `mypy` will which about explicitly having `Any` as the element type, even though that's part of the union that's declared there in the parent class (`Union[str, Sequence[Any], None]`)

tests/test_settings.py
```diff
@@ -128,6 +130,62 @@ def test_base_path_overrides_config_file_code_and_deps(
...
+        return chain.from_iterable(combinations(xs, k) for k in range(1, len(xs)))
+
+    opt_by_arg = {arg: opt for opt, argvals in OPTION_VALUES.items() for arg in argvals}
+    parameter_grids = []
+    for items in OPTION_VALUES.values():
+        split_2 = set(
+            chain.from_iterable(
+                [(x, y), (y, x)]
+                for x, y in combinations(powerset_non_empties(items), 2)
+                if len(x + y) == len(items) and not set(x) & set(y)
```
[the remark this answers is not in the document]
vreuter: Yes, maybe I should even write this as a small named helper, like "partition"

tests/test_settings.py
jherland: I see your point in favor of parametrization. Happy to see where this goes. 😄

tests/test_settings.py
```diff
@@ -128,6 +130,62 @@ def test_base_path_overrides_config_file_code_and_deps(
...
+    command_lines = []
+    for param_grid in parameter_grids:
+        xss = list(chain(*param_grid))
+        random.shuffle(xss)
+        optargs = list(chain(*[[f"--{opt_by_arg[xs[0]]}"] + list(xs) for xs in xss]))
+        command_lines.append(optargs)
+    return command_lines
+
+
+@pytest.mark.parametrize("optargs", multivalued_optargs_grid())
```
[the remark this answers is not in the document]
vreuter: OK, i make it lazier here: 400e34a7330944499435ce1a63439179a68d6ca2

tests/test_settings.py
vreuter: OK, i've gone with an implementation based on this idea here: 400e34a7330944499435ce1a63439179a68d6ca2

tests/test_settings.py
```diff
@@ -128,6 +130,62 @@ def test_base_path_overrides_config_file_code_and_deps(
...
+}
+
+
+def multivalued_optargs_grid() -> List[List[str]]:
+    """Create command-line option/argument combinations which
+    mix order and number of multivalued parameters."""
+
+    def powerset_non_empties(iterable):
+        xs = list(iterable)
+        return chain.from_iterable(combinations(xs, k) for k in range(1, len(xs)))
```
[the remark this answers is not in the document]
vreuter: this is moot now since this helper's replaced per conversation on another thread.

fawltydeps/settings.py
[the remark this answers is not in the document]
vreuter: yeah, it's gone as of this commit, thanks for catching this: 400e34a7330944499435ce1a63439179a68d6ca2

tests/test_settings.py
vreuter: see what you think here @jherland , my pref isn't super strong here: 400e34a7330944499435ce1a63439179a68d6ca2

tests/test_settings.py
```diff
@@ -128,6 +130,60 @@ def test_base_path_overrides_config_file_code_and_deps(
...
+    assert settings.code == to_path_set(OPTION_VALUES["code"])
+    assert settings.deps == to_path_set(OPTION_VALUES["deps"])
+    assert settings.ignore_undeclared == set(OPTION_VALUES["ignore-undeclared"])
+    assert settings.ignore_unused == set(OPTION_VALUES["ignore-unused"])
+
+
+@pytest.mark.parametrize(
+    "optname",
+    set(act.dest for act in build_parser()._actions)  # pylint: disable=protected-access
+    & set(Settings.__fields__),
```
jherland, line 179: Nit: `Settings.__fields__` is a `dict` if I recall? Wouldn't this be more readable as?:
```suggestion
    & Settings.__fields__.keys(),
```

tests/test_settings.py
```diff
@@ -128,6 +130,60 @@ def test_base_path_overrides_config_file_code_and_deps(
...
+)
+def test_settings_members_are_absent_from_namespace_if_not_provided_at_cli(optname):
+    parsed_cli = build_parser().parse_args([])
+    with pytest.raises(AttributeError):
+        getattr(parsed_cli, optname)
```
jherland, line 184: Nit: I think this is more readable, if I have not overlooked any details:
```suggestion
    assert not hasattr(parsed_cli, optname)
```

jherland (approved):
Looks great! Ready to 🚀  modulo two small nits.

tests/test_settings.py
vreuter: I adopted your suggestion in the final commit here, thanks

tests/test_settings.py
vreuter: This one I'm going to leave, as I think the `raises` assertion is a bit stronger, like the difference between the membership check with `in` on a `Mapping` vs. the direct `[]` accessor. Most of the time the same, but with the `defaultdict`, only one would fail. I want to make the stronger type of assertion with this test.

Issue #166, opened 2023-02-18
Test refactor


mknorps:
As we advance in the complexity of objects representing the FD domain, our test code also grows in complexity. 
It takes notably more time to change the test code base when changing the basic objects that are being tested.

We should:

- [ ] rethink the structure of tests, as discussed with the team, to test one feature at a time (for example change in `Location` object should not propagate to all tests using `UndeclaredDependencies`)
- [x] gather common helper functions in `utils.py` and refactor tests cases to use them
- [ ] Use more fixtures not to repeat test cases between tests files and/or generate test cases automatically
- [x] Consider using objects with reasonable defaults for `pytest.param` inputs (test in `compare_imports_to_dependencies) have now 5 variables, for which most is just an empty list
- [x] (may be a separate issue) Consider using `hypothesis` for property-based testing

jherland:
This issue was discussed in today's status meeting, and a couple of points from the meeting notes are worth repeating here:
- Testing with hypothesis is very valuable, but applying it at the CLI level is too slow. Write hypothesis tests directly against the `Analysis` object instead.
- Focus more on testing with the `Analysis` object. It should be our main test “surface” esp. if FD ends up being used as a library.
- Rephrase CLI tests as Settings+Analysis tests. Not having to go through a subprocess will speed these tests up considerably.
-  (for later:) Consider the overlap of "sample projects" defined in `tests/conftest.py` vs. the projects in `tests/sample_projects`.

PR #250, opened 2023-03-20
Tests refactor: use a vector of test cases for parsing pyproject.toml tests 

mknorps:
Refactor list of pyproject.toml test cases to use a vector with default values, similar to 'test_vectors' from utils.

Applied to input of `test_parse_pyproject_content__malformatted_poetry_dependencies__yields_no_dependencies`, because the rest of the tests have at most 3 parameters per `pytest.param`.

In this PR no new test is added, only code moved around for better readability, application of further changes and unification with FDTestVector style started by @jherland.

Analog of #248  and will close point:
>  Consider using objects with reasonable defaults for pytest.param inputs (test in `compare_imports_to_dependencies) have now 5 variables, for which most is just an empty list

of #166.

### DIFF
--- a/tests/test_extract_declared_dependencies_pyproject_toml.py
+++ b/tests/test_extract_declared_dependencies_pyproject_toml.py
@@ -1,5 +1,7 @@
 """Test extracting dependencies from pyproject.toml"""
 import logging
+from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


jherland (approved):
😻

PR #251, opened 2023-03-20
Tests refactor: Change `main` to explicitly take stdin, stdout for better testability

mknorps:
This refactor makes it easier to test the core functionality of FawltyDeps,
without the necessity to run fawltydeps in a subprocess.

The following things are changed:
- Extract assigning exit code and printing summary message from 'main'.
- Based on @jherland's comments, add arguments to main for: stdin, stdout, list of cmdline args.
Refactor extracting imports accordingly.
- for `--code` using stdin (`-` option), added handling that  checks if the stdin is interactive, and throws error in that case.

### DIFF
--- a/fawltydeps/extract_imports.py
+++ b/fawltydeps/extract_imports.py
@@ -3,9 +3,8 @@
 import ast
 import json
 import logging
-import sys
 from pathlib import Path
-from typing import Iterable, Iterator, Optional, Set, Tuple
+from typing import Iterable, Iterator, Optional, Set, TextIO, Tuple
 
 import isort
 
@@ -14,6 +13,7 @@ from fawltydeps.types import (
     ParsedImport,
     PathOrSpecial,
     UnparseablePathException,
+    StdInNotProvidedException,
 )
 from fawltydeps.utils import dirs_between, walk_dir
 
@@ -172,7 +172,9 @@ def parse_dir(path: Path) -> Iterator[ParsedImport]:
             yield from parse_notebook_file(file, local_context=local_context)
 
 
-def parse_any_arg(arg: PathOrSpecial) -> Iterator[ParsedImport]:
+def parse_any_arg(
+    arg: PathOrSpecial, stdin: Optional[TextIO] = None
+) -> Iterator[ParsedImport]:
     """Interpret the given command-line argument and invoke a suitable parser.
 
     These cases are handled:
@@ -180,11 +182,17 @@ def parse_any_arg(arg: PathOrSpecial) -> Iterator[ParsedImport]:
       - arg refers to a file: Call parse_python_file() or parse_notebook_file()
       - arg refers to a dir: Call parse_dir()
 
-    Otherwise raise UnparseablePathException with a suitable error message.
+    Otherwise raise UnparseablePathException or StdInNotProvidedException
+    with a suitable error message.
     """
     if arg == "<stdin>":
         logger.info("Parsing Python code from standard input")
-        return parse_code(sys.stdin.read(), source=Location(arg))
+        if stdin:
+            return parse_code(stdin.read(), source=Location(arg))
+        raise StdInNotProvidedException(
+            ctx="There is no stdin to parse."
+            "When giving '-' as '--code' argument, provide a stdin."
+        )
     assert isinstance(arg, Path)
     if arg.is_file():
         if arg.suffix == ".py":
@@ -205,10 +213,12 @@ def parse_any_arg(arg: PathOrSpecial) -> Iterator[ParsedImport]:
     )
 
 
-def parse_any_args(args: Set[PathOrSpecial]) -> Iterator[ParsedImport]:
+def parse_any_args(
+    args: Set[PathOrSpecial], stdin: Optional[TextIO] = None
+) -> Iterator[ParsedImport]:
     """Interpret given set of command line arguments.
 
     Pass a list of paths from which to discover and parse imports from the code.
     """
     for arg in args:
-        yield from parse_any_arg(arg)
+        yield from parse_any_arg(arg, stdin)
--- a/fawltydeps/extract_imports.py
+++ b/fawltydeps/extract_imports.py
@@ -13,7 +13,6 @@ from fawltydeps.types import (
     ParsedImport,
     PathOrSpecial,
     UnparseablePathException,
-    StdInNotProvidedException,
 )
 from fawltydeps.utils import dirs_between, walk_dir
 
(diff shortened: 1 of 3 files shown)


fawltydeps/extract_imports.py
```diff
@@ -172,19 +172,32 @@ def parse_dir(path: Path) -> Iterator[ParsedImport]:
...
     """Interpret the given command-line argument and invoke a suitable parser.
 
     These cases are handled:
       - arg == "-": Read code from stdin and pass to parse_code()
       - arg refers to a file: Call parse_python_file() or parse_notebook_file()
       - arg refers to a dir: Call parse_dir()
 
-    Otherwise raise UnparseablePathException with a suitable error message.
+    Otherwise raise UnparseablePathException or StdInNotProvidedException
+    with a suitable error message.
```
vreuter, line 186: Nit: I don't think we lose any clarity if we omit this part of the commit message.
```suggestion
```

fawltydeps/extract_imports.py
```diff
@@ -172,19 +172,32 @@ def parse_dir(path: Path) -> Iterator[ParsedImport]:
...
     """
     if arg == "<stdin>":
         logger.info("Parsing Python code from standard input")
-        return parse_code(sys.stdin.read(), source=Location(arg))
+        # 'isatty' checks if the stream is interactive.
+        # The way to use stdin is via piping data to the program:
+        # echo "import foo"| fawltydeps --code -
+        if stdin and not stdin.isatty():
```
vreuter, line 193: cool, TIL!

fawltydeps/main.py
```diff
@@ -211,53 +212,77 @@ def build_parser() -> argparse.ArgumentParser:
...
-
-    if args.generate_toml_config:
-        print_toml_config(settings, sys.stdout)
-        return 0
-
-    try:
-        analysis = Analysis.create(settings)
-    except UnparseablePathException as exc:
-        return parser.error(exc.msg)  # exit code 2
+def assign_exit_code(analysis: Analysis) -> int:
```
vreuter, line 215: love this

fawltydeps/main.py
```diff
@@ -211,53 +212,77 @@ def build_parser() -> argparse.ArgumentParser:
...
+    3 - undeclared dependencies found
+    4 - unused dependencies found
+    """
     exit_code = 0
     if analysis.is_enabled(Action.REPORT_UNDECLARED) and analysis.undeclared_deps:
         exit_code = 3
     elif analysis.is_enabled(Action.REPORT_UNUSED) and analysis.unused_deps:
         exit_code = 4
 
+    return exit_code
```
vreuter, line 232: Nit:

I'd slightly prefer if <A>: return 3, if <B>: return 4, then just return 0 at the end (i.e., `elif` becomes `if`, and rather than initalizing and reassigning `exit_code`, there are just the raw values), but OK either way

vreuter (approved):
looks good, just a couple suggestions

tests/test_extract_imports_errors.py
```diff
@@ -76,6 +79,13 @@ def test_parse_code__on_parse_error__logs_error(caplog):
     assert f"Could not parse code from {source}" in caplog.text
 
 
+def test_parse_any_args__no_stdin_provided__logs_error():
+    stdin = None
+    arg = "<stdin>"
+    with pytest.raises(StdInNotProvidedException):
+        parse_any_arg(arg, stdin)
```
jherland, line 86: Related to my reservations about `StdInNotProvidedException`, I'm questioning the reason for this test.

I think it tests something that a CLI user should not be able to trigger in any way, and I think if we make `parse_any_arg()` take a `stdin: TextIO = io.StringIO()` argument instead, we can avoid `StdInNotProvidedException` altogether. What do you think?

jherland (changes requested):
Looks good. I want to reconsider `StdInNotProvidedException`, but otherwise I have only minor comments.

fawltydeps/extract_imports.py
```diff
@@ -172,19 +172,32 @@ def parse_dir(path: Path) -> Iterator[ParsedImport]:
...
+            return parse_code(stdin.read(), source=Location(arg))
+        raise StdInNotProvidedException(
+            ctx="There is no stdin to parse.\n"
+            "When giving '-' as '--code' argument, provide a stdin input "
+            "and pipe it, e.g.:\n"
+            "`echo 'import foo'| fawltydeps --code -`"
+        )
```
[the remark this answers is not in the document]
mknorps: Thanks for the explanation, I think what you described deserves a FAQ entry.

The Usage of `--code -` was very vague for me, the only way I knew how to use it was with piping. 

> type in some python code at the terminal, ending it with EOF (Ctrl+D)

TIL! I did not know that this is how you should end stdin stream. And I know how to exit VIM :sweat_smile: 

I think we should either add this info to the README or display it when a user runs `fawltydeps --code -` (just as it is done in some REPLs).

`isatty` check will not be useful in that case indeed, and all the code that followed (incl. `StdInNotProvidedException`)

tests/test_extract_imports_errors.py
mknorps: Agreed.

fawltydeps/main.py
```diff
@@ -211,53 +212,77 @@ def build_parser() -> argparse.ArgumentParser:
...
-    # Exit codes:
-    # 0 - success, no problems found
-    # 1 - an exception propagates (this should not happen)
-    # 2 - command-line parsing error (see above)
-    # 3 - undeclared dependencies found
-    # 4 - unused dependencies found
+    Exit codes:
+    0 - success, no problems found
+    1 - an exception propagates (this should not happen)
+    2 - command-line parsing error (see above)
```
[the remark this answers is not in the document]
mknorps: rephrased

fawltydeps/main.py
```diff
@@ -211,53 +211,74 @@ def build_parser() -> argparse.ArgumentParser:
...
-        exit_code = 3
-    elif analysis.is_enabled(Action.REPORT_UNUSED) and analysis.unused_deps:
-        exit_code = 4
 
+def print_output(
+    analysis: Analysis,
+    exit_code: int,
+    stdout: TextIO = sys.stdout,
+) -> None:
+    """Print the summary message for CLI run."""
```
jherland, line 238: Nit: Docstring no longer correspond to function name:

```suggestion
    """Print the output of the given 'analysis' to 'stdout'."""
```

jherland (approved):
🚀

Issue #246, opened 2023-03-17
Unable to detect BeautifulSoup html5lib parser


sybo-matthew:
**Describe the bug**
FawltyDeps will recommend the removal of the HTML5 parser used by BeautifulSoup

**To Reproduce**
- install beautifulsoup
- set beautifulsoup to use the html5lib parser
- install the html5lib parser with pip install html5lib 
- run fawltydeps and it will recommend the removal of the library as it is not called by import
- attempt to run the script by following fawltydeps recommendations and it will not work as beautifulsoup cannot find the parser it needs

**Expected behavior**
- either to detect that the parser has been called and include it in the list of requirements or to add it to a package blacklist where it'll be ignored for removal considerations. 

**Environment**
 -  macOS Venture 13.2.1  Python 3.11.2 FawltyDeps v0.7.0

If needed I can make a simple test project to display the behavior as well.

Nour-Mws:
Thanks for reporting this @sybo-matthew!

We have an item on our short to medium term roadmap #195 to enhance imports parsing in non straightforward cases. Though this particular case has not been brought to our attention before.

In the meantime, you can run fawltydeps with the `--ignore-unused html5lib` option. 

## Steps to reproduce
1. Python project with `requirements.txt` and `script.py`:
```
# requirements.txt
beautifulsoup4
html5lib
```

```python
# script.py
from bs4 import BeautifulSoup
BeautifulSoup("<a></p>", "html5lib")

```

2. A Python environment `soup_env` where the above requirements are installed.

3. Run fawltydeps 0.7.0 as follows:
```shell
fawltydeps bsoup_project --pyenv soup_env
```

**Result**
```
These dependencies appear to be unused (i.e. not imported):
- 'html5lib'

For a more verbose report re-run with the `--detailed` option.
```

jherland:
I took a look at beautifulsoup itself, and it seems that since v4.4.0 (July 2015), it has provided an `html5lib` "extra" that pulls in `html5lib` as a dependency of `beautifulsoup4`. In other words: if you install `beautifulsoup4[html5lib]` _instead_ of just `beautifulsoup`, you will get `html5lib` automatically without having to install it separately. Since you can now remove your direct dependency on `html5lib`, FawltyDeps will no longer report it as unused.

This seems not to be advertised in beautifulsoup's own documentation, which IMHO is not ideal.

[...]

PR #253, opened 2023-03-21
Develop sample_projects to support multiple experiments and cached venvs

jherland:
This started as a simple wish to add the problem described in issue #246 as a sample project in our test suite.

However, I quickly discovered that our `sample_projects` did not support installing comples dependencies (such as `beautifulsoup4`, much less `beautifulsoup4[html5lib]`. Furthermore, it also did not support running multiple experiments inside a single `sample_project`.

Both of these features are already available in our `real_projects`, so I took this as an opportunity to do some extensive refactoring on `real_projects` and `sample_projects`, and bring these two parts of our test suite as close together as possible:

The common parts are refactored into a new `project_helpers` module (unders `tests/`), and the remaining differences are expressed in the `Experiment` and `*Project` subclasses that now remain in both `test_real_projects` and `test_sample_projects`.

The test suite is kept unchanged and working throughout the refactoring, save for some light refactoring of `sample_projects` TOML files in the second-last commit, and the addition of the new `beautifulsoup_and_html5lib` test (that triggered all of this) in the last commit.

Commits:
- `test_real_projects`: Refactor cached venv code into separate module
- `test_real_projects`: Refactor verification of expected analysis report
- `test_sample_projects`: Minimal rewrite to use `AnalysisExpectations`
- `test_real_project`: Refactor common project/experiment details
- `test_sample_projects`: Refactor to use `BaseProject`/`BaseExperiment`
- `test_sample_projects`: Add new project: `beautifulsoup_and_html5lib`

### DIFF
--- a/tests/project_helpers.py
+++ b/tests/project_helpers.py
@@ -5,10 +5,16 @@ import shlex
 import subprocess
 import sys
 from dataclasses import dataclass
+from dataclasses import fields as dataclass_fields
 from pathlib import Path
-from typing import List
+from typing import Any, Callable, Dict, Iterable, List, Optional, Set
 
 import pytest
+
+from fawltydeps.main import Analysis
+from fawltydeps.types import TomlData
+
+JsonData = Dict[str, Any]
 
 logger = logging.getLogger(__name__)
 
@@ -84,3 +90,58 @@ class CachedExperimentVenv:
         assert (venv_dir / ".installed").is_file()
         cache.set(f"fawltydeps/{self.venv_hash()}", str(venv_dir))
         return venv_dir
+
+
+@dataclass
+class AnalysisExpectations:
+    """Encode our expectations on the analysis resulting from an experiment.
+
+    Encapsulate the expected values of an Analysis object (or its JSON
+    counterpart) after a real/sample project experiment has been run.
+
+    The members of this expectation object are all optional; if omitted, the
+    corresponding Analysis member will _not_ be checked. Furthermore, no "deep"
+    verification of each member is performed here; rather, we only check that
+    the set of (import or dependency) names that can found in each Analysis
+    member match the expected set stored here.
+    """
+
+    imports: Optional[Set[str]] = None
+    declared_deps: Optional[Set[str]] = None
+    undeclared_deps: Optional[Set[str]] = None
+    unused_deps: Optional[Set[str]] = None
+
+    @classmethod
+    def parse_from_toml(cls, data: TomlData) -> "AnalysisExpectations":
+        """Read expectations from the given TOML table."""
+
+        def set_or_none(data: Optional[Iterable[str]]) -> Optional[Set[str]]:
+            return None if data is None else set(data)
+
+        return cls(
+            imports=set_or_none(data.get("imports")),
+            declared_deps=set_or_none(data.get("declared_deps")),
+            undeclared_deps=set_or_none(data.get("undeclared_deps")),
+            unused_deps=set_or_none(data.get("unused_deps")),
+        )
+
+    def _verify_members(self, member_extractor: Callable[[str], Set[str]]) -> None:
+        for member in dataclass_fields(self):
+            expected_names = getattr(self, member.name)
+            if expected_names is None:
+                print(f"Skip checking .{member.name}")
+                continue
+
+            print(f"Checking .{member.name}")
+            actual_names = member_extractor(member.name)
+            print(f"  Actual names: {sorted(actual_names)}")
+            print(f"  Expect names: {sorted(expected_names)}")
+            assert actual_names == expected_names
+
+    def verify_analysis(self, analysis: Analysis) -> None:
+        """Assert that the given Analysis object matches our expectations."""
+        self._verify_members(lambda member: {m.name for m in getattr(analysis, member)})
+
+    def verify_analysis_json(self, analysis: JsonData) -> None:
+        """Assert that the given JSON analysis matches our expectations."""
+        self._verify_members(lambda member: {m["name"] for m in analysis[member]})
--- a/tests/project_helpers.py
+++ b/tests/project_helpers.py
@@ -7,12 +7,28 @@ import sys
 from dataclasses import dataclass
 from dataclasses import fields as dataclass_fields
 from pathlib import Path
-from typing import Any, Callable, Dict, Iterable, List, Optional, Set
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Type,
+)
 
 import pytest
 
 from fawltydeps.main import Analysis
 from fawltydeps.types import TomlData
+
+if sys.version_info >= (3, 11):
+    import tomllib  # pylint: disable=E1101
+else:
+    import tomli as tomllib
 
 JsonData = Dict[str, Any]
 
@@ -145,3 +161,85 @@ class AnalysisExpectations:
     def verify_analysis_json(self, analysis: JsonData) -> None:
         """Assert that the given JSON analysis matches our expectations."""
         self._verify_members(lambda member: {m["name"] for m in analysis[member]})
+
+
+@dataclass
+class BaseExperiment:
+    """A single experiment, running FawltyDeps on a test project.
+
+    An experiment is part of a bigger project (see BaseProject below) and has:
+    - A name and description, for documentation purposes.
+    - A list of requirements, to be installed into a virtualenv and made
+      available to FawltyDeps when this experiment is run
+      (see CachedExperimentVenv for details).
+    - A set of expectations on the resulting Analysis object, to be verified
+      after the FawltyDeps has been run (see AnalysisExpectations for details).
+    """
+
+    name: str
+    description: Optional[str]
+    requirements: List[str]
+    expectations: AnalysisExpectations
+
+    @staticmethod
+    def init_args_from_toml(name: str, data: TomlData) -> Dict[str, Any]:
+        return dict(
+            name=name,
+            description=data.get("description"),
+            requirements=data.get("requirements", []),
+            expectations=AnalysisExpectations.parse_from_toml(data),
+        )
+
+    @classmethod
+    def parse_from_toml(cls, name: str, data: TomlData) -> "BaseExperiment":
+        raise NotImplementedError
+
+    def get_venv_dir(self, cache: pytest.Cache) -> Path:
+        """Get this venv's dir and create it if necessary."""
+        return CachedExperimentVenv(self.requirements)(cache)
+
+
+@dataclass
+class BaseProject:
+    """Encapsulate a Python project to be tested with FawltyDeps.
+
+    This represents a project on which we want to run FawltyDeps in one or more
+    experiments. It has at least:
+    - A name and optional description, for documentation purposes.
+    - A list of experiments (see BaseExperiment above), describing one or more
+      scenarios for running FawltyDeps on this project, and what results to
+      expect in those scenarios.
+    """
+
+    name: str
+    description: Optional[str]
+    experiments: List[BaseExperiment]
+
+    @staticmethod
+    def parse_toml(
+        toml_path: Path, ExperimentClass: Type[BaseExperiment]
+    ) -> Tuple[Dict[str, Any], TomlData]:
+        try:
+            with toml_path.open("rb") as f:
+                toml_data = tomllib.load(f)
+        except tomllib.TOMLDecodeError:
+            print(f"Error occurred while parsing file: {toml_path}")
+            raise
+
+        # We ultimately _trust_ the .toml files read here, so we can skip all
+        # the usual error checking associated with validating external data.
+        project_name = toml_data["project"]["name"]
+        init_args = dict(
+            name=project_name,
+            description=toml_data["project"].get("description"),
+            experiments=[
+                ExperimentClass.parse_from_toml(f"{project_name}:{name}", data)
+                for name, data in toml_data["experiments"].items()
+            ],
+        )
+        return init_args, toml_data
+
+    @classmethod
+    def collect(cls) -> Iterator["BaseProject"]:
+        """Find and generate all projects in this test suite."""
+        raise NotImplementedError
--- a/tests/project_helpers.py
+++ b/tests/project_helpers.py
@@ -7,18 +7,7 @@ import sys
 from dataclasses import dataclass
 from dataclasses import fields as dataclass_fields
 from pathlib import Path
-from typing import (
-    Any,
-    Callable,
-    Dict,
-    Iterable,
-    Iterator,
-    List,
-    Optional,
-    Set,
-    Tuple,
-    Type,
-)
+from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Set, Type
 
 import pytest
 
@@ -182,7 +171,8 @@ class BaseExperiment:
     expectations: AnalysisExpectations
 
     @staticmethod
-    def init_args_from_toml(name: str, data: TomlData) -> Dict[str, Any]:
+    def _init_args_from_toml(name: str, data: TomlData) -> Dict[str, Any]:
+        """Extract members from TOML into kwargs for a subclass constructor."""
         return dict(
             name=name,
             description=data.get("description"),
@@ -192,6 +182,7 @@ class BaseExperiment:
 
     @classmethod
     def from_toml(cls, name: str, data: TomlData) -> "BaseExperiment":
+        """Create an instance from TOML data."""
         raise NotImplementedError
 
     def get_venv_dir(self, cache: pytest.Cache) -> Path:
@@ -216,20 +207,14 @@ class BaseProject:
     experiments: List[BaseExperiment]
 
     @staticmethod
-    def parse_toml(
-        toml_path: Path, ExperimentClass: Type[BaseExperiment]
-    ) -> Tuple[Dict[str, Any], TomlData]:
-        try:
-            with toml_path.open("rb") as f:
-                toml_data = tomllib.load(f)
-        except tomllib.TOMLDecodeError:
-            print(f"Error occurred while parsing file: {toml_path}")
-            raise
-
+    def _init_args_from_toml(
+        toml_data: TomlData, ExperimentClass: Type[BaseExperiment]
+    ) -> Dict[str, Any]:
+        """Extract members from TOML into kwargs for a subclass constructor."""
         # We ultimately _trust_ the .toml files read here, so we can skip all
         # the usual error checking associated with validating external data.
         project_name = toml_data["project"]["name"]
-        init_args = dict(
+        return dict(
             name=project_name,
             description=toml_data["project"].get("description"),
             experiments=[
@@ -237,9 +222,17 @@ class BaseProject:
                 for name, data in toml_data["experiments"].items()
             ],
         )
-        return init_args, toml_data
 
     @classmethod
     def collect(cls) -> Iterator["BaseProject"]:
         """Find and generate all projects in this test suite."""
         raise NotImplementedError
+
+
+def parse_toml(toml_path: Path) -> TomlData:
+    try:
+        with toml_path.open("rb") as f:
+            return tomllib.load(f)
+    except tomllib.TOMLDecodeError:
+        print(f"Error occurred while parsing file: {toml_path}")
+        raise
--- a/tests/project_helpers.py
+++ b/tests/project_helpers.py
@@ -128,7 +128,7 @@ class AnalysisExpectations:
     unused_deps: Optional[Set[str]] = None
 
     @classmethod
-    def parse_from_toml(cls, data: TomlData) -> "AnalysisExpectations":
+    def from_toml(cls, data: TomlData) -> "AnalysisExpectations":
         """Read expectations from the given TOML table."""
 
         def set_or_none(data: Optional[Iterable[str]]) -> Optional[Set[str]]:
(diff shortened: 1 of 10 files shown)


tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+    member match the expected set stored here.
+    """
+
+    imports: Optional[Set[str]] = None
+    declared_deps: Optional[Set[str]] = None
+    undeclared_deps: Optional[Set[str]] = None
+    unused_deps: Optional[Set[str]] = None
+
+    @classmethod
+    def parse_from_toml(cls, data: TomlData) -> "AnalysisExpectations":
```
vreuter, line 130: What's the reason to name and type this as TOML-specific rather than mapping-like? Intuitively, I'd name this `parse_from_mapping` and have `data: Mapping[str, Any]`, but am I missing something?

vreuter: I mean, `TomlData = Dict[str, Any]`, right?

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+        """Read expectations from the given TOML table."""
+
+        def set_or_none(data: Optional[Iterable[str]]) -> Optional[Set[str]]:
```
vreuter, line 133: :+1: for the naming and for making this helper

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+            expected_names = getattr(self, field.name)
+            if expected_names is None:
+                print(f"Skip checking .{field.name}")
+                continue
+
+            print(f"Checking .{field.name}")
+            actual_names = member_extractor(field.name)
+            print(f"  Expect names: {sorted(expected_names)}")
+            print(f"  Actual names: {sorted(actual_names)}")
+            assert expected_names == actual_names
```
vreuter, line 154: Nit: this is the norm from most other languages I think, but doesn't it flip the Python norm (observed on left-hand side, expected on right-hand side)?

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+        )
+        # Make sure the venv has been installed
+        assert Path(venv_dir, ".installed").is_file()
+        cache.set(f"fawltydeps/{self.venv_hash()}", str(venv_dir))
+        return venv_dir
+
+
+@dataclass
+class AnalysisExpectations:
+    """Verify our expectations on the analysis resulting from an experiment.
```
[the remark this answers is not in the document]
vreuter: Nevermind this, it looks like this type has more responsibilities than I'd anticipated.

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+            print(f"  Actual names: {sorted(actual_names)}")
+            assert expected_names == actual_names
+
+    def verify_analysis(self, analysis: Analysis) -> None:
+        """Assert that the given Analysis object matches our expectations."""
+        self._verify_fields(lambda member: {m.name for m in getattr(analysis, member)})
+
+    def verify_analysis_json(self, analysis: JsonData) -> None:
+        """Assert that the given JSON analysis matches our expectations."""
+        self._verify_fields(lambda member: {m["name"] for m in analysis[member]})
```
vreuter, line 162: Ah, the world will be a happier place when we can fully leverage union types and structural pattern matching :)

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+    def init_args_from_toml(name: str, data: TomlData) -> Dict[str, Any]:
+        return dict(
+            name=name,
+            description=data.get("description"),
+            requirements=data.get("requirements", []),
+            expectations=AnalysisExpectations.parse_from_toml(data),
+        )
+
+    @classmethod
+    def parse_from_toml(cls, name: str, data: TomlData) -> "BaseExperiment":
```
vreuter, line 193: Nit:
```suggestion
    def from_toml(cls, name: str, data: TomlData) -> "BaseExperiment":
```

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+        raise NotImplementedError
+
+    def get_venv_dir(self, cache: pytest.Cache) -> Path:
+        """Get this venv's dir and create it if necessary."""
+        venv = CachedExperimentVenv(self.requirements)
+        return venv(cache)
```
vreuter, line 199: Nit:
```suggestion
        return CachedExperimentVenv(self.requirements)(cache)
```
if that's allowed syntax-wise

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+@dataclass
+class BaseProject:
+    """Encapsulate a Python project to be tested with FawltyDeps.
+
+    This represents project on which we want to run FawltyDeps in one or more
```
vreuter, line 206:
```suggestion
    This represents a project on which we want to run FawltyDeps in one or more
```

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+    experiments. It has at least:
+    - A name and description, for documentation purposes.
```
vreuter, line 208: Is this true? Below we have `description` as optional

tests/test_sample_projects.py
```diff
@@ -20,65 +19,100 @@
...
+
+    Input to the experiment consists of the following members:
+    - code: Settings.code paths relative to the sample project root
+    - deps: Settings.deps paths relative to the sample project root
+
+    See BaseExperiment for details on the inherited members.
+    """
+
+    code: Optional[List[str]]
+    deps: Optional[List[str]]
```
vreuter, line 52: Can we go with just `List` here and have `[]` be equivalent to `None`? An optional NonEmptyList makes sense, and an ordinary list makes sense, but having the two layers of possibility of emptiness feels strange.

tests/test_sample_projects.py
```diff
@@ -20,65 +19,100 @@
...
+
+    @classmethod
+    def parse_from_toml(cls, name: str, data: TomlData) -> "Experiment":
+        return cls(
+            code=data.get("code"),
+            deps=data.get("deps"),
```
vreuter, line 58:
```suggestion
            code=data.get("code") or [],
            deps=data.get("deps") or [],
```
if removing the optionality wrapper

tests/test_sample_projects.py
```diff
@@ -20,65 +19,100 @@
...
+        """Construct a Settings object appropriate for this experiment."""
+        return Settings(
+            actions={Action.REPORT_UNDECLARED, Action.REPORT_UNUSED},
+            code=[
+                (project_path / path)
+                for path in ([""] if self.code is None else self.code)
+            ],
+            deps=[
+                (project_path / path)
+                for path in ([""] if self.deps is None else self.deps)
```
vreuter, line 72:
```suggestion
        return Settings(
            actions={Action.REPORT_UNDECLARED, Action.REPORT_UNUSED},
            code=[
                (project_path / path)
                for path in (self.code or [""])
            ],
            deps=[
                (project_path / path)
                for path in (self.deps or [""])
```

vreuter: or if removing the optionality wrapper...
```suggestion
        return Settings(
            actions={Action.REPORT_UNDECLARED, Action.REPORT_UNUSED},
            code=[(project_path / path) for path in self.code] or [project_path],
            deps=[(project_path / path) for path in self.deps] or [project_path]
```

vreuter (changes requested):
overall looks good, just some suggestions

tests/test_sample_projects.py
mknorps: Or even more generic:


```suggestion
            code=data.get("code", []),
            deps=data.get("deps", []),
```

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
+"""Common helpers shared between test_real_project and test_sample_projects."""
+import hashlib
+import logging
+import shlex
+import subprocess
+import sys
+from dataclasses import dataclass, fields
```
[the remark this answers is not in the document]
mknorps: In the rest of the code it is often the case that we use `fields` that way. I think it is consistent with the rest.

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+    packages used in our integration tests. The goal is that venvs are created
+    _once_ and reused as long as the commands used to create them and install
+    packages into them remain unchanged. These use pytest's caching
+    infrastructure meaning that the venvs typically end up living under
+    ~/.cache/pytest/d/...
+    """
+
+    requirements: List[str]  # PEP 508 requirements, passed to 'pip install'
+
+    def venv_script(self, venv_path: Path):
```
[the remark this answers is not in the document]
mknorps: Or 
```suggestion
    def venv_script(self, venv_path: Path) -> List[str]:
```

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+
+    name: str
+    description: Optional[str]
+    experiments: List[BaseExperiment]
+
+    @staticmethod
+    def parse_toml(
+        toml_path: Path, ExperimentClass: Type[BaseExperiment]
+    ) -> Tuple[Dict[str, Any], TomlData]:
```
mknorps, line 221: I would expect that parsing tokml will result in a `BaseProject` object.

mknorps (changes requested):
This is a very elegant solution. I especially like what you gathered in `project_helpers` to give a proper structure to the experiments.

This PR became quite big and basically is a rewrite of the current way we are handling real-world and simple experiments. It is clear what is going on if you read the code, not the diff. That is why, I think it would be good to fill in the gaps in parsing toml for `BaseProject` and `BaseExperiment` in this PR and treat it as a closed feature implementation (along with a refactor).

tests/project_helpers.py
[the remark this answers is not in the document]
jherland: I'll do both 😆

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+
+        The installation script will change if the code to setup the venv on
+        `venv_script` changes or if the requirements field of the experiment
+        changes. It will also be different for different Python versions.
+        The Python version currently used to run the tests is used to compute
+        the hash and create the venv.
+        """
+        dummy_script = self.venv_script(Path("/dev/null"))
+        py_version = f"{sys.version_info.major},{sys.version_info.major}"
+        script_and_version_bytes = ("".join(dummy_script) + py_version).encode()
```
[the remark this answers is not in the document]
jherland: Hmm. It sort of works out, since we're only interested in calculating a hash over the contents here. That said I agree that it's not ideal.

A problem with changing this now is that it'll invalidate the current cache and have the next CI run be a lot slower.

@Nour-Mws is also currently working on setting up venvs as part of #252, and that code will likely look a bit different (less shell code), and I would consider rather rewriting this test code to reuse that functionality when it has been added. Will leave this as-is for now, I think.

tests/project_helpers.py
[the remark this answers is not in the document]
jherland: I like "encode", will use that. Thanks!

tests/project_helpers.py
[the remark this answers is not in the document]
vreuter: yeah I like that

tests/project_helpers.py
jherland: What @mknorps said. I want to communicate that this is data from a TOML file (which tells the reader more than a simple `Dict[str, Any]` or equivalent), and I _also_ want to make it easier (in a possible future where the TOML parser itself returns something more strictly typed than `Any`) to narrow the definition of `TomlData` to something stricter.

vreuter: ahhh OK, makes sense. i was trying to generalize / broaden but makes sense the potential desire to instead prepare for the opposite, potential to narrow.

tests/test_sample_projects.py
vreuter: haha yes definitely use @mknorps suggestion rather than mine here :sweat_smile:

tests/project_helpers.py
vreuter: This threw me / caused me a bit of pause at first, too, but it made a little more sense once I reconsidered and viewed it as an internal helper. I think a couple things would help lessen surprise here, though.
a) We strive to name members that parse some file or preprocessed data _and_ return an instance of the containing type as `from_*` (rather than `parse_*` or `parse_from_*`) and reserve the `parse_*`-like naming for helpers.
b) This appears to be a method called only from subclasses, so we could prefix with an underscore to reinforce the idea that it's a helper rather than an API member like a `from_*` method tagged as `classmethod` would be expected to be.

tests/project_helpers.py
jherland: It's allowed, but I've found that readers not familiar with FP can often get tripped up by the double parentheses. Then again, if FP is absolutely foreign to them, then I suspect they will have a bad time reading our codebase in any case... 😅

tests/project_helpers.py
jherland: True, will update the docstring.

tests/test_sample_projects.py
jherland: Agreed. I guess I was confused for a moment there between having something be optional in the TOML file (which is handled by the `data.get("code")` and being optional in the object itself).

Note, however that the default (that correspond to `None`) is not `[]` but rather `[""]` (which ends up being roughly equivalent to `--code=.` after everything is prefixed with `project_path` in `.settings()`)

tests/test_sample_projects.py
jherland: Actually, after dropping the `Optional` around the `List[str]` This becomes even simpler:
```python
    code: List[str]
    deps: List[str]

    @classmethod
    def from_toml(cls, name: str, data: TomlData) -> "Experiment":
        return cls(
            code=data.get("code", [""]),
            deps=data.get("deps", [""]),
            **cls.init_args_from_toml(name, data),
        )

    def settings(self, project_path: Path, cache: pytest.Cache) -> Settings:
        """Construct a Settings object appropriate for this experiment."""
        return Settings(
            actions={Action.REPORT_UNDECLARED, Action.REPORT_UNUSED},
            code=[(project_path / path) for path in self.code],
            deps=[(project_path / path) for path in self.deps],
            pyenv=self.get_venv_dir(cache),
        )

```

tests/project_helpers.py
jherland: Yeah, I tried that at first, but I'm not sure how to do a multi-stage construction like that with inherited dataclasses: The base class does not really exist as an instance separate from the subclass; when I construct a `SampleProject` or `ThirdPartyProject` object, that is done with _one_ `cls(...)` call (inside the respective `.collect()` class methods), and I don't know how to inject/pass a `BaseProject` object into that construction.

If we were writing "regular" classes we would be in charge of writing the constructor ourselves, and we would probably end up with something like this:

```python
class BaseProject:
    @classmethod
    def parse_toml(toml_path: Path) -> TomlData:
        ...
        with toml_path.open("rb") as f:
            toml_data = tomllib.load(f)

    def __init__(self, toml_data: TomlData):
        self.name = toml_data["project"]["name"]
        self.description = toml_data["project"].get("description")
        self.experiments = ...

class SampleProject(BaseProject):
    def __init__(self, toml_data: TomlData):
        super().__init__(self, toml_data)
        self.path = ...
```

But since `dataclass` creates the constructor for us, we cannot pass around the `toml_data` in the same manner and extract the relevant information at each level of the class hierarchy.

In the end I settled on the structure you can see in `BaseExperiment.init_args_from_toml()`: We parse the relevant members from the TOML in the base class, but then we return them back to be subclass to then be passed to the `cls()` constructore. I see now that I failed to do the ~same clearer design for `BaseProject`. Will make another small refactoring here with extra docstrings to try to make things clearer.

tests/project_helpers.py
jherland: fixed in [4b201cd](https://github.com/tweag/FawltyDeps/pull/253/commits/4b201cd3f9e2585b3e609762c1e55b1fca5bb2b6)

tests/project_helpers.py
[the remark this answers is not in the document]
jherland: fixed in 6e14ef652cfe0e795d16a174bb792c157758659c

tests/project_helpers.py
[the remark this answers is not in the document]
jherland: fixed in 6e14ef652cfe0e795d16a174bb792c157758659c

tests/project_helpers.py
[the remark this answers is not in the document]
jherland: fixed in be6624a3b381129cd8d0b966b98d558c61bc8cf7

tests/project_helpers.py
jherland: fixed in 6e14ef652cfe0e795d16a174bb792c157758659c

tests/project_helpers.py
jherland: fixed in e48632b68e64a1fb07cc5534abd98106774653f7

tests/project_helpers.py
jherland: fixed in b8865ad

tests/test_sample_projects.py
jherland: fixed in https://github.com/tweag/FawltyDeps/commit/7011cf26b9325c1f94b3ac5408908ded3e351a9b

tests/project_helpers.py
jherland: (crap, GitHub only showed me @vreuter's answer _after_ I had written the above in reply to @mknorps's comment)

Anyway, In b344f5b05a29ca1730f1fb256da069f704b08d02 I've tried to clean things up, and document the "protocol" here:
- The actual parsing of the TOML file is now in a `parse_toml()` helper function, as @vreuter suggests
- What remains in `BaseProject` is an `init_args_from_toml()` helper method, equivalent/corresponding to the one already in `BaseExperiment`. These helpers are indeed called from subclasses to delegate the extraction of the base class members from the parsed TOML data.

I added an underscore prefix as suggested by @vreuter in df87c66.

tests/project_helpers.py
```diff
@@ -0,0 +1,238 @@
...
+        """Get this venv's dir and create it if necessary.
+
+        The venv_dir is where we install the dependencies of the current
+        experiment. It is keyed by the sha256 checksum of the requirements
+        file and the script we use for setting up the venv. This way, we
+        don't risk using a previously cached venv for a different if the
+        script or the requirements to create that venv change.
+        """
+        # We cache venv dirs using the hash from create_venv_hash
+        cached_str = cache.get(f"fawltydeps/{self.venv_hash()}", None)
```
vreuter, line 81: Nit: should be able to drop the `None` here, right? I think that's the default.

tests/project_helpers.py
```diff
@@ -0,0 +1,245 @@
...
+        venv_dir = Path(cache.mkdir(f"fawltydeps_venv_{self.venv_hash()}"))
+        logger.info(f"Creating venv at {venv_dir}...")
+        venv_script = self.venv_script(venv_dir)
+        subprocess.run(
+            " && ".join(venv_script),
+            check=True,  # fail if any of the commands fail
+            shell=True,  # pass multiple shell commands to the subprocess
+        )
+        # Make sure the venv has been installed
+        assert Path(venv_dir, ".installed").is_file()
```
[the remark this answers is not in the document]
vreuter: Nit: slight preference for the second option just because it's shorter and _slightly_ more obvious what's going on

tests/project_helpers.py
```diff
@@ -0,0 +1,238 @@
...
+    """Encode our expectations on the analysis resulting from an experiment.
+
+    Encapsulate the expected values of an Analysis object (or its JSON
+    counterpart) after a real/sample project experiment has been run.
+
+    The members of this expectation object are all optional; if omitted, the
+    corresponding Analysis member will _not_ be checked. Furthermore, no "deep"
+    verification of each member is performed here; rather, we only check that
+    the set of (import or dependency) names that can found in each Analysis
+    member match the expected set stored here.
```
vreuter, line 111: Really good explanation :+1:

tests/test_sample_projects.py
```diff
@@ -20,65 +19,94 @@
...
+
+    @classmethod
+    def from_toml(cls, name: str, data: TomlData) -> "Experiment":
+        return cls(
+            code=data.get("code", [""]),
+            deps=data.get("deps", [""]),
+            **cls._init_args_from_toml(name, data),
+        )
+
+    def settings(self, project_path: Path, cache: pytest.Cache) -> Settings:
```
vreuter, line 62: Nit: slight preference with this named with a prefix that's more indicative of the action, e.g. `as_settings`, `to_settings`, `build_settings`, `create_settings`

vreuter: Probably `build_settings` or `create_settings` among those 4 mentioned above, just since addition of additional data is needed, rather than simply converting the existing instance

vreuter (approved):
Looks good, just a few nits

mknorps (approved):
As previously said, I like this refactor a lot!

It is ready to :rocket: 

Small nit: Reding this for a second time, I realized, that `BaseExperiment` and `BaseProject` are abstract base classes for projects and sample project tests. It may be good to use `abc` to mark them as such,  with `from_toml`  and `collect` being  `abstractmethod`s. The small reservation I have about it is the behavior of a combination of ABC and dataclass, which I am not familiar with.

tests/project_helpers.py
jherland: Actually, no, this is pytest's `Cache.get()`m and neither the [docs](https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=cache%20get#pytest.Cache.get) nor the [source itself](https://docs.pytest.org/en/7.1.x/_modules/_pytest/cacheprovider.html#Cache.get) provide a `None` default-value for `default`, so we must pass it ourselves.

jherland:
> Small nit: Reding this for a second time, I realized, that `BaseExperiment` and `BaseProject` are abstract base classes for projects and sample project tests. It may be good to use `abc` to mark them as such, with `from_toml` and `collect` being `abstractmethod`s. The small reservation I have about it is the behavior of a combination of ABC and dataclass, which I am not familiar with.

Seems to work well. Added in b75b3be511ce74f866777db08be8f5a5ba057698. 👍🏻

tests/project_helpers.py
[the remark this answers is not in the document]
jherland: fixed in 344dc23b2e1ffe14725490443f0fc99ca56a1c68

PR #261, opened 2023-03-22
Extract default values of settings in cmdline tests

Nour-Mws:
After introducing `--install-deps` in [1443b05](https://github.com/tweag/FawltyDeps/pull/252/commits/1443b05ccd31adca20290f8069b3315d70281668), I had to update the expected json for multiple tests in the cmdline test module.
This could have been avoided if suitable defaults were provided.

Since there are 3 different tests concerned by this, and following @jherland's rule of three, I decided to extract the settings dict into a default customizable representation as has been done in the `test_settings` module. Note that I could not re-use the `make_settings_dict` function from that module as its representation is different from the json representation.

### DIFF
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -21,6 +21,25 @@ from .utils import assert_unordered_equivalence, run_fawltydeps
 logger = logging.getLogger(__name__)
 
 
+def make_json_settings_dict(**kwargs):
+    """
+    Create an expected version of Settings.dict(), with customizations.
+    """
+    settings = {
+        "actions": ["check_undeclared", "check_unused"],
+        "code": ["."],
+        "deps": ["."],
+        "pyenv": None,
+        "output_format": "human_summary",
+        "ignore_undeclared": [],
+        "ignore_unused": [],
+        "deps_parser_choice": None,
+        "verbosity": 0,
+    }
+    settings.update(kwargs)
+    return settings
+
+
 def test_list_imports_detailed__from_dash__prints_imports_from_stdin():
     code = dedent(
         """\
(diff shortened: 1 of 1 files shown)


tests/test_cmdline.py
```diff
@@ -21,6 +21,27 @@
 logger = logging.getLogger(__name__)
 
 
+def make_settings_dict(**kwargs):
+    """Create an expected version of Settings.dict(), with customizations.
+
+    Return a copy of DEFAULTS, with the given customizations applied.
+    """
```
jherland, line 28: I'd like "json" mentioned somewhere in the name or docstring here, to emphasize that this creates a settings dict of the JSON flavor, as opposed to the equivalent function in `test_settings`. Also, the `DEFAULTS` in the above docstring does not really make sense now.

tests/test_cmdline.py
```diff
@@ -21,6 +21,27 @@
...
+        "pyenv": None,
+        "output_format": "human_summary",
+        "ignore_undeclared": [],
+        "ignore_unused": [],
+        "deps_parser_choice": None,
+        "verbosity": 0,
+    }
+    ret = default_values.copy()
+    ret.update(kwargs)
+    return ret
```
jherland, line 42: The extra `.copy()` is not needed here when the dict is local to this function (and thus a new dict is create on each call):

```suggestion
    ret = {
        "actions": ["check_undeclared", "check_unused"],
        "code": ["."],
        "deps": ["."],
        "pyenv": None,
        "output_format": "human_summary",
        "ignore_undeclared": [],
        "ignore_unused": [],
        "deps_parser_choice": None,
        "verbosity": 0,
    }
    ret.update(kwargs)
    return ret
```

jherland (approved):
Only two small nits, otherwise ready to go! Appreciate the cleanup!

tests/test_cmdline.py
Nour-Mws: Done in [b4398da](https://github.com/tweag/FawltyDeps/pull/261/commits/b4398da4fa25c129c7fd37b78c1f01a6e43b0340).

tests/test_cmdline.py
Nour-Mws: Done in [b4398da](https://github.com/tweag/FawltyDeps/pull/261/commits/b4398da4fa25c129c7fd37b78c1f01a6e43b0340).

Issue #256, opened 2023-03-21
Mapping resolution strategy: discussion and implementation.


mknorps:
Execution plan:

- [x] Lay the groundwork for resolving packages via a list of resolvers: #268
- [x] #263 
- [x] [closed the part relevant to this milestone] #249
- [x] #264
- [x] #260
  - [x]  #289 
  - [x] #290 
- [x] #266 (do after 273, 249 and 260 are done)
- [x] #282
- [x] #299

Extra (related to the mapping strategy, but we decided to exclude them from the milestone):
- #262
- #291
- #273

The question is how to combine different ways of mapping dependencies to imports:
- user-defined mapping
- identity mapping
- mapping from a local environment via automatic discovery (`basepath`, `sys.path`)
- mapping added from local virtual environments via `--pyenv` option
- installing packages from PyPI to an isolated local env
- download packages from PyPI and extract import names from `toplevel.txt` or other (to be investigated)
- local cache

This sparked from various discussions and conversation in a https://github.com/tweag/FawltyDeps/pull/252.

3 rounds of visualization of the flow of this translation funnel and we have following diagram, by @jherland 
![image](https://user-images.githubusercontent.com/27200848/226679208-61e4a506-91c0-4ca6-83a3-54234b2f6ccf.png)

We should also decide what customization options should be given to a user:
- switching on/off connecting to PyPI
- use/not use/clean cache
- use/not use `sys.path`, `basepath`
- other?

Nour-Mws:
## Caches
If this is not complex enough, I'd like to also note that we can have two different types of caches: per-project and system-wide. The latter would include cached mappings from various previous runs of fawltydeps (not necessarily related to the project currently checked).

Another consideration is that this central cache can be populated not only from PyPI installations but also from FawtyDeps looking at various local Python envs in previous runs.

That may create privacy concerns so we might want to have this cache be user-based rather than system-wide?

## Order of mappings
My understanding of the diagram above is that it dictates the default order the resolver will take through mappings.

We have already established that some of those can be turned off (i.e. skipped):
- The user-provided mapping and mapping via `--pyenv` are turned off by default. 
- Identity mapping is turned on by default, but I'm partial to allowing the user to turn it off.

[...]

PR #268, opened 2023-03-23
package: Introduce a list of package resolvers

jherland:
This does not change current behavior, but does a preliminary
refactoring on the way towards having multiple package-to-import
resolvers as discussed in #256 and associated issues.

Introduce a `BasePackageResolver` abstract base class to define the
interface for package resolvers, and provide, for now, two subclasses:

- `LocalPackageResolver` (renamed from `LocalPackageLookup`)
- `IdentityMapping` (refactored from `resolve_dependencies()`)

Also refactor `resolved_dependencies()` to operate on a list of
`BasePackageResolver` objects: For each dependency name we go through the
list of resolvers until a resolver returns a `Package` object for this
name. We then add this `Package` to the returned mapping.

For now this list of resolvers is hardcoded to exactly two entries, the
`LocalPackageResolver` we were already using, and the `IdentityMapping` that
we were effectively falling back on for packages not found in the former.

Ideas for future work that build on this:
- Introduce more/different `BasePackageResolver` subclasses to do
  different kinds of package resolving. E.g. user-defined mapping from a
  config file. Resolving by looking up packages remotely on PyPI, etc.
- Make the list of resolvers user-configurable instead of hardcoded.
- Properly handle the case where none of the resolver find a package
  mapping, i.e. _unmapped_ depenedencies.

### DIFF
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -90,13 +90,15 @@ class BasePackageResolver(ABC):
     """Define the interface for doing package -> import names lookup."""
 
     @abstractmethod
-    def lookup_package(self, package_name: str) -> Optional[Package]:
-        """Convert a package name into a Package object with available imports.
+    def lookup_packages(self, package_names: Set[str]) -> Dict[str, Package]:
+        """Convert package names into a Package objects with available imports.
 
-        Return a Package object that encapsulates the package-name-to-import-
-        names mapping for the given package name.
+        Resolve as many of the given package names as possible into their
+        corresponding import names, and return a dict that maps the resolved
+        names to their corresponding Package objects.
 
-        Return None if this PackageResolver is unable to resolve the package.
+        Return an empty dict if this PackageResolver is unable to resolve any
+        of the given packages.
         """
         raise NotImplementedError
 
@@ -176,22 +178,27 @@ class LocalPackageResolver(BasePackageResolver):
 
         return self._packages
 
-    def lookup_package(self, package_name: str) -> Optional[Package]:
-        """Convert a package name to a locally available Package object.
+    def lookup_packages(self, package_names: Set[str]) -> Dict[str, Package]:
+        """Convert package names to locally available Package objects.
 
         (Although this function generally works with _all_ locally available
         packages, we apply it only to the subset that is the dependencies of
         the current project.)
 
-        Return the Package object that encapsulates the package-name-to-import-
-        names mapping for the given package name.
+        Return a dict mapping package names to the Package objects that
+        encapsulate the package-name-to-import-names mappings.
 
-        Return None if we're unable to find any import names for the given
-        package name. This is typically because the package is missing from the
-        current environment, or because we fail to determine its provided import
-        names.
+        Only return dict entries for the packages that we manage to find in the
+        local environment. Omit any packages for which we're unable to determine
+        what imports names they provide. This applies to packages that are
+        missing from the local environment, or packages where we fail to
+        determine its provided import names.
         """
-        return self.packages.get(Package.normalize_name(package_name))
+        return {
+            name: self.packages[Package.normalize_name(name)]
+            for name in package_names
+            if Package.normalize_name(name) in self.packages
+        }
 
 
 @contextmanager
@@ -223,7 +230,8 @@ class IdentityMapping(BasePackageResolver):
     name (modulo normalization, see Package.normalize_name() for details).
     """
 
-    def lookup_package(self, package_name: str) -> Optional[Package]:
+    @staticmethod
+    def lookup_package(package_name: str) -> Package:
         """Convert a package name into a Package with the same import name."""
         ret = Package(package_name)
         import_name = Package.normalize_name(package_name)
@@ -233,6 +241,10 @@ class IdentityMapping(BasePackageResolver):
             f"Assuming it can be imported as {import_name!r}."
         )
         return ret
+
+    def lookup_packages(self, package_names: Set[str]) -> Dict[str, Package]:
+        """Convert package names into Package objects w/the same import name."""
+        return {name: self.lookup_package(name) for name in package_names}
 
 
 def resolve_dependencies(
@@ -250,7 +262,7 @@ def resolve_dependencies(
     Return a dict mapping dependency names to the resolved Package objects.
     """
     # consume the iterable once
-    deps = list(dep_names)
+    deps = set(dep_names)
 
     if install_deps:
         logger.info("Installing dependencies into a new temporary Python environment.")
@@ -258,7 +270,7 @@ def resolve_dependencies(
             logger.info(
                 f"Will not use Python environment defined by --pyenv: {pyenv_path}"
             )
-        local_packages_context = temp_installed_requirements(deps)
+        local_packages_context = temp_installed_requirements(list(deps))
     else:
         local_packages_context = nullcontext(LocalPackageResolver(pyenv_path))  # type: ignore
 
@@ -271,15 +283,16 @@ def resolve_dependencies(
             local_packages,
             IdentityMapping(),
         ]
-        ret = {}
+        ret: Dict[str, Package] = {}
 
-        for name in deps:
-            if name not in ret:
-                for resolver in resolvers:
-                    package = resolver.lookup_package(name)
-                    if package is None:  # skip to next resolver
-                        continue
-                    ret[name] = package
-                    break  # skip to next dependency name
+        for resolver in resolvers:
+            unresolved = deps - ret.keys()
+            if not unresolved:  # no unresolved deps left
+                logger.debug("No deps left to resolve!")
+                break
+            logger.debug(f"Trying to resolve {unresolved!r} with {resolver}")
+            resolved = resolver.lookup_packages(unresolved)
+            logger.debug(f"  Resolved {resolved!r} with {resolver}")
+            ret.update(resolved)
 
         return ret
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -81,26 +81,6 @@ class Package:
         self.mappings.setdefault(mapping, set()).update(import_names)
         self.import_names.update(import_names)
 
-    def add_identity_import(self) -> None:
-        """Add identity mapping to this package.
-
-        This builds on an assumption that a package 'foo' installed with e.g.
-        `pip install foo`, will also provide an import name 'foo'. This
-        assumption does not always hold, but sometimes we don't have much else
-        to go on...
-        """
-        self.add_import_names(
-            self.normalize_name(self.package_name),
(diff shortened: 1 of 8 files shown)


fawltydeps/packages.py
```diff
@@ -194,32 +210,54 @@ def lookup_package(self, package_name: str) -> Optional[Package]:
         return self.packages.get(Package.normalize_name(package_name))
 
 
+class IdentityMapping(BasePackageResolver):
+    """An imperfect package resolver that assmues package name == import name.
```
mknorps, line 214: It looks like the lyrical subject has strong feelings about this mapping ;)
It is a joy to read those docstrings :D

```suggestion
    """An imperfect package resolver that assumes package name == import name.
```

fawltydeps/packages.py
```diff
@@ -194,32 +210,54 @@ def lookup_package(self, package_name: str) -> Optional[Package]:
...
     dependencies inside the virtualenv given by 'pyenv_path'. When 'pyenv_path'
     is None (the default), look for packages in the current Python environment
     (i.e. equivalent to sys.path).
 
-    For dependencies that cannot be found with LocalPackageLookup,
-    fabricate an identity mapping (a pseudo-package making available an import
-    of the same name as the package, modulo normalization).
+    For dependencies that cannot be found with LocalPackageResolver, fall back
+    an identity mapping (a pseudo-package making available an import of the same
+    name as the package, modulo normalization).
```
mknorps, line 242: I think with this refactor we do not need this sentence. Resolution order is defined in `resolvers` list.
```suggestion
```

mknorps (changes requested):
I love this refactor :heart_eyes: 
The structure of how to add the resolvers and stack them is clean and visible now.


What we should still consider is a special place of identity mapping inside `Package` object.  I think we should not include `add_identity_import` and `identity_mapping` in the `Package` object, but rather move it to `IdentityMapping`. So conceptually, `Package` is independent of any resolver.

fawltydeps/packages.py
jherland: Haha. True, but I believe I can defend that assertion: the identity mapping is _objectively_ worse (more imperfect) than the other mappings we want to use, and that is exactly why we put it at the bottom of our stack. There is no plausible scenario in which placing the identity mapping at the top of the stack would yield better mapping results.

jherland:
> What we should still consider is a special place of identity mapping inside `Package` object. I think we should not include `add_identity_import` and `identity_mapping` in the `Package` object, but rather move it to `IdentityMapping`. So conceptually, `Package` is independent of any resolver.

Fully agreed. I'll add a commit on top to do just that.

jherland:
> > What we should still consider is a special place of identity mapping inside `Package` object. I think we should not include `add_identity_import` and `identity_mapping` in the `Package` object, but rather move it to `IdentityMapping`. So conceptually, `Package` is independent of any resolver.
> 
> Fully agreed. I'll add a commit on top to do just that.

Added in e16c2d0

mknorps (approved):
:rocket:

Nour-Mws (commented):
I only read enough of this PR to know that this works on a per-dependency basis rather than on a collection of dependencies.
My question is: would we now need to create a new venv for each and every dependency in the "install_deps" mapping if these don't get matched by previous mappings?

jherland:
> I only read enough of this PR to know that this works on a per-dependency basis rather than on a collection of dependencies.
> My question is: would we now need to create a new venv for each and every dependency in the "install_deps" mapping if these don't get matched by previous mappings?

Good catch! Yes, we will need to change the API here to have each resolver handle multiple deps at the same time. I'll attempt to do this when rebasing this PR on top of #252.

One complication I'm seeing: when handling multiple deps (A, B, C) and one of them fails (e.g. C does not exist on PyPI), the resolver should to try to handle this failure somehow (to let A and B resolve successfully).

jherland:
I've now rebased on top of #252, which has brought in a bunch of new changes. Hence I'm tagging this for re-review. The good news is that this brings in more of the changes we've discussed in #256 and associated issues. Here are the commits with some short comments to guide your review:

1. ab33919 `test_sample_projects`: Print experiment settings
    - Minor quality-of-life thing that I ended up adding while debugging test failures
2. c911378 `packages`: Introduce a list of package resolvers
    -  The same commit that started this PR - in _spirit_: the rebase onto #252 made things a little more complicated...
3. bab45cd `packages`: Move id mapping logic into `IdentityMapping` class
    -  The refactoring that @mknorps proposed in her review. Not much has changed here, except for quoting the import name in the INFO log message that is printed when we use the identity mapping.
4. 1aff0ae `packages`: Change `BasePackageResolver` API: resolve all names in one call
    - The API change that I proposed in response to @Nour-Mws's comment. I'm fairly happy with how this turned out.
5. c955096 `packages`: Rewrite `temp_installed_requirements()` as a "proper" resolver
    - Refactor the context manager that @Nour-Mws added in #252 into a "proper" `BasePackageResolver` subclass, using the new API. I'm very happy with how this simplfies `resolve_dependencies()` and removes the slightly awkward management of the context managers (we no longer need `nullcontext`!).
6. 32b6e23 `packages`: Use both local + `TemporaryPipInstallResolver` at the same time
    - This is outside the original scope of this PR, but it was really easy to add on top: Instead of choosing between the "local" or the "installing" resolver, we can now use both!

fawltydeps/packages.py
```diff
@@ -180,44 +178,93 @@ def packages(self) -> Dict[str, Package]:
...
-            check=True,  # fail if any of the commands fail
+
+    @staticmethod
+    def lookup_package(package_name: str) -> Package:
+        """Convert a package name into a Package with the same import name."""
+        ret = Package(package_name)
+        import_name = Package.normalize_name(package_name)
+        ret.add_import_names(import_name, mapping=DependenciesMapping.IDENTITY)
+        logger.info(
+            f"Could not find {package_name!r} in the current environment. "
```
Nour-Mws, line 260: I think this should be reworded. Below is a lazy suggestion:
```suggestion
            f"{package_name!r} was not resolved. "
```

jherland: A little more verbose:

```suggestion
            f"Failed to resolve {package_name!r} in the given Python environments. "
```

fawltydeps/packages.py
```diff
@@ -227,41 +274,32 @@ def resolve_dependencies(
...
-                    )
-                ret[name] = package
+        resolvers += [TemporaryPipInstallResolver()]
+    resolvers += [IdentityMapping()]
+
+    ret: Dict[str, Package] = {}
+    for resolver in resolvers:
+        unresolved = deps - ret.keys()
+        if not unresolved:  # no unresolved deps left
+            logger.debug("No deps left to resolve!")
```
Nour-Mws, line 299:
```suggestion
            logger.debug("No dependencies left to resolve!")
```

tests/test_cmdline.py
```diff
@@ -465,7 +465,7 @@ def test_check__simple_project_with_extra_deps__reports_unused(
         "--check", "--detailed", "-v", f"--code={tmp_path}", f"--deps={tmp_path}"
     )
     assert output.splitlines() == expect
-    assert errors == "\n".join(expect_logs)
+    assert_unordered_equivalence(errors.splitlines(), expect_logs)
```
Nour-Mws, line 468: Why unordered?

fawltydeps/packages.py
```diff
@@ -180,44 +178,93 @@ def packages(self) -> Dict[str, Package]:
...
+
+
+class TemporaryPipInstallResolver(BasePackageResolver):
+    """Resolve packages by installing them in to a temporary venv.
 
-        Return None if we're unable to find any import names for the given
-        package name. This is typically because the package is missing from the
-        current environment, or because we fail to determine its provided import
-        names.
+    This provide a resolver for packages that are not installed in an existing
```
Nour-Mws, line 207:
```suggestion
    This provides a resolver for packages that are not installed in an existing
```

Nour-Mws (approved):
Looks good and it was a great read :)
Feel free to resolve conversations.

fawltydeps/packages.py
Nour-Mws: My problem with the above:
- When a user mapping is provided, it's no longer just an issue of failing to resolve in a Python env.
- When identity mapping is the only mapping, we simply cannot say that we "failed" to resolve.

Again, this is for later, so we can open an issue to remember to review this log message when we're done with the mapping milestone.

tests/test_cmdline.py
jherland: Generally, I feel that our `test_cmdline` tests are getting more and more in the way of needed refactoring. One of the problems is that we have too many tests that - in addition to verifying the behavior described in the test name - also check for the existence and ordering of log messages produced by our code.

Specifically, in this case, since we now resolve _sets_ of dependency names at a time, the order in which we process each individual dependency name is no loger deterministic. Thus these two INFO log messages can show up in any order:
- `Could not find 'requests' in the current environment. Assuming it can be imported as 'requests'.`
- `Could not find 'pandas' in the current environment. Assuming it can be imported as 'pandas'.`

(There is more of this coming as part of the traverse refactoring that I talked about in today's meeting...)

fawltydeps/packages.py
```diff
@@ -227,41 +274,32 @@ def resolve_dependencies(
...
-                package = local_packages.lookup_package(name)
-                if package is None:  # fall back to identity mapping
-                    package = Package.identity_mapping(name)
-                    logger.info(
-                        f"Could not find {name!r} in the current environment. Assuming "
-                        f"it can be imported as {', '.join(sorted(package.import_names))}"
-                    )
-                ret[name] = package
+        resolvers += [TemporaryPipInstallResolver()]
+    resolvers += [IdentityMapping()]
```
mknorps, line 293: I would add a TODO-warning here:
```suggestion
    # Identity mapping being at the bottom of the resolvers stack
    # ensures that _all_ deps are matched. If you consider to make 
    # identity mapping optional, remember about handling unresolved dependcies.
    resolvers += [IdentityMapping()]
```

mknorps:
I really like this refactor and I am eagerly waiting for it to be merged  :100: 

One thing I noticed in rereading the changes is that there are no tests for `TemporaryPipInstallResolver`. You may consider adding it in the follow-up PR, or just create some dummy TODO test in `test_packages.py`, so it does not skip our attention. This can also be a good first issue :)

fawltydeps/packages.py
jherland:
```suggestion
    # Identity mapping being at the bottom of the resolvers stack ensures that
    # _all_ deps are matched. TODO: If we make the identity mapping optional,
    # we must remember to properly handle/signal unresolved dependencies.
    resolvers += [IdentityMapping()]
```

PR #269, opened 2023-03-23
Add the code of conduct

mknorps:
Following the example of pandas, in this PR the code of conduct is added. The document is strictly based on https://www.contributor-covenant.org/. The statements are along the lines "do no evil" and are quite universal.

We need to agree on how to handle violations of the code of conduct, especially in `Enforcement` section - supply contact methods. I suggest our tweag emails unless you have some other idea.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,132 @@
+# Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual
+identity and orientation.
+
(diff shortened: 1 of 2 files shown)


jherland:
I gave this a quick read, and it seems very well considered and written. Thanks for adding it.

Agree that it makes sense to use our Tweag email addresses in the Enforcement section.

One minor technicality: Do we add this at the top level, or under `docs/`? We are also thinking of splitting up the `README.md` into multiple files, and it might be time to decide on how to organize this (that said, it's also easy to move files around in the future).

mknorps:
> One minor technicality: Do we add this at the top level, or under `docs/`? We are also thinking of splitting up the `README.md` into multiple files, and it might be time to decide on how to organize this (that said, it's also easy to move files around in the future).

I thought about adding this at the top level, like in [click](https://github.com/pallets/click) library along with CONTRIBUTING.md.

jherland:
> I thought about adding this at the top level, like in [click](https://github.com/pallets/click) library along with CONTRIBUTING.md.

Makes sense. The next thing to split out of the `README.md` is probably the FAQ section. Maybe that makes more sense under `docs/`?

jherland (approved):

PR #271, opened 2023-03-24
Tests refactor: running fawltydeps on 'main' instead of in subprocess in command line tests

mknorps:
Tests refactor aiming at speeding up the tests and reducing some unnecessarily tested logs equalities, that may end up slowing down refactors of the code.

Things improved:
- fixed bug in `print_output` function, where one of the writers were still using `sys.stdout`.
- introduce `ProjectTestsVector` and parameterized some of the tests in `tests_cmdline`
- introduce `run_fawltydeps_function` that uses `main` function instead of subprocesses to run FawltyDeps. Tests with using this function are simpler, we do not collect `stderr` and only compare output and exit code.
- rewrite some of tests in `test_cmdine` to use the above function to reduce unnecessary checks of log messages,

## Experiments
From a simple check below, we see that there is some improvement in the execution time of tests suites. There are some more gains, if we use `run_fawltydeps_function` instead of `run_faltydeps` more extensively. But for this PR it is enough :)
**Current**
```
time nox -s tests
...
real	2m11.658s
user	1m55.989s
sys	0m7.951s

```

```
time nox -Rs tests
...
real	1m39.510s
user	1m33.693s
sys	0m5.844s
```

**Previous**
```
time nox -s tests
...
real	2m22.484s
user	2m5.909s
sys	0m8.902s
```

```
time nox -Rs tests
...
real	1m58.316s
user	1m51.029s
sys	0m7.309s

```

### DIFF
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -16,7 +16,7 @@ from fawltydeps.main import UNUSED_DEPS_OUTPUT_PREFIX, VERBOSE_PROMPT, Analysis,
 from fawltydeps.types import Location, UnusedDependency
 
 from .test_extract_imports_simple import generate_notebook
-from .utils import assert_unordered_equivalence, run_fawltydeps
+from .utils import assert_unordered_equivalence, run_fawltydeps, run_fawltydeps_function
 
 logger = logging.getLogger(__name__)
 
@@ -152,7 +152,7 @@ def test_list_imports_json__from_py_file__prints_imports_from_file(write_tmp_fil
         "unused_deps": None,
         "version": version(),
     }
-    output, _errors, returncode = run_fawltydeps(
+    output, returncode = run_fawltydeps_function(
         "--list-imports", "--json", f"--code={tmp_path}/myfile.py"
     )
     assert json.loads(output) == expect
@@ -333,7 +333,7 @@ def test_list_deps_json__dir__prints_deps_from_requirements_txt(
         "unused_deps": None,
         "version": version(),
     }
-    output, _errors, returncode = run_fawltydeps(
+    output, returncode = run_fawltydeps_function(
         "--list-deps", "--json", f"--deps={tmp_path}"
     )
     assert json.loads(output) == expect
@@ -605,7 +605,7 @@ def test_check_json__simple_project__can_report_both_undeclared_and_unused(
         ],
         "version": version(),
     }
-    output, _errors, returncode = run_fawltydeps(
+    output, returncode = run_fawltydeps_function(
         "--check", "--json", f"--code={tmp_path}", f"--deps={tmp_path}"
     )
     assert json.loads(output) == expect
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -637,7 +637,6 @@ def test_check_unused__simple_project__reports_only_unused(
     )
 
     expect = [
-        "",
         "These dependencies appear to be unused (i.e. not imported):",
         "- 'pandas' declared in:",
         f"    {tmp_path / 'requirements.txt'}",
@@ -663,7 +662,6 @@ def test__no_action__defaults_to_check_action(
     )
 
     expect = [
-        "",
         "These imports appear to be undeclared dependencies:",
         "- 'requests' imported at:",
         f"    {tmp_path / 'code.py'}:1",
@@ -688,7 +686,6 @@ def test__no_options__defaults_to_check_action_in_current_dir(
     )
 
     expect = [
-        "",
         "These imports appear to be undeclared dependencies:",
         "- 'requests' imported at:",
         f"    {tmp_path / 'code.py'}:1",
@@ -697,7 +694,7 @@ def test__no_options__defaults_to_check_action_in_current_dir(
         "- 'pandas' declared in:",
         f"    {tmp_path / 'requirements.txt'}",
     ]
-    output, returncode = run_fawltydeps_function("--detailed", "-v", cwd=tmp_path)
+    output, returncode = run_fawltydeps_function("--detailed", "-v", basepath=tmp_path)
     assert output.splitlines() == expect
     assert returncode == 3
 
@@ -740,7 +737,6 @@ def test_check__simple_project_in_fake_venv__resolves_imports_vs_deps(
         "--detailed", f"--code={tmp_path}", f"--deps={tmp_path}", f"--pyenv={venv_dir}"
     )
     assert output.splitlines() == [
-        "",
         Analysis.success_message(check_undeclared=True, check_unused=True),
     ]
     assert returncode == 0
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -5,10 +5,12 @@ overall behavior of the command line interface, rather than testing our
 core exhaustively (which is what the other unit tests are for.
 """
 
+from dataclasses import dataclass, field
 import json
 import logging
 from itertools import dropwhile
 from textwrap import dedent
+from typing import List
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


tests/test_cmdline.py
```diff
@@ -399,166 +396,154 @@ def test_list_deps__pick_multiple_listed_files__prints_all_dependencies(
...
+    # Order of output is determined, as we use alphabetical ordering.
+    assert output.splitlines() == [
+        v.format(path=tmp_path) for v in vector.expect_output
+    ]
+    # We do set comparison here, as the evaluation order of
+    # input is not determined. In the process of evaluation,
+    # we use sets (like in IdentityMapping fallback).
+    assert {l for l in logs.split("\n") if l != ""} == {
+        v.format(path=tmp_path) for v in vector.expect_logs
+    }
```
Nour-Mws, line 545: It might be relevant that the test that is [failing the CI](https://github.com/tweag/FawltyDeps/actions/runs/4615271918/jobs/8158990052) in the commit before the last one is `tests/test_cmdline.py::test_check_undeclared_and_unused[simple_project_with_extra_deps__reports_unused]`, which has a different log assertion from the other similar tests:
```python
assert_unordered_equivalence(errors.splitlines(), expect_logs)
```
Compared to:
```
assert errors == "\n".join(expect_logs)
```

So this unordered set of logs might have been a concern before and had led to this difference.

Do you think this will still work if you change the set comparison you provide above with `assert_unordered_equivalence`?

tests/test_cmdline.py
Nour-Mws, line 1: I don't understand why in some tests we're switching to `run_fawltydeps_function` while we continue using `run_fawltydeps` in other tests. What's the reason for that?

Nour-Mws (approved):
I am very happy with what this PR does, especially the refactoring in `test_cmdline`. Great work :)

tests/test_cmdline.py
mknorps: Tests, where we want to check some logs, are left with `run_fawltydeps`.
Tests, where logs are repeatedly tested and have a test for a particular log message already tested in other tests (as with this bunch of `requests` & `pandas` variations) are changed to `run_fawltydeps_function` for decoupling and faster runs.

tests/test_cmdline.py
mknorps: Set comparison is not the only reason why this like looks like this. There is also filtering of the empty lines on the LHS  and formatting of the expected output string.

tests/test_cmdline.py
Nour-Mws: In that case, wouldn't it make sense to have `run_fawltydeps_function` output logs as well, and switch all tests away from subprocess? We could still ignore the logs in the cases where we're not interested in checking their values.

mknorps: That would be a perfect solution, which I planned to implement first. 
The problem is that I could not make the [log capture ](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stderr) function work properly in tests.
We could do a more explicit logging handling in the main function as we do for stdout, but this would require a bit more refactor I think :thinking:

tests/test_cmdline.py
```diff
@@ -652,16 +637,12 @@ def test_check_unused__simple_project__reports_only_unused(
...
         "- 'pandas' declared in:",
         f"    {tmp_path / 'requirements.txt'}",
     ]
-    expect_logs = [
-        f"INFO:fawltydeps.extract_imports:Parsing Python files under {tmp_path}",
-        "INFO:fawltydeps.packages:'pandas' was not resolved."
-        " Assuming it can be imported as 'pandas'.",
-    ]
-    output, errors, returncode = run_fawltydeps(
+    output, returncode = run_fawltydeps_function(
```
jherland, line 645: I suspect the extra empty line is needed because `run_fawltydeps()` performs a `.strip()` on the captured output where `run_fawltydeps_function()` does not?

tests/test_cmdline.py
```diff
@@ -713,22 +688,17 @@ def test__no_options__defaults_to_check_action_in_current_dir(
...
         "",
         "These dependencies appear to be unused (i.e. not imported):",
         "- 'pandas' declared in:",
-        "    requirements.txt",
-    ]
-    expect_logs = [
-        "INFO:fawltydeps.extract_imports:Parsing Python files under .",
-        "INFO:fawltydeps.packages:'pandas' was not resolved."
-        " Assuming it can be imported as 'pandas'.",
+        f"    {tmp_path / 'requirements.txt'}",
```
jherland, line 698: I'm a little worried about these changes. As I understand it, where `run_fawltydeps()` can use the `cwd` to actually set the current working directory for the subprocess, `run_fawltydeps_functon()` cannot do the same, since everything happens in the same process. Instead, you work around that by passing `cwd` as a `basepath`/positional argument instead, and the consequence of that is to adjust the expected output here.

I would like to keep test coverage of the fact that when we work on paths relative to cwd, then that is reflected in the output.

I wonder if we can use [`monkeypatch.chdir`](https://docs.pytest.org/en/latest/reference/reference.html#pytest.MonkeyPatch.chdir) to temporarily change cwd while running the tests? However, it probably means changing `run_fawltydeps_function()` into a pytest fixture (to have access to `monkeypatch`).

Alternatively, kick this to a later PR, and leave it as `run_fawltydeps()` instead of `run_fawltydeps_function()` for now.

tests/utils.py
```diff
@@ -95,6 +98,35 @@ def run_fawltydeps(
...
 
+def run_fawltydeps_function(
+    *args: str,
+    config_file: Path = Path("/dev/null"),
+    to_stdin: Optional[str] = None,
+    cwd: Optional[Path] = None,
+) -> Tuple[str, int]:
+    """Run FawltyDeps with `main` function. Designed for unit tests.
+
+    Ignores logging output and returns stdout and the exti code
```
jherland, line 109:
```suggestion
    Ignores logging output and returns stdout and the exit code
```

tests/utils.py
```diff
@@ -95,6 +98,35 @@ def run_fawltydeps(
     return proc.stdout.strip(), proc.stderr.strip(), proc.returncode
 
 
+def run_fawltydeps_function(
+    *args: str,
+    config_file: Path = Path("/dev/null"),
+    to_stdin: Optional[str] = None,
+    cwd: Optional[Path] = None,
```
jherland, line 105: I understand the impulse to reuse the same API as for `run_fawltydeps()`, but if we're not actually going to change the current working directory here (see my `monkeypatch.chdir` suggestion above), then this should be renamed so as to not mislead the caller:
```suggestion
    basepath: Optional[Path] = None,
```

jherland (commented):
Love these improvements, but I think there's a question about how `run_fawltydeps_function()` deals with `cwd`...

tests/test_cmdline.py
mknorps: :+1:  https://github.com/tweag/FawltyDeps/issues/301
(there is no "salute" emoji :cry: )

tests/test_cmdline.py
```diff
@@ -7,16 +7,18 @@
...
+from typing import List
 
 import pytest
 
 from fawltydeps.main import UNUSED_DEPS_OUTPUT_PREFIX, VERBOSE_PROMPT, Analysis, version
 from fawltydeps.types import Location, UnusedDependency
 
 from .test_extract_imports_simple import generate_notebook
-from .utils import assert_unordered_equivalence, run_fawltydeps
+from .utils import assert_unordered_equivalence, run_fawltydeps, run_fawltydeps_function
```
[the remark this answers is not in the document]
mknorps: Keeping this in mind. Will post it just after this one is done, not to obfuscate the current changes.

tests/test_cmdline.py
```diff
@@ -399,166 +396,154 @@ def test_list_deps__pick_multiple_listed_files__prints_all_dependencies(
...
-    ]
-    output, errors, returncode = run_fawltydeps(
-        "--check", "--detailed", "-v", f"--code={tmp_path}", f"--deps={tmp_path}"
-    )
-    assert output.splitlines() == expect
-    assert errors == "\n".join(expect_logs)
-    assert returncode == 3
 
+full_success_message = (
+    Analysis.success_message(check_undeclared=True, check_unused=True) or ""
```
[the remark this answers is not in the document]
mknorps: yep...

tests/test_cmdline.py
mknorps: Yes! :bowing_woman: 
Adding `.strip()` to the latter.

tests/test_cmdline.py
mknorps: Created #302 not to forget.

PR #272, opened 2023-03-27
deps: isort (5.11.4 -> 5.11.5)

janydoe:
These changes fix the following issue:

```
Processing /build/isort-5.11.4
  Running command Preparing metadata (pyproject.toml)
  Traceback (most recent call last):
    File "/nix/store/5shvyhxgc5km8iljl9xh5sfnmwbw3kas-python3.10-pip-23.0.1/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
      main()
    File "/nix/store/5shvyhxgc5km8iljl9xh5sfnmwbw3kas-python3.10-pip-23.0.1/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/nix/store/5shvyhxgc5km8iljl9xh5sfnmwbw3kas-python3.10-pip-23.0.1/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 149, in prepare_metadata_for_build_wheel
      return hook(metadata_directory, config_settings)
    File "/nix/store/xx78pfds0r7150fczmswvv2wnvif7wfp-python3.10-poetry-core-1.5.2/lib/python3.10/site-packages/poetry/core/masonry/api.py", line 40, in prepare_metadata_for_build_wheel
      poetry = Factory().create_poetry(Path(".").resolve(), with_groups=False)
    File "/nix/store/xx78pfds0r7150fczmswvv2wnvif7wfp-python3.10-poetry-core-1.5.2/lib/python3.10/site-packages/poetry/core/factory.py", line 57, in create_poetry
      raise RuntimeError("The Poetry configuration is invalid:\n" + message)
  RuntimeError: The Poetry configuration is invalid:
    - [extras.pipfile_deprecated_finder.2] 'pip-shims<=0.3.4' does not match '^[a-zA-Z-_.0-9]+$'
```


jherland:
I'm trying to understand how you got the original error message, as I have not yet been able to reproduce the issue locally. 

Also, even if the poetry.lock is autogenerated, the PR does introduce a _lot_ of churn in an otherwise quite "stable" file. Which command did you use to generate the new `poetry.lock`?

jherland:
I managed to produce the ~same diff with `poetry update isort` and `poetry lock --no-update` on a newer version of Poetry (1.4.1, and poetry-core 1.5.2), so the churn in the diff is likely caused by the newer Poetry version. So I don't have a problem with the PR per se, but still unsure how to reproduce the initial error message.

For reference, here is a link to the upstream issue in isort: https://github.com/PyCQA/isort/commit/b3d35702cf92cdeff5b46cc3719f9661c8c91bb2

The issue I'm looking at now is that rerunning `poetry lock --no-update` with the older version of Poetry (1.2.2) in our nix-shell seems to get stuck in an infinite loop. However, that issue seems to be there even without this PR applied, so we clearly have some problems with our Poetry setup... 😒 

Another semi-related issue is that [isort v5.12.0 drops support for Python 3.7](https://github.com/PyCQA/isort/blob/main/CHANGELOG.md#5120-january-28-2023) which means that we should probably pin `isort` at `<5.12.0` until we are ready to drop v3.7 support ourselves.

jherland:
@janydoe: Thanks again for submitting this. I rephrased your commit in #279, but retained your authorship. Hope this is OK with you? (In short I removed the collateral `poetry.lock` changes that seem to cause older Poetry versions to loop indefinitely when running `poetry lock`.)

I've confirmed that your Nix expression above also passes with #279.

PR #275, opened 2023-03-28
Preparing for the traversal refactoring

jherland:
While working on the project traversal refactoring that I talked about yesterday,
these commits that do not strictly need to be part of that work has settled down
to the bottom of my branch. They deserve to be reviewed/considered separately before
the rest of the (still-ongoing) traversal work is submitted.

Commits:
- `test_extract_imports_simple`: Minor inconsequential fixes
- `sample_projects`: Add empty project
- `sample_projects`: Add mixed project
- `main`: Refactor `Analysis` object to remove annoying asserts

### DIFF
--- a/fawltydeps/main.py
+++ b/fawltydeps/main.py
@@ -14,10 +14,9 @@ import argparse
 import json
 import logging
 import sys
-from dataclasses import dataclass
-from functools import partial
+from functools import partial, wraps
 from operator import attrgetter
-from typing import Dict, List, Optional, TextIO
+from typing import Callable, Dict, List, Optional, TextIO, TypeVar
 
 from pydantic.json import custom_pydantic_encoder  # pylint: disable=no-name-in-module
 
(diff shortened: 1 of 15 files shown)


tests/test_cmdline_options.py
```diff
@@ -145,7 +145,7 @@ def cli_arguments_combinations(draw):
     # to point to a parser that we want fawltydeps to use.
     # Following code ensures that we do not encounter trying to match
     # explicit file with incorrect parser (square in the oval-shaped hole problem)
-    drawn_deps = draw(deps_option_strategy(dep_paths))
+    drawn_deps = [] if not dep_paths else draw(deps_option_strategy(dep_paths))
```
mknorps, line 148: The idea with the drawing from `deps_option_strategy` with additional `map` function should be enough to handle empty list case. Was there some problem with this parsing? The `code_option_strategy` should be symmetrical.

mknorps (changes requested):
I love the idea of `calculated_once` :heart: 
Using the underscore makes it quite readable.
I somehow feel that we diverge from mainstream Python but I have nothing against it :) 

Also a very good idea to add a "null project" :clap: 
Let's see how "mixed project" will be used :)

What would be good to see here is a test that checks that the properties that should be calculated once indeed are.

fawltydeps/main.py
```diff
@@ -99,44 +162,17 @@ def create(cls, settings: Settings, stdin: Optional[TextIO] = None) -> "Analysis
...
-            ret.resolved_deps = resolve_dependencies(
-                (dep.name for dep in ret.declared_deps),
-                pyenv_path=settings.pyenv,
-                install_deps=settings.install_deps,
-            )
+        ret = cls(settings, stdin)
 
+        # Realize only the above properties needed to satisfy settings.actions:
+        if ret.is_enabled(Action.LIST_IMPORTS):
+            ret.imports  # pylint: disable=pointless-statement
```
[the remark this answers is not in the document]
jherland: 🤷🏻‍♂️

tests/test_cmdline_options.py
jherland: Without this line, this test case fails with:
```
hypothesis.errors.InvalidArgument: Cannot sample from a length-zero sequence.
```
I'm not exactly sure why, but the traceback from Hypothesis does indicate that `draw()` is not happy with being passed an empty sequence. I first thought the following line further down would the also be problematic:
```python
    code = draw(code_option_strategy(code_paths))
```
but it isn't and I guess that's because `code_paths` always contains at least the STDIN option, and can therefore never be empty.

jherland:
> What would be good to see here is a test that checks that the properties that should be calculated once indeed are.

Added a test for `@calculated_once` in 99685eb. Between this and the `--json` tests in `test_cmdline` (already verifying that `Analysis` members are _not_ computed unless required) I feel we have covered this fairly well.

The only thing we aren't explicitly testing is that the `Analysis` members - when required - are only calculated _once_, but I find it hard to test this on the actual `Analysis` object without doing some surgery on it to make it less [idempotent](https://en.wikipedia.org/wiki/Idempotence) (so that multple calculations are observable).

What do you think?

tests/test_cmdline_options.py
mknorps: I added the fixup to this in 03af92f7960850ecab3e95605c8f8e76ae0676c9. We want to keep the generation logic as close to the strategy definition as possible.

The problem you had looks like the `list` strategy could not use an empty list as an input. That is why I added a guard for an empty input to the `code` and `deps` strategies.

mknorps:
> Added a test for `@calculated_once` in [99685eb](https://github.com/tweag/FawltyDeps/commit/99685eb983c6b8c15dfe026e9c066b254ccccd4c). Between this and the `--json` tests in `test_cmdline` (already verifying that `Analysis` members are _not_ computed unless required) I feel we have covered this fairly well.



> > What would be good to see here is a test that checks that the properties that should be calculated once indeed are.
> 
> Added a test for `@calculated_once` in [99685eb](https://github.com/tweag/FawltyDeps/commit/99685eb983c6b8c15dfe026e9c066b254ccccd4c). Between this and the `--json` tests in `test_cmdline` (already verifying that `Analysis` members are _not_ computed unless required) I feel we have covered this fairly well.
> 

Yes, they are great!


> The only thing we aren't explicitly testing is that the `Analysis` members - when required - are only calculated _once_, but I find it hard to test this on the actual `Analysis` object without doing some surgery on it to make it less [idempotent](https://en.wikipedia.org/wiki/Idempotence) (so that multple calculations are observable).
> 
> What do you think?

I think that we still plan to do the `Analysis` refactor and one of the reasons in the testability of this object. So I would not spend much time doing elaborate and complicated tests. We know that `calculated_once` works as it should, so it should be enough.

mknorps (approved):
I think this is ready to merge :rocket: 

Let the _traversing_ begin ;)

PR #279, opened 2023-03-28
Update isort to v5.11.5, but no further

jherland:
This supersedes #272, to fix using FawltyDeps with newer versions of Poetry (poetry-core>=1.5.0).

Thanks to @janydoe for submitting #272. This PR rephrases their commit to only include the `poetry.lock`
changes directly associated with the `isort` upgrade, but omitting collateral changes caused by the newer
Poetry version. AFAICS, these collateral changes cause an older Poetry version (e.g. the v1.2.2 version
used by our `shell.nix`) to loop forever when running `poetry lock --no-udpate`.

I also add another commit preventing `isort` from upgrading to v5.12.0, which is the version where isort
drop support for Python v3.7.

Commits:
- `poetry.lock`: Update `isort` (5.11.4 -> 5.11.5)
- `pyproject.toml`: Don't upgrade `isort` to 5.12.0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,8 @@ fawltydeps = "fawltydeps.main:main"
 # Do not add anything here that is only needed by CI/tests/linters/developers
 python = "^3.7.2"
 importlib_metadata = "^6.0.0"
-isort = "^5.10"
+# isort 5.12.0 drops support for Python v3.7:
+isort = ">=5.10,<5.12.0"
 pip-requirements-parser = "^32.0.1"
 pydantic = "^1.10.4"
 tomli = {version = "^2.0.1", python = "<3.11"}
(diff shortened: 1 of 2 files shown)


mknorps (approved):
:rocket: 

We will be upgrading poetry with the upgrade of the nix packages we are using. So I agree that minimal change is the best now.

pyproject.toml
```diff
@@ -70,7 +71,7 @@ optional = true
 [tool.poetry.group.format.dependencies]
 black = "^22"
 colorama = "^0.4.6"
-isort = "^5.10"
+isort = "<5.12.0"
```
Nour-Mws, line 74: Thinking out loud here: this translates to "any isort version lower than 5.12.0" while the old specification was "any version >= 5.10 and <5.11".
I am not sure how we initially decided on 5.10, but it seems reasonable to keep that lower bound because older version might not support the use we make of isort in our code?

jherland: Added 5.10 as a lower bound in c3cf723.

PR #280, opened 2023-03-29
Minor Nox/Pyproject fixes

jherland:
Commits:
- `noxfile.py`: Install FD in editable mode in Nox sessions
- `pyproject.toml`: Remove redundant `isort` from our `dev` dependencies


Nour-Mws:
So that me and Maria's questions form a complete set over your PR: what caused the switch to editable mode? :D

jherland:
@mknorps:
> Why is isort no longer needed in the dev environment?

`isort` is still needed and present in the dev environment, because the dev environment is always installed _together_ with FD's main dependencies. (Contrast: `poetry install --sync --with=dev` in our `shell.nix` vs something like `poetry install --sync --only=dev` which I'd argue does not make sense.)

The only thing fixed in this PR is that `isort` is not explicitly mentioned as _both_ a main dependency and a dev dependency (which is redundant, AFAICS). Note that none of our other main dependencies are redundantly listed as dev dependencies.

@Nour-Mws:
> what caused the switch to editable mode?

This is a small annoyance I've had for a long time without properly looking into it until now: When using `-R` to speed up my Nox runs (the intention being to reuse the venv from the last run and avoiding to re-install the same, unchanged dependencies), I would experience frequent test failures that go away when I skip the `-R` argument. Typically it would go something like this:

1. Write a test case for some new code.
2. Run the test suite (`nox -s tests-3.10 -- -vv`).
3. Test suite fails on a relevant test case.
4. Fix the failure somewhere in FD.
5. Re-run the test suite w/`-R` (since we haven't changed out deps): `nox -Rs tests-3.10 -- -vv`.
6. Test still fails in the same way as before the fix.
7. Re-run the test suite without `-R`.
8. Test suite passes.

When we run our test suite with Nox (e.g. in step 2, without `-R`) Nox creates a venv, installs the stuff listed under `[tool.poetry.group.test.dependencies]`, and then installs FD (w/deps) into this venv. This,_installed_ version of FD is the one that actually ends up being tested[^1]. When this install is _not_ editable, it does not point directly back to our work tree, so any changes we make in our work tree are not automatically reflected in the version being tested unless we re-install FD into the venv.

Running with `-R` (step 5) skips this re-install, so we end up testing _without_ the changes we made in step 4.[^2]

By making the FD install editable, it _always_ points directly to the code in our worktree, and we avoid this hybrid situation where we're not testing what actually is in our worktree.

With this the `-R` option should behave much more predictably/correctly, and it should now be safe to use as long as our dependencies have not changed since the last run.

[^1]: Nox ends up running `pytest` in our worktree, which ostensibly _should_ test the code in our worktree, but when our test scripts do `from fawltydeps import ...`, the `fawltydeps` package they end up referencing is the one in the venv, and not the one in the worktree (`../fawltydeps/` relative to `tests/`).

[^2]: Technically, we end up running _with_ any `tests/` changes we made in step 4 (since `pytest` runs our test scripts from the worktree), but _without_ any `fawltydeps/` changes we made in step 4 (since the test scripts import the installed version of FD). This explains the problems I observed, both why they seemed intermittent (things would work when my changes only touched `tests/`), and why they sometimes did not make sense at all (when my changes spanned both `tests/` and `fawltydeps/`).

Nour-Mws:
Thanks for the detailed explanation @jherland!

PR #292, opened 2023-04-04
User-defined mapping: add option to read mapping from configuration file

mknorps:
Follow-up on #287 .

In this PR an option to read the mapping directly from the configuration file (pyproject.toml). This is a useful option when we have only a few mapping entries we want to adjust.

What changed:
- in `Settings`, field `custom_mapping` was renamed to `custom_mapping_file` (and the related elements accordingly). This field keeps a path to mapping file
- in `Settings` a new field `custom_mapping` was added, which is a dictionary of packages and list of modules they export.
- `UserDefinedMapping` is using now both, the file and the custom mapping in form of a dictionary. If both the mapping from file, and the custom mapping contain the same key (package), the mapping from pyproject.toml is taken. This is because  pyproject.toml is usually better maintained, with smaller chance of getting stale.
- refactored `setup_fawltydeps_config` to be able to add a subsection of fawltydeps configuration
- added `test_configured_run.py` to the test suite to have a place where we test how fawltydeps runs on configuration files only.

### DIFF
--- a/fawltydeps/main.py
+++ b/fawltydeps/main.py
@@ -82,7 +82,8 @@ class Analysis:  # pylint: disable=too-many-instance-attributes
         """The resolved mapping of dependency names to provided import names."""
         return resolve_dependencies(
             (dep.name for dep in self.declared_deps),
-            custom_mapping_path=self.settings.custom_mapping,
+            custom_mapping_path=self.settings.custom_mapping_file,
+            custom_mapping=self.settings.custom_mapping,
             pyenv_path=self.settings.pyenv,
             install_deps=self.settings.install_deps,
         )
--- a/fawltydeps/main.py
+++ b/fawltydeps/main.py
@@ -82,7 +82,7 @@ class Analysis:  # pylint: disable=too-many-instance-attributes
         """The resolved mapping of dependency names to provided import names."""
         return resolve_dependencies(
             (dep.name for dep in self.declared_deps),
-            custom_mapping_path=self.settings.custom_mapping_file,
+            custom_mapping_file=self.settings.custom_mapping_file,
             custom_mapping=self.settings.custom_mapping,
             pyenv_path=self.settings.pyenv,
             install_deps=self.settings.install_deps,
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -136,26 +136,44 @@ class UserDefinedMapping(BasePackageResolver):
         This enumerates the available packages  _once_, and caches the result for
         the remainder of this object's life in _packages.
         """
-        user_mapping: TomlData = {}
+        custom_mapping_from_file: Dict[str, List[str]] = {}
 
         if self.mapping_path is not None:
             logger.debug(f"Loading user-defined mapping from {self.mapping_path}")
             with open(self.mapping_path, "rb") as mapping_file:
-                toml_data = tomllib.load(mapping_file)
-                user_mapping = {**user_mapping, **toml_data}
+                custom_mapping_from_file = tomllib.load(mapping_file)
 
-        if self.custom_mapping is not None:
-            logger.debug("Loading user-defined mapping from custom mapping.")
-            # Custom mapping overrides mapping from file
-            user_mapping = {**user_mapping, **self.custom_mapping}
-
-        return {
+        mapping = {
             Package.normalize_name(name): Package(
                 name,
                 {DependenciesMapping.USER_DEFINED: set(imports)},
             )
-            for name, imports in user_mapping.items()
+            for name, imports in custom_mapping_from_file.items()
         }
+
+        if self.custom_mapping is not None:
+            logger.debug("Loading user-defined mapping from custom mapping.")
+
+            for name, imports in self.custom_mapping.items():
+                normalised_name = Package.normalize_name(name)
+                if normalised_name in mapping:
+                    logger.info(
+                        "Mapping for %s already found in %s. Import names "
+                        "from the configuration file are appended to ones "
+                        "found in the mapping file.",
+                        normalised_name,
+                        self.mapping_path,
+                    )
+                    mapping[normalised_name].add_import_names(
+                        *imports, mapping=DependenciesMapping.USER_DEFINED
+                    )
+                else:
+                    mapping[normalised_name] = Package(
+                        name,
+                        {DependenciesMapping.USER_DEFINED: set(imports)},
+                    )
+
+        return mapping
 
     def lookup_packages(self, package_names: Set[str]) -> Dict[str, Package]:
         """Convert package names to locally available Package objects."""
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -131,7 +131,11 @@ class UserDefinedMapping(BasePackageResolver):
     @property
     @calculated_once
     def packages(self) -> Dict[str, Package]:
-        """Gather mapping from a mapping file in a given path.
+        """Gather a custom mapping given by a user.
+
+        Mapping may come from two sources:
+        * _mapping_path_ which is a file in a given path, which is parsed to a ditionary
+        * _custom_mapping_ which is a dictionary of package to imports mapping.
 
         This enumerates the available packages  _once_, and caches the result for
         the remainder of this object's life in _packages.
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -24,7 +24,7 @@ from importlib_metadata import (
     _top_level_inferred,
 )
 
-from fawltydeps.types import UnparseablePathException
+from fawltydeps.types import TomlData, UnparseablePathException
 from fawltydeps.utils import calculated_once, hide_dataclass_fields
 
 if sys.version_info >= (3, 11):
@@ -114,12 +114,17 @@ class BasePackageResolver(ABC):
 class UserDefinedMapping(BasePackageResolver):
     """Use user-defined mapping loaded from a toml file"""
 
-    def __init__(self, mapping_path: Path) -> None:
+    def __init__(
+        self,
+        mapping_path: Optional[Path] = None,
+        custom_mapping: Optional[TomlData] = None,
+    ) -> None:
         self.mapping_path = mapping_path
-        if not self.mapping_path.is_file():
+        if self.mapping_path and not self.mapping_path.is_file():
             raise UnparseablePathException(
                 ctx="Given mapping path is not a file.", path=self.mapping_path
             )
+        self.custom_mapping = custom_mapping
         # We enumerate packages declared in the mapping _once_ and cache the result here:
         self._packages: Optional[Dict[str, Package]] = None
 
@@ -131,10 +136,18 @@ class UserDefinedMapping(BasePackageResolver):
         This enumerates the available packages  _once_, and caches the result for
         the remainder of this object's life in _packages.
         """
+        user_mapping: TomlData = {}
 
-        logger.debug(f"Loading user-defined mapping from {self.mapping_path}")
-        with open(self.mapping_path, "rb") as mapping_file:
-            user_mapping = tomllib.load(mapping_file)
+        if self.mapping_path is not None:
+            logger.debug(f"Loading user-defined mapping from {self.mapping_path}")
+            with open(self.mapping_path, "rb") as mapping_file:
+                toml_data = tomllib.load(mapping_file)
+                user_mapping = {**user_mapping, **toml_data}
+
+        if self.custom_mapping is not None:
+            logger.debug("Loading user-defined mapping from custom mapping.")
+            # Custom mapping overrides mapping from file
+            user_mapping = {**user_mapping, **self.custom_mapping}
 
         return {
             Package.normalize_name(name): Package(
@@ -320,6 +333,7 @@ class IdentityMapping(BasePackageResolver):
 def resolve_dependencies(
     dep_names: Iterable[str],
     custom_mapping_path: Optional[Path] = None,
+    custom_mapping: Optional[Dict[str, List[str]]] = None,
     pyenv_path: Optional[Path] = None,
     install_deps: bool = False,
 ) -> Dict[str, Package]:
@@ -339,8 +353,12 @@ def resolve_dependencies(
     # each resolver in order until one of them returns a Package object. At
     # that point we are happy, and don't consult any of the later resolvers.
     resolvers: List[BasePackageResolver] = []
-    if custom_mapping_path:
-        resolvers.append(UserDefinedMapping(custom_mapping_path))
+    if custom_mapping_path or custom_mapping:
+        resolvers.append(
+            UserDefinedMapping(
+                mapping_path=custom_mapping_path, custom_mapping=custom_mapping
+            )
+        )
 
     resolvers.append(LocalPackageResolver(pyenv_path))
     if install_deps:
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -332,7 +332,7 @@ class IdentityMapping(BasePackageResolver):
 
 def resolve_dependencies(
     dep_names: Iterable[str],
-    custom_mapping_path: Optional[Path] = None,
+    custom_mapping_file: Optional[Path] = None,
     custom_mapping: Optional[Dict[str, List[str]]] = None,
     pyenv_path: Optional[Path] = None,
     install_deps: bool = False,
@@ -353,10 +353,10 @@ def resolve_dependencies(
     # each resolver in order until one of them returns a Package object. At
     # that point we are happy, and don't consult any of the later resolvers.
     resolvers: List[BasePackageResolver] = []
-    if custom_mapping_path or custom_mapping:
+    if custom_mapping_file or custom_mapping:
         resolvers.append(
             UserDefinedMapping(
-                mapping_path=custom_mapping_path, custom_mapping=custom_mapping
+                mapping_path=custom_mapping_file, custom_mapping=custom_mapping
             )
         )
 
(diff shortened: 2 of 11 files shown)


fawltydeps/main.py
```diff
@@ -82,7 +82,8 @@ def resolved_deps(self) -> Dict[str, Package]:
         """The resolved mapping of dependency names to provided import names."""
         return resolve_dependencies(
             (dep.name for dep in self.declared_deps),
-            custom_mapping_path=self.settings.custom_mapping,
+            custom_mapping_path=self.settings.custom_mapping_file,
```
Nour-Mws, line 85: Why not change to `custom_mapping_path` since this name is already used at different places?

fawltydeps/packages.py
```diff
@@ -131,10 +136,18 @@ def packages(self) -> Dict[str, Package]:
...
+            with open(self.mapping_path, "rb") as mapping_file:
+                toml_data = tomllib.load(mapping_file)
+                user_mapping = {**user_mapping, **toml_data}
 
-        logger.debug(f"Loading user-defined mapping from {self.mapping_path}")
-        with open(self.mapping_path, "rb") as mapping_file:
-            user_mapping = tomllib.load(mapping_file)
+        if self.custom_mapping is not None:
+            logger.debug("Loading user-defined mapping from custom mapping.")
+            # Custom mapping overrides mapping from file
```
Nour-Mws, line 149: You mean toml config custom mapping overrides custom mapping from file?

fawltydeps/packages.py
```diff
@@ -131,10 +136,18 @@ def packages(self) -> Dict[str, Package]:
...
+            user_mapping = {**user_mapping, **self.custom_mapping}
```
Nour-Mws, line 150: As per your comment above, and the way the dict is constructed here, I assume that if a package mapping is defined twice, the mapping from the toml section will override the mapping from file. But this is not visible to the user. When there's a mismatch between the two, I would expect:
- either an error to be thrown, informing the user that there's a mismatch between what they defined in the config section and what they defined in the file and that they need to fix that.
- or a union to be performed between the two, while also log.info-rming the user that we are performing a union.
What do you think?

fawltydeps/settings.py
```diff
@@ -122,7 +122,8 @@ class Settings(BaseSettings):  # type: ignore
     code: Set[PathOrSpecial] = {Path(".")}
     deps: Set[Path] = {Path(".")}
     pyenv: Optional[Path] = None
-    custom_mapping: Optional[Path] = None
+    custom_mapping_file: Optional[Path] = None
```
Nour-Mws, line 125: Same comment as above about using `custom_mapping_file` when `custom_mapping_path` is already in use elsewhere.

fawltydeps/settings.py
```diff
@@ -215,9 +216,6 @@ def create(cls, cmdline_args: argparse.Namespace) -> "Settings":
         if {"verbose", "quiet"}.intersection(args_dict.keys()):
             ret["verbosity"] = args_dict.get("verbose", 0) - args_dict.get("quiet", 0)
 
-        if "custom_mapping" in args_dict.keys():
```
Nour-Mws, line 218: The `cli_parser.py` file is not changed in this PR so I cannot comment on it, but shouldn't we change the argument name in that module from `custom-mapping` to `custom-mapping-path`?

fawltydeps/settings.py
```diff
@@ -215,9 +216,6 @@ def create(cls, cmdline_args: argparse.Namespace) -> "Settings":
...
-            ret["custom_mapping"] = args_dict.get("custom_mapping")
```
Nour-Mws, line 219: Assuming this should be `custom-mapping-path` not `custom-mapping`, shouldn't setting this option in the CLI override what has been configured via other means. Therefore, shouldn't these lines stay?

tests/test_packages.py
```diff
@@ -167,6 +167,11 @@ def test_user_defined_mapping__input_is_no_file__raises_unparsable_path_exeption
         UserDefinedMapping(SAMPLE_PROJECTS_DIR)
 
 
+def test_ser_defined_mapping__no_input__returns_empty_mapping():
```
Nour-Mws, line 170: is `ser` a typo?

tests/test_settings.py
```diff
@@ -239,6 +239,34 @@ class SettingsTestVector:
...
+        ),
+    ),
+    SettingsTestVector(
+        "config_file_with_mapping_and_cli__overrides_some_defaults",
+        config=dict(
+            actions=["list_deps"],
+            deps=["my_requirements.txt"],
+            custom_mapping={"package": ["foo", "bar"]},
+        ),
+        cmdline=dict(custom_mapping_file="mapping.toml"),  # should be list/set, not str
```
Nour-Mws, line 262: Is this comment meant for future work?

tests/test_settings.py
```diff
@@ -239,6 +239,34 @@ class SettingsTestVector:
             custom_mapping_file=Path("mapping.toml"),
         ),
     ),
+    SettingsTestVector(
```
Nour-Mws, line 242: Is it possible to add a test where a different `custom_mapping_path` is provided by both the cli and the config?

Nour-Mws (changes requested):
Great work Maria!
It took me some time to review since I had to wrap my head around how the settings object interacts with the cli.

I left mostly minor comments, but the two major issues are:
- Changing the name of the option from `custom-mapping` to `custom-mapping-path` in ArgParse.
- How to merge custom mappings from the config and a mapping file: override, error out or union?

fawltydeps/main.py
mknorps: I think the change is good for consistency, but rather to `custom_mapping_file`. We do not do any traversing here, and `_path` suffix indicates it may be file or directory.

tests/test_packages.py
mknorps: Yes , fixed :)

fawltydeps/packages.py
mknorps: So let's go with the union approach. I added also the info message for cases, where the same name is encountered in mapping from file and from configuration. I also added test for this interaction in `test_packages`.

fawltydeps/packages.py
mknorps: Changed according to the comment below.

fawltydeps/settings.py
mknorps: I kept `custom_mapping_file` - we do not traverse directories here, and `_path` suggests we have either directory or a file.

tests/test_settings.py
mknorps: Done -> `mixed_deps__uses_mixture_of_user_defined_identity_and_local_env_mapping` in test_packages

tests/test_settings.py
mknorps: Nope, this is irrelevant. Removed.

fawltydeps/settings.py
mknorps: It by default overrides configuration file. This is not needed, in fact was an artifact in the previous PR.

tests/test_settings.py
Nour-Mws: That test is great, thanks!
I was thinking more of adding a test in `test_settings.py` rather than `test_packages.py` to make sure the value in cli overrides the value in config.

fawltydeps/packages.py
Nour-Mws: That looks great!

tests/test_settings.py
Nour-Mws: That said, if it's the default behavior, I'm not sure how useful that test is. I haven't checked if we have tested that for other settings fields before.

fawltydeps/packages.py
```diff
@@ -131,19 +136,45 @@ def packages(self) -> Dict[str, Package]:
         This enumerates the available packages  _once_, and caches the result for
         the remainder of this object's life in _packages.
         """
```
Nour-Mws, line 138: I think the documentation of this function should be updated.

Nour-Mws (approved):
:tada: :tada: :tada:

tests/test_settings.py
mknorps: The test is still useful, as there may be some other ways to implement overriding defaults. There is a test that checks that `custiom_mapping_file` ovverrides default `None`: `config_file_with_mapping_and_cli__overrides_some_defaults`.
I added another one to check if the CLI value overrides config value: `config_file_with_mapping_and_cli__cli_mapping_overrides_config`

fawltydeps/packages.py
mknorps: Changed to:
```
        """Gather a custom mapping given by a user.
        
        Mapping may come from two sources:
        * _mapping_path_ which is a file in a given path, which is parsed to a ditionary
        * _custom_mapping_ which is a dictionary of package to imports mapping.
  ```****

File: fawltydeps/packages.py [hunk 11]
```diff
@@ -114,36 +114,71 @@ def lookup_packages(self, package_names: Set[str]) -> Dict[str, Package]:
...
+                    mapping[normalised_name].add_import_names(
+                        *imports, mapping=DependenciesMapping.USER_DEFINED
+                    )
+                else:
+                    mapping[normalised_name] = Package(
+                        name,
+                        {DependenciesMapping.USER_DEFINED: set(imports)},
+                    )
+
+        return mapping
```
jherland, line 180: I wonder if the code here becomes simpler if we merge the two dicts _before_ we instantiate the `Package` objects, something like:
```python
from collections import defaultdict

...

        mapping = defaultdict(set)
        for name, imports in custom_mapping_from_file.items():
            normalised_name = Package.normalize_name(name)
            mapping[normalised_name].update(imports)
        for name, imports in self.custom_mapping.items():
            normalised_name = Package.normalize_name(name)
            if normalised_name in mapping:
                logger.info(
                    "Mapping for %s already found in %s. Import names "
                    "from the configuration file are appended to ones "
                    "found in the mapping file.",
                    normalised_name,
                    self.mapping_path,
                )
            mapping[normalised_name].update(imports)

        return {
            name: Package(name, {DependenciesMapping.USER_DEFINED: set(imports)})
            for name, imports in mapping.items()
        }
```

(There are some details around normalization of package names that I might have glossed over here...)

fawltydeps/settings.py
```diff
@@ -238,12 +236,28 @@ def print_toml_config(settings: Settings, out: TextIO = sys.stdout) -> None:
...
+
     lines = [
         "# Copy this TOML section into your pyproject.toml to configure FawltyDeps",
         "# (default values are commented)",
         "[tool.fawltydeps]",
-    ] + [
-        f"{'# ' if has_default_value[name] else ''}{name} = {value!r}"
-        for name, value in simple_settings.items()
-    ]
+    ] + [_option_to_toml(name, value) for name, value in simple_settings.items()]
```
jherland, line 262: Nice! I hadn't thought at all about how to solve this, but this looks good!

tests/test_cmdline.py
```diff
@@ -917,7 +918,8 @@ def test_cmdline_on_ignored_undeclared_option(
                 # code = ['.']
                 deps = ['foobar']
                 # pyenv = None
-                # custom_mapping = None
+                # custom_mapping_file = None
+                # [tool.fawltydeps.custom_mapping]
```
jherland, line 922: Hmm. This should be at the bottom og the generated text, as otherwise, the options below will be considered part of this new section 😱

jherland (commented):
I've glossed over the tests so far, but this looks really good, and most of my comments are minor.

fawltydeps/packages.py
```diff
@@ -319,7 +354,8 @@ def lookup_packages(self, package_names: Set[str]) -> Dict[str, Package]:
 
 def resolve_dependencies(
     dep_names: Iterable[str],
-    custom_mapping_path: Optional[Path] = None,
+    custom_mapping_file: Optional[Path] = None,
+    custom_mapping: Optional[Dict[str, List[str]]] = None,
     pyenv_path: Optional[Path] = None,
     install_deps: bool = False,
 ) -> Dict[str, Package]:
```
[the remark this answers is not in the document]
mknorps: I would do it in a later PR. It may even require grouping of `Setting` object attributes. We have a lot of settings that are unrelated to resolving dependencies, such as verbosity, actions, etc.

fawltydeps/settings.py
```diff
@@ -122,7 +122,8 @@ class Settings(BaseSettings):  # type: ignore
     code: Set[PathOrSpecial] = {Path(".")}
     deps: Set[Path] = {Path(".")}
     pyenv: Optional[Path] = None
-    custom_mapping: Optional[Path] = None
+    custom_mapping_file: Optional[Path] = None
+    custom_mapping: Optional[Dict[str, List[str]]] = None
```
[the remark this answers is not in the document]
mknorps: > This means that we end up parsing the two mapping sources at different times

Yes, that is correct. But if we want to have both, mapping file and the mapping provided explicitly in the configuration file, I do not see how can we work around it. We have to parse config file anyway when creating a `Settings` object. What we could do is to abstract out the mapping parser itself, but the mapping is just a simple dictionary for now.

I think that giving a separate name for `CustomMapping` should be sufficient for now. I applied your suggestion.
We can observe how this evolves and if there will be more need to look into that dichotomy further.

fawltydeps/packages.py
```diff
@@ -114,36 +114,71 @@ def lookup_packages(self, package_names: Set[str]) -> Dict[str, Package]:
 class UserDefinedMapping(BasePackageResolver):
     """Use user-defined mapping loaded from a toml file"""
 
-    def __init__(self, mapping_path: Path) -> None:
+    def __init__(
+        self,
+        mapping_path: Optional[Path] = None,
+        custom_mapping: Optional[TomlData] = None,
```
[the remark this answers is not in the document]
mknorps: As in the above comment, changed to `CustomMapping` type.

fawltydeps/packages.py
mknorps: This is simpler indeed, but there will be a problem if mapping from settings has a different unnormalised name than mapping from the file (like `SciKit-Learn` vs `scikit-learn`).

The latter will update only with the latest seen mapping in your solution. In the current one, the mappings are merged (see Nour's comment: https://github.com/tweag/FawltyDeps/pull/292#discussion_r1157718469, although the `name` field in `Package` object will be the one first encountered.

jherland: AFAICS, after another look, all names are normalized before they are put into the `mapping` dict, so imports that are associated to the same _normalized_ name should all end up in the same `Package` object.

The difference is that the `.name` members of the resulting `Package` objects are _also_ normalized in my suggestion, whereas they are not in your diff. I'm not sure that makes any actual difference in practice.

Anyway, this is beyond nitpicking, and the PR is correct AFAICS in any case. Please consider this resolved.

tests/test_packages.py
```diff
@@ -264,41 +269,78 @@ def test_LocalPackageResolver_lookup_packages(dep_name, expect_import_names):
...
-    user_mapping_path = None
+    custom_mapping_file = None
+    custom_mapping = None
     if user_mapping is not None:
-        tmp_path = write_tmp_files({"mapping.toml": user_mapping})
-        user_mapping_path = tmp_path / "mapping.toml"
+        custom_mapping = user_mapping.get("configuration")
+        if "file" in user_mapping:
+            tmp_path = write_tmp_files({"mapping.toml": user_mapping["file"]})
+            custom_mapping_file = tmp_path / "mapping.toml"
```
jherland, line 336: In this specific use case of writing a single file I don't think `write_tmp_files()` adds much value. Consider using `tmp_path` directly instead:
```suggestion
            custom_mapping_file = tmp_path / "mapping.toml"
            custom_mapping_file.write_text(dedent(user_mapping["file"]))
```
Then, replace `write_tmp_files` with `tmp_path` in the argument list, and drop the `dedent()` calls in the parametrization above.

fawltydeps/settings.py
```diff
@@ -238,12 +236,43 @@ def print_toml_config(settings: Settings, out: TextIO = sys.stdout) -> None:
...
+        [
+            "# Copy this TOML section into your pyproject.toml to configure FawltyDeps",
+            "# (default values are commented)",
+            "[tool.fawltydeps]",
+        ]
+        + [
+            _option_to_toml(name, value)
+            for name, value in simple_settings.items()
+            # First handle non-dictionary options,
+            # as rest will be treated as members of a `name` dictionary
```
jherland, line 268: Last part of this comment does not make sense to me:
```suggestion
            # First handle non-dictionary options, as we don't want them to end
            # up inside the config section resulting from a dictionary option.
```

fawltydeps/settings.py
```diff
@@ -238,12 +236,43 @@ def print_toml_config(settings: Settings, out: TextIO = sys.stdout) -> None:
...
+            if name not in dictionary_options
+        ]
+        + [
+            _option_to_toml(name, value)
+            for name, value in simple_settings.items()
+            # Export dictionary options to TOML format
```
jherland, line 274:
```suggestion
            # Export dictionary options as separate TOML  config sections
```

jherland (approved):
Gone through the tests now as well. Really good stuff!

tests/test_configured_run.py
```diff
@@ -0,0 +1,74 @@
+"""Test configuration-based run of FawltyDeps tool.
+
+Use only pyproject.toml contents to determine setting of the run.
+"""
```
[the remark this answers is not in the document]
mknorps: You are right, this is more of an integration test. I marked it such. Most of our CLI tests also falls into that category, maybe we should also consider moving them to this category. On the other hand they are often indicating that the change may break something :woman_shrugging:

PR #296, opened 2023-04-06
Refactor resolver tests, transition to PBT

Nour-Mws:
This refactor was done to make it easier to test the resolver with the boolean parameters `install_deps` and `identity_mapping` (in future PRs).
Without this refactor, the list of parameterized cases for the resolver test will grow by a factor of 2 for each additional parameter. This would require manual maintenance and is increasingly hard to parse.


tests/test_resolver.py
```diff
@@ -0,0 +1,129 @@
...
+
+user_mappings = st.one_of(st.none(), st.just(user_defined_mapping))
+
+
+def generate_expected_resolved_deps(
+    locally_installed_deps: Optional[Dict[str, List[str]]] = None,
+    non_locally_installed_deps: Optional[List[str]] = None,
+    user_defined_deps: Optional[Dict[str, List[str]]] = None,
+):
+    ret = {}
```
mknorps, line 57: From current implementation I get following resolution importance order:
- Identity Mapping
- User-defined mapping
- local env



In `resolve_dependencies` function we have:
- User-defined mapping
- local env
- temporary pip-install resolver
- Identity Mapping

What do you think about it?
It would be also helpful here to add a docstring that says what is the resolution order. I know it repeats what is in `resolved_deps`, but `generate_expected_resolved_deps` should just emulate the real function's behaviour.

tests/test_resolver.py
```diff
@@ -0,0 +1,129 @@
...
+                dep: Package(dep, {DependenciesMapping.IDENTITY: {dep}})
+                for dep in non_locally_installed_deps
+            }
+        )
+    return ret
+
+
+# Suppressing the warning on function scoped fixtures: even if the tmp writing
+# fixture only runs once for all the test cases, it will be writing the same
+# content
```
mknorps, line 90: :+1:
I would be more explicit relating to docummentation:
```suggestion
# Suppressing the warning on function scoped fixtures: even if the tmp writing
# fixture only runs once for all the test cases, it is not a problem,
# as the fixture-generated content does not have to be reset between test examples.
# The test function only reads the file content and filters needed input.
```

tests/test_resolver.py
```diff
@@ -0,0 +1,129 @@
...
+    user_deps,
+    installed_deps,
+    non_installed_deps,
+    user_mapping,
+    write_tmp_files,
+):
+
+    # The case where there are user-defined deps but no user-defined mapping
+    # provided is not valid
+    assume(not (len(user_deps) > 0 and user_mapping is None))
```
mknorps, line 108: WIP - I will have a comment here

mknorps (changes requested):
I love that you moved those tests to a separate module and parametrized them with hypothesis!

tests/test_resolver.py
Nour-Mws: Updated in [2e3b6db](https://github.com/tweag/FawltyDeps/pull/296/commits/2e3b6db8f89dd4294fb3387c06e48803a463a31d).

tests/test_resolver.py
Nour-Mws: So the `generate_expected_resolved_deps` function does not actually resolve dependencies. It just replaces the generation of the "expected" dict that we were manually specifying in the parameterized tests.
There's no order for how the different categories are resolved. The resulting dictionary is just constructed from the different categories of deps.

To make this clearer, I added a docstring to `generate_expected_resolved_deps` in [e70a35a](https://github.com/tweag/FawltyDeps/pull/296/commits/e70a35ac22f13cdadbb92a46f9161f6106c34330) and added asserts to the test to ensure the condition of no overlap between different categories is respected in [e7443ab](https://github.com/tweag/FawltyDeps/pull/296/commits/e7443ab7a4ec6d04370d903794a92ffecd09db39).

tests/test_resolver.py
Nour-Mws: Now that this is rebased on the changes that include having both a file-defined mapping and a toml-defined mapping, there are three different variables to distinguish:
- A file-defined user mapping + a strategy that either includes that mapping in the test input or passes `None`
- A toml user mapping + a strategy to either include it or not in the test input
- A list of user deps. This list is part of the deps that would be passed to the resolver. It has no overlap with the other categories of deps (locally-installed and other deps). And it includes the deps that should be resolved from the user-provided mapping.

It is not valid to have some deps that need to be resolved from the user-defined mapping while no user-defined mapping is provided. This could happen when the strategies for both the file-defined and the toml-defined user mappings produce `None`, but `user_deps` is not an empty list (it contains e.g. `apache-airflow`, which would not be resolved as a user-defined dependency in this case).

tests/test_resolver.py
mknorps: Thank you, now and after the pair programming discussion I know we were thinking about validation checks in different places. You made sure that the dependencies sets were disjoint in the main test function, and I wanted to make sure that we have a good resolution order not knowing that `generate_expected_resolver_deps` already tasks a disjoint sets.

With the current version, all is good :+1:

mknorps (approved):
Great work on this @Nour-Mws !
I think this PR is ready to be merged :rocket: 

If you anyway have some spare 10 minutes, please take a look at some simplification I propose in the branch `maria/nour-simplified-refactor`. I described them commit by commit.

tests/test_resolver.py
```diff
@@ -0,0 +1,169 @@
...
+    def sample_dict_keys_and_values_strategy(input_dict: Dict[str, List[str]]):
+        """Returns a hypothesis strategy to choose keys and values from a dict."""
+        keys = draw(st.lists(st.sampled_from(list(input_dict.keys())), unique=True))
+        return {
+            key: draw(st.lists(st.sampled_from(input_dict[key]), unique=True))
+            for key in keys
+        }
+
+    user_mapping_in_file = sample_dict_keys_and_values_strategy(user_mapping)
+    user_mapping_in_config = sample_dict_keys_and_values_strategy(user_mapping)
```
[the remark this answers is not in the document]
Nour-Mws: That makes sense. Done in [4c892b7](https://github.com/tweag/FawltyDeps/pull/296/commits/4c892b7db64c1746644b6760ae4b6d09b00d72d7).

PR #297, opened 2023-04-06
Test the resolver with the `install_deps` option

Nour-Mws:
:warning: Do not merge until the following has landed:
- [x] #296 

I opted to make this into a separate PR from the refactor PR because the battery of hypothesis tests slows down significantly with this change (it now takes ~35 seconds to run tests in `test_resolver.py`).

The second commit is my attempt to speed things up by installing packages from wheels instead of tarballs. Although the package is installed from the wheel (see log below), the gain in speed is not significant.
```sh
Looking in links: /home/nour/.cache/pytest/d/fawltydeps
Processing /home/nour/.cache/pytest/d/fawltydeps/leftpadx-0.0.1-py3-none-any.whl
Installing collected packages: leftpadx
Successfully installed leftpadx-0.0.1
```

I've also tried to skip the `get_tarballs()` step (which logically should not involve any time-consuming logic if the tarballs are already downloaded). Unfortunately, I did not see any speed improvement.

### DIFF
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -8,10 +8,9 @@ import venv
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from dataclasses import dataclass, replace
+from functools import partial
 from pathlib import Path
 from typing import Dict, Iterable, Iterator, List, Optional, Set, Tuple, Type, Union
-
-import pytest
 
 # importlib_metadata is gradually graduating into the importlib.metadata stdlib
 # module, however we rely on internal functions and recent (and upcoming)
@@ -310,15 +309,28 @@ class TemporaryPipInstallResolver(BasePackageResolver):
     packages in this venv. The venv is automatically deleted before as soon as
     the packages have been resolved."""
 
-    def __init__(
-        self,
-        cache: Optional[pytest.Cache] = None,
-    ) -> None:
-        self.cache = cache
+    # This is only used in tests by `test_resolver`
+    cached_venv: Optional[Path] = None
 
     @staticmethod
     @contextmanager
-    def temp_installed_requirements(requirements: List[str]) -> Iterator[Path]:
+    def installed_requirements(
+        venv_dir: Path, requirements: List[str]
+    ) -> Iterator[Path]:
+        """Install the given requirements into venv_dir with `pip install`."""
+        marker_file = venv_dir / ".installed"
+        if not marker_file.is_file():
+            venv.create(venv_dir, clear=True, with_pip=True)
+        subprocess.run(
+            [f"{venv_dir}/bin/pip", "install", "--no-deps"] + requirements,
+            check=True,  # fail if any of the commands fail
+        )
+        marker_file.touch()
+        yield venv_dir
+
+    @classmethod
+    @contextmanager
+    def temp_installed_requirements(cls, requirements: List[str]) -> Iterator[Path]:
         """Create a temporary venv and install the given requirements into it.
 
         Provide a path to the temporary venv into the caller's context in which
@@ -326,44 +338,8 @@ class TemporaryPipInstallResolver(BasePackageResolver):
         the venv at the end of the context.
         """
         with tempfile.TemporaryDirectory() as tmpdir:
-            venv_dir = Path(tmpdir)
-            venv.create(venv_dir, clear=True, with_pip=True)
-            subprocess.run(
-                [f"{venv_dir}/bin/pip", "install", "--no-deps"] + requirements,
-                check=True,  # fail if any of the commands fail
-            )
-            (venv_dir / ".installed").touch()
-            yield venv_dir
-
-    # This function is only tagged as a context manager to facilitate an
-    # intechangeable use with `temp_installed_requirements()`
-    @staticmethod
-    @contextmanager
-    def cached_venv_with_requirements(
-        requirements: List[str], cache: pytest.Cache
-    ) -> Iterator[Path]:
-        """Install the given requirements into a cached venv.
-
-        Provide a path to the cached venv.
-        The venv persists after the call (and the installed deps are not
-        uninstalled).
-        The `--upgrade` and `--force-reinstall` options are not used to avoid
-        parallel tests over-writing the same package.
-        """
-
-        venv_dir = Path(
-            cache.mkdir(
-                f"fawltydeps_reused_venv_{sys.version_info.major}.{sys.version_info.minor}"
-            )
-        )
-        if not (venv_dir / ".installed").is_file():
-            venv.create(venv_dir, clear=True, with_pip=True)
-        subprocess.run(
-            [f"{venv_dir}/bin/pip", "install", "--no-deps"] + requirements,
-            check=True,  # fail if any of the commands fail
-        )
-        (venv_dir / ".installed").touch()
-        yield venv_dir
+            with cls.installed_requirements(Path(tmpdir), requirements) as venv_dir:
+                yield venv_dir
 
     def lookup_packages(
         self,
@@ -376,16 +352,15 @@ class TemporaryPipInstallResolver(BasePackageResolver):
         on this venv to provide the Package objects that correspond to the
         package names.
         """
-        if self.cache:
-            venv_context = self.cached_venv_with_requirements(
-                sorted(package_names), self.cache
-            )
+        if self.cached_venv is None:
+            installed = self.temp_installed_requirements
+            logger.info("Installing dependencies into a temporary Python environment.")
+        # If self.cached_venv has been set, then use that path instead of creating
+        # a temporary venv for package installation.
         else:
-            venv_context = self.temp_installed_requirements(sorted(package_names))
-            logger.info(
-                "Installing dependencies into a new temporary Python environment."
-            )
-        with venv_context as venv_dir:
+            installed = partial(self.installed_requirements, self.cached_venv)
+            logger.info(f"Installing dependencies into {self.cached_venv}.")
+        with installed(sorted(package_names)) as venv_dir:
             resolver = LocalPackageResolver(venv_dir)
             return {
                 name: replace(
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -11,6 +11,8 @@ from dataclasses import dataclass, replace
 from pathlib import Path
 from typing import Dict, Iterable, Iterator, List, Optional, Set, Tuple, Type, Union
 
+import pytest
+
 # importlib_metadata is gradually graduating into the importlib.metadata stdlib
 # module, however we rely on internal functions and recent (and upcoming)
 # bugfixes that will first be available in the stdlib version in Python v3.12
@@ -308,6 +310,12 @@ class TemporaryPipInstallResolver(BasePackageResolver):
     packages in this venv. The venv is automatically deleted before as soon as
     the packages have been resolved."""
 
+    def __init__(
+        self,
+        cache: Optional[pytest.Cache] = None,
+    ) -> None:
+        self.cache = cache
+
     @staticmethod
     @contextmanager
     def temp_installed_requirements(requirements: List[str]) -> Iterator[Path]:
@@ -327,7 +335,40 @@ class TemporaryPipInstallResolver(BasePackageResolver):
             (venv_dir / ".installed").touch()
             yield venv_dir
 
-    def lookup_packages(self, package_names: Set[str]) -> Dict[str, Package]:
+    # This function is only tagged as a context manager to facilitate an
+    # intechangeable use with `temp_installed_requirements()`
+    @staticmethod
+    @contextmanager
+    def cached_venv_with_requirements(
+        requirements: List[str], cache: pytest.Cache
+    ) -> Iterator[Path]:
+        """Install the given requirements into a cached venv.
+
+        Provide a path to the cached venv.
+        The venv persists after the call (and the installed deps are not
+        uninstalled).
+        The `--upgrade` and `--force-reinstall` options are not used to avoid
+        parallel tests over-writing the same package.
+        """
+
+        venv_dir = Path(
+            cache.mkdir(
+                f"fawltydeps_reused_venv_{sys.version_info.major}.{sys.version_info.minor}"
+            )
+        )
+        if not (venv_dir / ".installed").is_file():
+            venv.create(venv_dir, clear=True, with_pip=True)
+        subprocess.run(
+            [f"{venv_dir}/bin/pip", "install", "--no-deps"] + requirements,
+            check=True,  # fail if any of the commands fail
+        )
+        (venv_dir / ".installed").touch()
+        yield venv_dir
+
+    def lookup_packages(
+        self,
+        package_names: Set[str],
+    ) -> Dict[str, Package]:
         """Convert package names into Package objects via temporary pip install.
 
         Use the temp_installed_requirements() above to `pip install` the given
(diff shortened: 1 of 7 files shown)


Nour-Mws:
@jherland, Having merged #296, I rebased this PR on main and it's now ready for review.

tests/test_resolver.py
```diff
@@ -103,12 +106,23 @@ def generate_expected_resolved_deps(
         resolved_packages = user_mapping.lookup_packages(set(user_defined_deps))
         ret.update(resolved_packages)
```
jherland, line 107: I realize this is really a comment on #296, but I'm a little suspicious of _using_ `UserDefinedMapping` to test the correctness of `UserDefinedMapping`. It's easy for a bug in that code to manifest in the same way both in the actual data produced by `resolve_dependencies()` _and_ in the expected data produced here, and thus not trigger a failed test.

In other words (and I noticed this while rebasing #313 across #296), there is currently no mention of `DependenciesMapping.USER_DEFINED` in any of our test code... 🤔

Nour-Mws: True, and I pondered on this before using `UserDefinedMapping` in #296. But `test_resolver.py` is not supposed to test the correctness of `UserDefinedMapping`. It's only meant to test the logic in `resolve_dependencies`. If a bug is to manifest in `UserDefinedMapping`, it should be the job of that class's test to catch it.

I can alternatively switch to `DependenciesMapping.USER_DEFINED` in `test_resolver.py` but then I'll also have to copy a large part of the merge currently happening in `UserDefinedMapping` into `generate_expected_resolved_deps` to construct the user mapping from the mapping file and the custom dictionary mapping.

What do you suggest?

tests/test_resolver.py
```diff
@@ -119,14 +133,22 @@ def generate_expected_resolved_deps(
...
     user_mapping=user_mapping_strategy(user_defined_mapping),
     installed_deps=dict_subset_strategy(locally_installed_deps),
-    other_deps=st.lists(st.sampled_from(other_deps), unique=True),
+    other_deps=dict_subset_strategy(other_deps),
+    install_deps=st.booleans(),
+)
+@settings(
+    suppress_health_check=[HealthCheck.function_scoped_fixture],
+    deadline=5000,
+    max_examples=50,
```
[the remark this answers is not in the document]
Nour-Mws: These are needed because of the following:
>I opted to make this into a separate PR from the refactor PR because the battery of hypothesis tests slows down significantly with this change (it now takes ~35 seconds to run tests in test_resolver.py).

So I had to both limit the number of generated examples and increase the time limit so that hypothesis testing does not fail on tests taking too long.

tests/python_packages.toml
```diff
@@ -17,3 +17,7 @@ sha256 = "b305b30c99526df5e63d6022dd2310a0a941a187bd3884f4c8ef0418df6c39f3"
 [click]
 url = "https://files.pythonhosted.org/packages/59/87/84326af34517fca8c58418d148f2403df25303e02736832403587318e9e8/click-8.1.3.tar.gz"
 sha256 = "7682dc8afb30297001674575ea00d1814d808d6a36af415a82bd481d37ba7b8e"
+
+[leftpadx-wheel]
+url = "https://files.pythonhosted.org/packages/37/ca/41e9d0964799b00ad00c6254e6c0bc1a26c604456ab659e13cf91e5bdd81/leftpadx-0.0.1-py3-none-any.whl"
+sha256 = "512084949099fab88042be250a6da55dcf933ac3af8feffddefd52dace66160c"
```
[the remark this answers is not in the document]
jherland: Ah, ok. In general I suspect that any `*-py3-none-any.whl` should be fine to cache instead of the corresponding `*.tar.gz`, as we should never run into platform/version incompatibilities for those.

I agree with your disappointment that installation is not much improved by using wheels instead of tarball. Maybe the overhead of creating the venv itself is too large in comparison? Looking at `test_local_env`, we have two tests that create a new venv with `pip` inside:
- `test_local_env__default_venv__contains_pip_and_setuptools` (creates the venv in the test itself)
- `test_on_installed_venv__returns_local_deps` (creates the venv inside `TemporaryPipInstallResolver`)
and they each take 2-3s to run on my machine. All the other `test_local_env` tests are <0.05s.

I suspect this ends up supporting the argument in favor of caching the results of `install_deps`, so that these venvs can be used as little as possible. Won't help much with test coverage of this code path, though... 😿

tests/test_resolver.py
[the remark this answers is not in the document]
jherland: I see. I suspect the underlying reason is that anytime we pass `install_deps=True`, we're going to spend ~2s extra simply to create the venv, and this this is fundamentally unfriendly to hypothesis which obviuosly wants to call `resolve_dependencies` _a lot_...

As you say above, `test_resolver` is not _really_ about testing each individual resolver, but rather about the logic of combining resolvers in `resolve_dependencies()`. Maybe we can extend the "test intrumentation"  of `TemporaryPipInstallResolver`: Not only do we want a local PyPI mirror to avoid contacting pypi.org in each test, we also want to somehow reuse/cache/fake the virtualenv in which `TemporaryPipInstallResolver` calls `pip install`, so that we can speed up these tests?

tests/test_resolver.py
jherland: Ah, I see. I agree with your analysis that `test_resolver` is supposed to test the _combination_ of resolvers, and not each individual one.

We do have `test_user_defined_mapping__well_formated_input_file__parses_correctly` in `test_packages` that does indeed test `UserDefinedMapping` separately (although it only tests `.import_names` and not `.mappings`, which is why I couldn't find `DependenciesMapping.USER_DEFINED` anywhere).

I suggest we leave this PR as-is on this point, and then I'll add some relevant asserts to `test_user_defined_mapping__well_formated_input_file__parses_correctly` in my #313.

jherland: Added this in commit 5e5904a on #313.

tests/test_resolver.py
```diff
@@ -72,10 +74,11 @@ def user_mapping_to_file_content(user_mapping: Dict[str, List[str]]) -> str:
 
 def generate_expected_resolved_deps(
     locally_installed_deps: Optional[Dict[str, List[str]]] = None,
-    other_deps: Optional[List[str]] = None,
+    other_deps: Optional[Dict[str, List[str]]] = None,
     user_defined_deps: Optional[List[str]] = None,
     user_mapping_file: Optional[Path] = None,
     user_mapping_from_config: Optional[Dict[str, List[str]]] = None,
```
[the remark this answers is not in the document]
jherland: > > What is the difference between None and empty dict for these arguments? Could we drop the Optional?
> 
> Agree it's worth dropping the `Optional`. Any suggestion on how not to use a default `{}` argument (this is not a dataclass so I cannot use `field(default_factory=dict)`)?

AFAICS this function is called _once_, and all arguments are passed in that case, so there should not be any need for default arguments at all?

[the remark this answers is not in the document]
Nour-Mws: That's true. I removed the `optional` typing in [2d1086f](https://github.com/tweag/FawltyDeps/pull/297/commits/2d1086f0f947d2e14d68cb53906e7c5483b3e9ba).

tests/test_resolver.py
```diff
@@ -13,8 +13,10 @@
     resolve_dependencies,
 )
 
+from .project_helpers import TarballPackage
+
 # The deps in each category should be disjoint
-other_deps = ["pandas", "numpy", "other"]
+other_deps = {"leftpadx": ["leftpad"]}
```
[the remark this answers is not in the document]
Nour-Mws: >This replaces three deps with one, which removes some test coverage of the case where there is more than one "other" dep. I understand that these have to be available in our package cache in order for tests to work with install_deps == True, but could we add another package here like click or flit-core?

I didn't fully read your comment before implementing the solution so I went with another package than those you suggested. See [66b3d4f](https://github.com/tweag/FawltyDeps/pull/297/commits/66b3d4fbb8d67240a202b4b08996727baba0e234).

>Furthermore, the module added here does not "work" with the identity mapping (leftpadx vs leftpad). Is this change intenttional? Or does it not really matter to the tests?

The original `other_deps` object was a list when we were solely using identity mapping as the actual imports exposed by the package did not matter. Now that we're optionally using `install_deps`, `other_deps` need to define the imports exposed by each package so that those can be checked in the case `install_deps` is used. But those will be ignored in the case identity mapping is used.

tests/test_resolver.py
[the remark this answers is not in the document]
Nour-Mws: > Maybe we can extend the "test intrumentation" of TemporaryPipInstallResolver: Not only do we want a local PyPI mirror to avoid contacting pypi.org in each test, we also want to somehow reuse/cache/fake the virtualenv in which TemporaryPipInstallResolver calls pip install, so that we can speed up these tests?

That would be indeed great. However, I don't have any immediate solution popping in my mind right now. Let me sleep on it until tomorrow.

tests/python_packages.toml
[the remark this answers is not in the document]
jherland: Ah, sorry, I was only referring to the cache mentioned here:

fawlty-deps-pypi-realm

With something like that in place, the user should most of the time not have to wait for FD to create venvs and install deps.

But that does not help us now, precisely because we _want_ to test that creating the venv and installing the deps actually works as it should.

Or at least: we want to test this in the `TemporaryPipInstallResolver` unit tests. The `test_resolver` PBT tests should maybe be allowed to instrument/mock this part of `TemporaryPipInstallResolver` to speed things up? Maybe using a single venv across the entire `test_resolver` session (so as to avoid the setup cost for venvs) is sufficient to make the tests fast enough?

Nour-Mws:
@jherland [dc402f6](https://github.com/tweag/FawltyDeps/pull/297/commits/dc402f6bdab25421f4d07b641340fa72aa7c617e) implements the solution we agreed upon this morning.

fawltydeps/packages.py
```diff
@@ -13,6 +13,8 @@
 from pathlib import Path
 from typing import Dict, Iterable, Iterator, List, Optional, Set
 
+import pytest
```
jherland, line 16: This is the core of what I want to avoid, since I don't want to have `pytest` become a _main_ dependency of FawltyDeps.

fawltydeps/packages.py
```diff
@@ -309,6 +311,12 @@ class TemporaryPipInstallResolver(BasePackageResolver):
     packages in this venv. The venv is automatically deleted before as soon as
     the packages have been resolved."""
 
+    def __init__(
+        self,
+        cache: Optional[pytest.Cache] = None,
+    ) -> None:
+        self.cache = cache
+
```
jherland, line 319: Instead of this we can use a `cached_venv` class member that defaults to `None`, and that `test_resolver` sets to the cached venv dir before running the test. See below.

fawltydeps/packages.py
```diff
@@ -328,16 +336,57 @@ def temp_installed_requirements(requirements: List[str]) -> Iterator[Path]:
...
+            )
+        )
+        if not (venv_dir / ".installed").is_file():
+            venv.create(venv_dir, clear=True, with_pip=True)
+        subprocess.run(
+            [f"{venv_dir}/bin/pip", "install", "--no-deps"] + requirements,
+            check=True,  # fail if any of the commands fail
+        )
+        (venv_dir / ".installed").touch()
+        yield venv_dir
```
jherland, line 367: This method is _very_ similar to the `temp_installed_requirements()` above, and I think some more refactoring can remove the duplication (while at the same time removing the pytest-specific stuff):

```python
    @staticmethod
    @contextmanager
    def installed_requirements(
        venv_dir: Path, requirements: List[str]
    ) -> Iterator[Path]:
        """Install the given requirements into venv_dir with `pip install`."""
        marker_file = venv_dir / ".installed"
        if not marker_file.is_file():
            venv.create(venv_dir, clear=True, with_pip=True)
        subprocess.run(
            [f"{venv_dir}/bin/pip", "install", "--no-deps"] + requirements,
        )
        marker_file.touch()
        yield venv_dir

    @classmethod
    @contextmanager
    def temp_installed_requirements(cls, requirements: List[str]) -> Iterator[Path]:
        """Create a temporary venv and install the given requirements into it.

        Provide a path to the temporary venv into the caller's context in which
        the given requirements have been `pip install`ed. Automatically remove
        the venv at the end of the context.
        """
        with tempfile.TemporaryDirectory() as tmpdir:
            with cls.installed_requirements(Path(tmpdir), requirements) as venv_dir:
                yield venv_dir
```
The first context manager takes care of creating a venv (only if needed) in a given `venv_dir` and installing the given requirements, and the second context manager then only needs to create a temporary directory, and call the first context manager to do the rest.

There is one difference in the argument list between the two context managers, but we can fix that with `functools.partial` below...

fawltydeps/packages.py
```diff
@@ -328,16 +336,57 @@ def temp_installed_requirements(requirements: List[str]) -> Iterator[Path]:
...
+        if self.cache:
+            venv_context = self.cached_venv_with_requirements(
+                sorted(package_names), self.cache
+            )
+        else:
+            venv_context = self.temp_installed_requirements(sorted(package_names))
+            logger.info(
+                "Installing dependencies into a new temporary Python environment."
+            )
+        with venv_context as venv_dir:
```
jherland, line 389: This code can stay more or less the same IMHO, but with the use of the `.cached_venv` class member and the two context managers above it looks like this in my version:

```python
from functools import partial

...

        if self.cached_venv is None:
            installed = self.temp_installed_requirements
            logger.info("Installing dependencies into a temporary Python environment.")
        else:
            installed = partial(self.installed_requirements, self.cached_venv)
            logger.info(f"Installing dependencies into {self.cached_venv}.")
        with installed(sorted(package_names)) as venv_dir:
```

pyproject.toml
```diff
@@ -53,6 +53,8 @@ optional = true
 [tool.poetry.group.test.dependencies]
 hypothesis = "^6.68.2"
 pytest = "^7.1.0"
+pytest-mock = "^3.10.0"
```
jherland, line 56: I think we can avoid this added dependency with the `.cached_venv` class member I suggest above (and set in `test_resolver` below).

tests/test_resolver.py
```diff
@@ -187,6 +184,14 @@ def test_resolve_dependencies__generates_expected_mappings(
...
     )
 
+    # patch TemporaryPipInstallResolver so that it's called with the cache
+    # option in the resolver
+    tmp_pip_install_resolver = TemporaryPipInstallResolver(cache=request.config.cache)
+    mocker.patch(
+        "fawltydeps.packages.TemporaryPipInstallResolver",
+        return_value=tmp_pip_install_resolver,
+    )
+
```
jherland: Here is where I'd like to place the code that "knows" about the pytest cache. We turn that into a simple `Path` that we can store into `TemporaryPipInstallResolver.cached_venv`.

Additionally, since I'm not 100% sure how changing the `TemporaryPipInstallResolver` class will work across various test cases, I want to _reset_ the class member back to `None` after the test is run, with a `try ... finally` block. After all, we _don't_ want a `TemporaryPipInstallResolver.cached_venv` from this test to affect the `TemporaryPipInstallResolver` unit tests!

```python
    # Tell TemporaryPipInstallResolver to reuse our cached venv, instead of
    # potentially creating a new venv for every test case.
    cached_venv = Path(
        request.config.cache.mkdir(
            f"fawltydeps_reused_venv_{sys.version_info.major}.{sys.version_info.minor}"
        )
    )
    try:
        TemporaryPipInstallResolver.cached_venv = cached_venv

        obtained = resolve_dependencies(
            ...
        )
    finally:
        TemporaryPipInstallResolver.cached_venv = None

    assert obtained == expected
```

jherland (changes requested):
I love the idea of using the pytest cache to cache this venv across test runs, but I really don't like having to `import pytest` in our shipped FawltyDeps code in order to make it work. I think it should be possible to move all the pytest-specific stuff into the `test_resolver`, and only leave a path to the prepared venv here.

See below for individual suggestions, or look at https://github.com/jherland/FawltyDeps/commit/841ab916f7bfa00b5ab99511e7fe8117a11670ad for a commit on top of this that contains all my suggestions.

Nour-Mws:
@jherland this is ready for a final look! Thanks a lot for the helpful suggestions and detailed comments.

jherland (approved):
🚀 🚀 🚀

PR #300, opened 2023-04-10
Fix boolean representation and skip null values in generated toml config

### LINKED ISSUE #265: Fix generating toml config to skip null values and correctly encode bool values
From @jherland in https://github.com/tweag/FawltyDeps/pull/252#discussion_r1145530626:
> I am reminded here that I took a shortcut when I wrote the code that generates the TOML section from our settings: TOML's booleans (true/false) are slightly different than Python's (True/False), and we don't take this into account when generating the TOML config section.
> 
> (The same goes for None which does not really exist at all in TOML - the way to pass None from TOML is to not set the corresponding config directive).
> 
> Anyway this should be solved in a separate issue/PR. Probably not very high-priority, as the generated TOML is only advisory, and must pass through human hands in any case before it's interpreted.

Nour-Mws:
Fixes #265:
- TOML's booleans (true/false) are slightly different to Python's (True/False).
- The way to pass None from TOML is to not set the corresponding config directive.

This was needed to let new tests pass in #298.

### DIFF
--- a/fawltydeps/settings.py
+++ b/fawltydeps/settings.py
@@ -232,15 +232,32 @@ def print_toml_config(settings: Settings, out: TextIO = sys.stdout) -> None:
         Options that are of dictionary type must be given a section entry.
         Assumption: dictionaries options are not nested.
         """
-        comment_char = "# " if has_default_value[name] else ""
+        toml_value_map = {
+            "None": "...",  # always commented, see sanity check below
+            "False": "false",
+            "True": "true",
+            # add more values here for which repr(value) is not valid TOML
+        }
+
+        if value is None:
+            # sanity check: None values are represented in TOML by omission,
+            # hence make sure these are always commented (i.e. equal to default)
+            assert has_default_value[name]
+
+        prefix = "# " if has_default_value[name] else ""
+
         if name in dictionary_options:
-            toml_option = f"{comment_char}[tool.fawltydeps.{name}]"
+            toml_option = f"{prefix}[tool.fawltydeps.{name}]"
             if value is not None:
                 toml_option += "".join(
-                    [f"\n{comment_char}{k} = {v!r}" for k, v in value.items()]
+                    [
+                        f"\n{prefix}{k} = {toml_value_map.get(str(v), repr(v))}"
+                        for k, v in value.items()
+                    ]
                 )
         else:
-            toml_option = f"{comment_char}{name} = {value!r}"
+            toml_value = toml_value_map.get(str(value), repr(value))
+            toml_option = f"{prefix}{name} = {toml_value}"
         return toml_option
 
     lines = (
--- a/fawltydeps/settings.py
+++ b/fawltydeps/settings.py
@@ -251,12 +251,12 @@ def print_toml_config(settings: Settings, out: TextIO = sys.stdout) -> None:
             if value is not None:
                 toml_option += "".join(
                     [
-                        f"\n{prefix}{k} = {toml_value_map.get(str(v), repr(v))}"
+                        f"\n{prefix}{k} = {toml_value_map.get(repr(v), repr(v))}"
                         for k, v in value.items()
                     ]
                 )
         else:
-            toml_value = toml_value_map.get(str(value), repr(value))
+            toml_value = toml_value_map.get(repr(value), repr(value))
             toml_option = f"{prefix}{name} = {toml_value}"
         return toml_option
 
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -859,6 +859,30 @@ def test_cmdline_on_ignored_undeclared_option(
             ).splitlines(),
             id="generate_toml_config_with_combo_of_config_and_cmdline_options",
         ),
+        pytest.param(
+            {"actions": ["check_undeclared"]},
+            ["--pyenv=None", "--generate-toml-config"],
+            dedent(
+                """\
+                # Copy this TOML section into your pyproject.toml to configure FawltyDeps
+                # (default values are commented)
+                [tool.fawltydeps]
+                actions = ['check_undeclared']
+                # output_format = 'human_summary'
+                # code = ['.']
+                # deps = ['.']
+                pyenv = 'None'
+                # custom_mapping_file = ...
+                # ignore_undeclared = []
+                # ignore_unused = []
+                # deps_parser_choice = ...
+                # install_deps = false
+                # verbosity = 0
+                # [tool.fawltydeps.custom_mapping]
+                """
+            ).splitlines(),
+            id="generate_toml_config_with_a_setting_set_to_str_None",
+        ),
     ],
 )
 def test_cmdline_args_in_combination_with_config_file(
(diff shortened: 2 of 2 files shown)


fawltydeps/settings.py
```diff
@@ -242,8 +242,13 @@ def print_toml_config(settings: Settings, out: TextIO = sys.stdout) -> None:
...
-        for name, value in simple_settings.items()
     ]
+    for name, value in simple_settings.items():
+        if value is None:
+            continue
+        line = f"{'# ' if has_default_value[name] else ''}"
+        if isinstance(value, bool):
+            lines.append(line + f"{name} = {str(value).lower()}")
+        else:
+            lines.append(line + f"{name} = {value!r}")
```
jherland, line 253: After a second look, I wonder if this is closer to what we want:

```suggestion
    toml_value_map = {
        None: "...",  # always commented, see sanity check below
        False: "false",
        True: "true",
        # add more values here for which repr(value) is not valid TOML
    }
    for name, value in simple_settings.items():
        prefix = "# " if has_default_value[name] else ""
        if value is None:
            # sanity check: None values are represented in TOML by omission, 
            # hence make sure these are always commented (i.e. equal to default)
            assert has_default_value[name]
        toml_value = toml_value_map.get(value, repr(value))
        lines.append(f"{prefix}{name} = {toml_value}")
```

tests/test_cmdline.py
```diff
@@ -916,12 +916,9 @@ def test_cmdline_on_ignored_undeclared_option(
                 output_format = 'human_detailed'
                 # code = ['.']
                 deps = ['foobar']
-                # pyenv = None
-                # custom_mapping = None
                 # ignore_undeclared = []
                 # ignore_unused = []
-                # deps_parser_choice = None
-                # install_deps = False
```
jherland, line 924: As stated above, we should consider _not_ removing the `None`-valued  directives here, as that makes them less visible to the user as potential customization points.

Either that, or we should consider a bigger change where we omit default-valued directives altogether. (The argument being that the user should rather look to our documentation for potential customization points. Not sure I agree with that, though...)

jherland (changes requested):
See the individual comments for more context, but the more I think about it, the less I think it makes sense to _skip_ null values without also skipping other default values. At that point I'd rather _keep_ the null values, but at the same time make sure that they are (a) commented, and (b) that it is clear to the user that they must be customized before uncommenting.

tests/test_cmdline.py
Nour-Mws: Tests fixed to replace `None` with the `...` placeholder in revised [cd44db9](https://github.com/tweag/FawltyDeps/pull/300/commits/cd44db9ed259b358338a61341975c8990a2df9f6).

fawltydeps/settings.py
Nour-Mws: Adopted your suggestion for using a `toml_value_map` dict with some needed changes, and rebasing on main (in which this code has been changed). See revised [cd44db9](https://github.com/tweag/FawltyDeps/pull/300/commits/cd44db9ed259b358338a61341975c8990a2df9f6).

fawltydeps/settings.py
```diff
@@ -242,8 +242,13 @@ def print_toml_config(settings: Settings, out: TextIO = sys.stdout) -> None:
         "# Copy this TOML section into your pyproject.toml to configure FawltyDeps",
         "# (default values are commented)",
         "[tool.fawltydeps]",
-    ] + [
-        f"{'# ' if has_default_value[name] else ''}{name} = {value!r}"
-        for name, value in simple_settings.items()
     ]
+    for name, value in simple_settings.items():
+        if value is None:
+            continue
```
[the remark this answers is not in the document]
Nour-Mws: Fixed to not skip on main. See my comment below.

fawltydeps/settings.py
```diff
@@ -232,15 +232,32 @@ def _option_to_toml(name, value) -> str:  # type: ignore
...
                 toml_option += "".join(
-                    [f"\n{comment_char}{k} = {v!r}" for k, v in value.items()]
+                    [
+                        f"\n{prefix}{k} = {toml_value_map.get(str(v), repr(v))}"
+                        for k, v in value.items()
+                    ]
                 )
         else:
-            toml_option = f"{comment_char}{name} = {value!r}"
+            toml_value = toml_value_map.get(str(value), repr(value))
```
jherland, line 259: These two places where we call `toml_value_map.get(str(value), repr(value))` should ideally be `toml_value_map.get(value, repr(value))` instead (i.e. without the `str(...)`).

However, I suspect there's a good reason for having to string-ify this?

If so, I think it's better to use `repr(...)` instead of `str(...)` here: `toml_value_map.get(repr(value), repr(value))`, as that would map the `--pyenv None` corner-case I discuss above to `"None"` instead of `None`.

(Maybe worth adding a test case for `--pyenv None`?)

jherland (approved):
Looks good! AFACIS, only one corner case remains to be resolved.

fawltydeps/settings.py
```diff
@@ -232,15 +232,32 @@ def _option_to_toml(name, value) -> str:  # type: ignore
         Options that are of dictionary type must be given a section entry.
         Assumption: dictionaries options are not nested.
         """
-        comment_char = "# " if has_default_value[name] else ""
+        toml_value_map = {
+            "None": "...",  # always commented, see sanity check below
+            "False": "false",
+            "True": "true",
+            # add more values here for which repr(value) is not valid TOML
+        }
```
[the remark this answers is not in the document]
Nour-Mws: That's correct. The `toml_value_map` is keyed on the values you get for the settings options, and some of those are not hashable (e.g. lists), so I went the easy way of converting them to strings first. I suppose I can do a try-catch to avoid elaborate checking of types before using `toml_value_map`. What do you think?

fawltydeps/settings.py
Nour-Mws: I could replace `str` by `repr`. but as you have underlined, that would still map `None` to `"None"`, which I understand from your comment above is not the behavior we're expecting.

jherland: I think it would work because `toml_value_map` contains the `None` string, but it does not contain the _quoted_ `None` string, i.e.:
```python
{
    "None": "...",  # This is what repr(None) maps to
    "'None'": NotImplemented,  # This is what repr("None") maps to...
}
```
Since the second is not in the dict, we would fallback to the default handling, which is what we want for `--pyenv None`.

But this is all irrelevant if we're instead going for the try-except solution above.

Nour-Mws: Replaced `str` with `repr` in [2790667](https://github.com/tweag/FawltyDeps/pull/300/commits/2790667c6fce15b40584d6dd646bac5429ba30a4) as per our discussion.

fawltydeps/settings.py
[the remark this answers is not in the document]
Nour-Mws: Opted to go with `repr` instead of try-catching.

PR #303, opened 2023-04-12
Rename run_fawltydeps to run_fawltydeps_subprocess

mknorps:
Following conversation with @jherland in https://github.com/tweag/FawltyDeps/pull/271#discussion_r1163887987 a name change of `run_fawltydeps` to `run_fawltydeps_subprocess` to follow the same naming convention as with `run_fawltydeps_function`.

The change is done separately from #271 for more atomic PRs.


jherland (approved):
🚀

PR #306, opened 2023-04-14
User defined mapping: add option and handling for multiple mapping files

mknorps:
Extend `--custom-mapping-file` option to handle more than one mapping file.

With this change, improved handling of repeated package entry in custom mapping is introduced. We use normalized package names (as they are used only for debug messages) and prepare a `CustomMapping` dictionary from all sources of the custom mapping, be it files or entries from falwtydeps config.

Commits may be read one by one:
- 9cb16299c502dd80214506db58364b888c4dcbef introduces the cli option and do not change behavior of current implementation
- 44011a0ffc0363c640dc5b834405da3e2ef7e7f2 adds actual handling of multiple files
- 44011a0ffc0363c640dc5b834405da3e2ef7e7f2 adds tests for combination of different mapping inputs
- 0a0546ae977718830a643a8fb8a5c7b965230e7e better handling of multiple packages mapping in input files (following a team meeting conversation with @Nour-Mws and @jherland)


fawltydeps/packages.py
```diff
@@ -116,68 +117,80 @@ class UserDefinedMapping(BasePackageResolver):
...
-                        *imports, mapping=DependenciesMapping.USER_DEFINED
-                    )
-                else:
-                    mapping[normalised_name] = Package(
-                        name,
-                        {DependenciesMapping.USER_DEFINED: set(imports)},
-                    )
+            custom_mapping = self.add_mapping_dicts(
+                self.custom_mapping, custom_mapping_from_files
+            )
```
jherland, line 183: With my suggested `accumulate_mappings()` above, I think we can make this both lazier and more efficient by passing a lazily-computed iterator of mappings to a single `accumulate_mappings()` call:

```python
        def _custom_mappings() -> Iterator[CustomMapping]:
            if self.custom_mapping is not None:
                logger.debug("Applying user-defined mapping from settings.")
                    yield self.custom_mapping

            if self.mapping_paths is not None:
                for path in self.mapping_paths:
                    logger.debug(f"Loading user-defined mapping from {path}")
                    with open(path, "rb") as mapping_file:
                        yield tomllib.load(mapping_file)

        custom_mapping = accumulate_mappings(_custom_mappings())
```

fawltydeps/packages.py
```diff
@@ -116,68 +117,80 @@ class UserDefinedMapping(BasePackageResolver):
...
 
-        return mapping
+        # package name is normalised at this stage, `add_mapping_dicts` was called on
+        # input files and mapping from the configuration
```
jherland, line 186: As we discussed in the meeting, I think `Package.name` is not used for anything important, so I think this comment can be dropped with no ill effect. Maybe rename the `name` variable below to `normalized_name` instead?

jherland (approved):
Looks mostly good to me. My most important question is whether we can drop the `Optional` from `Optional[Set[Path]]`: I think an empty set will achieve the same as `None`?

fawltydeps/packages.py
```diff
@@ -142,20 +142,22 @@ def packages(self) -> Dict[str, Package]:
         This enumerates the available packages  _once_, and caches the result for
         the remainder of this object's life in _packages.
         """
-        custom_mapping_from_file: Dict[str, List[str]] = {}
+        custom_mapping_from_files: Dict[str, List[str]] = {}
```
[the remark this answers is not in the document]
mknorps: No, changed to use more descriptive `CustomMapping` type.

fawltydeps/packages.py
mknorps: applied

Issue #307, opened 2023-04-14
pyproject.toml: False positive: colorama


pawamoy:
**Describe the bug**
Hey, nice project! I'm running this in one of my own, and `fawltydeps` says `colorama` is not declared, while it is. It also says it is unused (??), while it is as well.

**To Reproduce**
```bash
git clone https://github.com/pawamoy/griffe
cd griffe
fawltydeps src --deps pyproject.toml
```

Output:
```console
% fawltydeps src --deps pyproject.toml --detailed

These imports appear to be undeclared dependencies:
- 'colorama' imported at:
    src/griffe/cli.py:26
    src/griffe/diff.py:9
- 'typing_extensions' imported at:
    src/griffe/docstrings/dataclasses.py:15
    src/griffe/docstrings/google.py:42
    src/griffe/docstrings/numpy.py:79

These dependencies appear to be unused (i.e. not imported):
- 'aiofiles' declared in:
    pyproject.toml
- 'colorama' declared in:
    pyproject.toml
```

```console
% git grep colorama
pyproject.toml:    "colorama>=0.4",
src/griffe/cli.py:import colorama
src/griffe/cli.py:    colorama.deinit()
src/griffe/cli.py:    colorama.init()
src/griffe/diff.py:from colorama import Fore, Style
```

**Expected behavior**
`colorama` properly detected as used and declared.

**Environment**
 - OS name + version: 6.2.10-arch1-1
 - Version of the code: 0.8.0

mknorps:
Hi @pawamoy !

Thank you for submitting the issue. 

I see you are using ArchLinux. Unfortunately, I cannot reproduce this problem on Ubuntu, and in our CI we test on Ubuntu-22.04 , which is why we might have missed the underlying problem.
The output on Ubuntu is, as expected:
```
fawltydeps src --deps pyproject.toml --detailed    

These imports appear to be undeclared dependencies:
- 'typing_extensions' imported at:
    src/griffe/docstrings/dataclasses.py:15
    src/griffe/docstrings/google.py:42
    src/griffe/docstrings/numpy.py:79

These dependencies appear to be unused (i.e. not imported):
- 'aiofiles' declared in:
    pyproject.toml
```

Could you run fawltydeps with verbose option so we can compare logs of running on those two systems?
```
fawltydeps src --deps pyproject.toml --detailed -vv
```

[...]

PR #312, opened 2023-04-18
Add support for PEP582 `__pypackages__` directories

jherland:
Even though [PEP582](https://peps.python.org/pep-0582/) was recently rejected, it still has a significant user base via tools like [PDM](https://github.com/pdm-project/pdm) (see issue #307), and it's worth supporting these as long as the added burden is not great.

The layout of `__pypackages__` directories is different from what we currently use to identify a Python environment: Where we expect to find:

```
.../bin/python
.../lib/pythonX.Y/site-packages/<packages installed here>
```

there is instead:

```
__pypackages__/X.Y/bin/    (but NO python!)
__pypackages__/X.Y/lib/<packages installed here>
```

Since our `LocalPackageResolver.pyenv_path` member already points directly to the directory holding the packages (the `site-packages` dir in the former case), we can fairly easily add support for `__pypackage__` directories, simply by providing an alternative check for the `__pypackages__` directory structure.

This should provide a fix/workaround for issue #307 by allowing `--pyenv path/to/__pypackages__` to be passed on the command-line. FawltyDeps should then be able to find the appropriate deps, and correctly match imports and declared deps as a result.

Note that we still require the venv/`__pypackage__` dir to be passed with `--pyenv`; auto-detection of environments (both kinds) is still a planned feature.

**Commits**:
- `packages`: Log path to local packages as a `DEBUG` message
- `packages`: Support PEP582 `__pypackages__` directories for `--pyenv`


mknorps (approved):
You described the solution in detail, and reference the issue it is relevant for.
Tests are clear and looking into [PEP582](https://peps.python.org/pep-0582/) and [PDM documentation](https://pdm.fming.dev/latest/usage/pep582/), all is covered.

:rocket:

PR #318, opened 2023-04-26
`LocalPackageResolver`: Handle multiple package dirs inside one Python env

jherland:
Until now, the `--pyenv` option has accepted any path that points inside a Python environment, e.g.:
 - `.venv/`
 - `.venv/bin`
 - `.venv/bin/python`
 - `.venv/lib`
 - `.venv/lib/python3.11`
 - `.venv/lib/python3.11/site-packages`

These all unambiguously point to the Python environment rooted at `.venv/`, and they will all result in `LocalPackageResolver` looking for packages inside `.venv/lib/python3.11/site-packages`.

But this is all assuming that `.venv/lib/python3.11/site-packages` is the _only_ directory inside that Python environment that can hold Python packages.

We could have a situation where there are _multiple_ `site-packages` dirs inside `.venv/`:

 - `.venv/lib/python3.9/site-packages`
 - `.venv/lib/python3.11/site-packages`

In this case[^1], things start to go off the rails. Before this commit, we would first follow parents until `.venv/` was found, and then we would use the first result of `path.glob("lib/python?.*/site-packages")`, in other words, they would end up using `.venv/lib/python3.9/site-packages`.

This is _maybe_ not what the user intended if they passed in `.venv/` or `.venv/lib`, but it's _certainly_ not what the user wanted if they passed in a full/versioned path like `.venv/lib/python3.11/site-packages`.

There are two things to be fixed here, and this commit fixes both:

 1. When passing the path to a Python environment, there might be _multiple_ package dirs hiding inside. In other words:

        determine_package_dir(cls, path: Path) -> Optional[Path]

    is NOT a suitable signature, we rather need the more general:

        find_package_dirs(cls, path: Path) -> Iterator[Path]

    signature, which is able to return zero, one, or more results. And the surrounding code should handle this by adding _all_ matching package dirs to the `LocalPackageResolver.package_dirs`.

 2. We should also _filter_ the package dirs found inside the given Python environment to _exclude_ those that are not relative to the given path. That is, if you pass `.venv/lib/python3.11/site-packages`, you should only get `.venv/lib/python3.11/site-packages` back, and not `.venv/lib/python3.9/site-packages`, even if that also exists within the same Python environment.

As a result, a user can now specify `--pyenv .venv` to get _all_ Python packages within any `.venv/lib/python?.*/site-packages/` directory, OR they can specify `--pyenv .venv/lib/python3.11/site-packages` to be very explicit about which package directory they do want.

The only "casualty" of this change is that specifying a path within the environments that is _not_ a parent of any package dir (for example: `--pyenv .venv/bin/python`) is no longer supported. This was already a corner case, and I don't think we lose much by no longer supporting it.

[^1]: This whole argument is somewhat strained when using `.venv/` virtualenvs as an example: virtualenvs typically only exist within the context of a _single_ Python version (the one in `.venv/bin/python`), and there is always only one `site-packages` dir in a virtualenv. However, the same argument applies equally to system environments under e.g. `/usr/` or `/usr/local/`, and they certainly apply to `__pypackages__` environments that are not bound to a specific interpreter.

### DIFF
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -6,7 +6,7 @@ import sys
 import tempfile
 import venv
 from abc import ABC, abstractmethod
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from dataclasses import dataclass, replace
 from functools import partial
 from pathlib import Path
@@ -203,24 +203,25 @@ class LocalPackageResolver(BasePackageResolver):
         """
         self.package_dirs: Set[Path] = set()  # empty => use sys.path instead
         for path in pyenv_paths:
-            package_dir = self.determine_package_dir(path)
-            if package_dir is None:
+            package_dirs = set(self.find_package_dirs(path))
+            if not package_dirs:
                 logger.warning(f"Could not find a Python env at {path}!")
-            else:
-                self.package_dirs.add(package_dir)
+            self.package_dirs.update(package_dirs)
         if pyenv_paths and not self.package_dirs:
             raise ValueError(f"Could not find any Python env in {pyenv_paths}!")
         # We enumerate packages for pyenv_path _once_ and cache the result here:
         self._packages: Optional[Dict[str, Package]] = None
 
     @classmethod
-    def determine_package_dir(cls, path: Path) -> Optional[Path]:
-        """Return the site-packages directory corresponding to the given path.
+    def find_package_dirs(cls, path: Path) -> Iterator[Path]:
+        """Return the packages directories corresponding to the given path.
 
         The given 'path' is a user-provided directory path meant to point to
         a Python environment (e.g. a virtualenv, a poetry2nix environment, or
-        similar). Deduce the appropriate site-packages directory from this path,
-        or return None if no environment could be found at the given path.
+        similar). Deduce the appropriate package directories inside this path,
+        and yield them.
+
+        Yield nothing if no package dirs was found at the given path.
         """
         # We define a "valid Python environment" as a directory that contains
         # a bin/python file, and a lib/pythonX.Y/site-packages subdirectory.
@@ -229,20 +230,31 @@ class LocalPackageResolver(BasePackageResolver):
         # From there, the returned directory is that site-packages subdir.
         # Note that we must also accept lib/pythonX.Y/site-packages for python
         # versions X.Y that are different from the current Python version.
+        found = False
         if (path / "bin/python").is_file():
             for site_packages in path.glob("lib/python?.*/site-packages"):
                 if site_packages.is_dir():
-                    return site_packages
+                    yield site_packages
+                    found = True
+            if found:
+                return
 
         # Workaround for projects using PEP582:
         if path.name == "__pypackages__":
             for site_packages in path.glob("?.*/lib"):
                 if site_packages.is_dir():
-                    return site_packages
+                    yield site_packages
+                    found = True
+            if found:
+                return
 
         # Given path is not a python environment, but it might be _inside_ one.
         # Try again with parent directory
-        return None if path.parent == path else cls.determine_package_dir(path.parent)
+        if path.parent != path:
+            for package_dir in cls.find_package_dirs(path.parent):
+                with suppress(ValueError):
+                    package_dir.relative_to(path)  # ValueError if not relative
+                    yield package_dir
 
     def _from_one_env(
         self, env_paths: List[str]
--- a/tests/test_local_env.py
+++ b/tests/test_local_env.py
@@ -1,6 +1,8 @@
 """Verify behavior of package module looking at a given Python environment."""
 import sys
 import venv
+from pathlib import Path
+from typing import Iterator
 
 import pytest
 
@@ -18,7 +20,6 @@ major, minor = sys.version_info[:2]
 # that they might point at (and that we should accept)?
 env_subdirs = [
     "",
-    "bin",
     "lib",
     f"lib/python{major}.{minor}",
     f"lib/python{major}.{minor}/site-packages",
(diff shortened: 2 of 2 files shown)


fawltydeps/packages.py
```diff
@@ -229,20 +230,31 @@ def determine_package_dir(cls, path: Path) -> Optional[Path]:
...
+                return
 
         # Given path is not a python environment, but it might be _inside_ one.
         # Try again with parent directory
-        return None if path.parent == path else cls.determine_package_dir(path.parent)
+        if path.parent != path:
+            for package_dir in cls.find_package_dirs(path.parent):
+                with suppress(ValueError):
+                    package_dir.relative_to(path)  # ValueError if not relative
+                    yield package_dir
```
mknorps, line 257: Why do we need this mechanism?

`package_dir.relative_to(path)` looks like it was called but its outputs were never used.

tests/test_local_env.py
```diff
@@ -58,23 +57,62 @@ def test_determine_package_dir__various_paths_in_poetry2nix_env(
...
+        }
+    )
+    path = tmp_path / subdir
+    actual = set(LocalPackageResolver.find_package_dirs(path))
+
+    def expected_package_dirs(base: Path, subdir: str) -> Iterator[Path]:
+        # When given:
+        # __pypackages__/  -->  return all .../X.Y/lib dirs
+        # __pypackages__/X.Y/*  -->  return the only matching .../lib dir
+        # .venv/  -->  return all .../lib/pythonX.Y/site-packages dirs
```
mknorps, line 103: There is a mismatch between the description and the code. There is an assumption that `all` `.../X.Y/lib` dirs mean `{major}.{minor}` and `{major}.{minor+1}`. It may be enough to rewrite the code below as suggested in the comment below and remove the description if you fid the suggestion useful.

tests/test_local_env.py
```diff
@@ -58,23 +57,62 @@ def test_determine_package_dir__various_paths_in_poetry2nix_env(
...
+        if subdir.startswith("__pypackages__"):
+            yield base / f"__pypackages__/{major}.{minor}/lib"
+            if f"{major}.{minor}" not in subdir:
+                yield base / f"__pypackages__/{major}.{minor + 1}/lib"
+        else:
+            assert subdir.startswith(".venv/")
+            yield base / f".venv/lib/python{major}.{minor}/site-packages"
+            if f"{major}.{minor}" not in subdir:
+                yield base / f".venv/lib/python{major}.{minor + 1}/site-packages"
+
```
mknorps, line 114:
```suggestion
        is_version_subdir = f"{major}.{minor}" not in subdir
        if subdir.startswith("__pypackages__"):
            yield base / f"__pypackages__/{major}.{minor}/lib"
            if is_version_subdir:
                yield base / f"__pypackages__/{major}.{minor + 1}/lib"
        elif subdir.startswith(".venv/"):
            yield base / f".venv/lib/python{major}.{minor}/site-packages"
            if is_version_subdir:
                yield base / f".venv/lib/python{major}.{minor + 1}/site-packages"
        else:
            raise ValueError("Unsupported subdir")

```

mknorps (approved):
Thank you @jherland !
It looks like good handling of  multiple Python version environment handling. I think we are very good at making sure that many corner cases are covered in FawltyDeps :100:

I have some small suggestions and a question if we need the suppress mechanism. Otherwise, I think this is good to go :)

mknorps (approved):
Thank you @jherland !
It looks like good handling of  multiple Python version environment handling. I think we are very good at making sure that many corner cases are covered in FawltyDeps :100:

I have some small suggestions and a question if we need the suppress mechanism. Otherwise, I think this is good to go :)

tests/test_local_env.py
jherland: I wrote both the comment and the test with the assumption that `X.Y` refers to _any_ Python version, no matter what the _current_ Python version happens to be. To that end I use `{major}.{minor}` and `{major}.{minor+1}` to mean "two different Python version, one that happens to match the current version and one that doesn't".

Then, in the parametrized inputs (ultimately from `*_subdirs` above) there are both paths that _do_ include `{major}.{minor}` and paths that _don't_. For those that _do_, we expect only the _one_ matching package dir to be returned, and for those that _don't_ we expect _both_ package dirs to be returned.

I'm not sure I see the mismatch between the comment and the code, but it probably makes sense to drop the comment and make the code more readable instead.

tests/test_local_env.py
jherland: I believe the `is_version_subdir` should instead be named `is_not_version_subdir` or maybe `is_version_agnostic`?

fawltydeps/packages.py
mknorps: So to sum up this:
```python
                with suppress(ValueError):
                    package_dir.relative_to(path)  # ValueError if not relative
                    yield package_dir
```
is a hacky trick to:
```python
                if package_dir.relative_to(path)
                    yield package_dir
```
which is available only for Python 3.9+?

If yes, then I get it and indeed looks like the next best thing.

tests/test_local_env.py
mknorps: So maybe just drop this comment?

My point was that realisation of a more general `{major}.{minor1}` and `{major}.{minor2}` is done here with `major2 = major1 + 1`. Which is OK, but then the comment does not match.

tests/test_local_env.py
mknorps: `is_version_agnostic` ❤️

jherland: Fixed in e0d7c83.

tests/test_local_env.py
jherland: Dropped in e0d7c83.

PR #321, opened 2023-05-06
Support multiple `--pyenv` options

jherland:
These are the commits that were originally part of #313, but now split into this PR.

Commits:
- `LocalPackageResolver`: Accept multiple pyenvs
- Add support for multiple `--pyenv` options
- `settings.print_toml_config()`: Represent sets as sorted lists

### DIFF
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -10,7 +10,18 @@ from contextlib import contextmanager
 from dataclasses import dataclass, replace
 from functools import partial
 from pathlib import Path
-from typing import Dict, Iterable, Iterator, List, Optional, Set, Tuple, Type, Union
+from typing import (
+    AbstractSet,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Type,
+    Union,
+)
 
 # importlib_metadata is gradually graduating into the importlib.metadata stdlib
 # module, however we rely on internal functions and recent (and upcoming)
@@ -181,21 +192,24 @@ class UserDefinedMapping(BasePackageResolver):
 class LocalPackageResolver(BasePackageResolver):
     """Lookup imports exposed by packages installed in a Python environment."""
 
-    def __init__(self, pyenv_path: Optional[Path] = None) -> None:
+    def __init__(self, pyenv_paths: AbstractSet[Path] = frozenset()) -> None:
         """Lookup packages installed in the given virtualenv.
 
-        Default to the current python environment if `pyenv_path` is not given
-        (or None).
+        Default to the current python environment if `pyenv_paths` is empty
+        (the default).
 
         Use importlib_metadata to look up the mapping between packages and their
         provided import names.
         """
-        if pyenv_path is not None:
-            self.pyenv_path = self.determine_package_dir(pyenv_path)
-            if self.pyenv_path is None:
-                raise ValueError(f"Could not find a Python env at {pyenv_path}!")
-        else:
-            self.pyenv_path = None
+        self.package_dirs: Set[Path] = set()  # empty => use sys.path instead
+        for path in pyenv_paths:
+            package_dir = self.determine_package_dir(path)
+            if package_dir is None:
+                logger.warning(f"Could not find a Python env at {path}!")
+            else:
+                self.package_dirs.add(package_dir)
+        if pyenv_paths and not self.package_dirs:
+            raise ValueError(f"Could not find any Python env in {pyenv_paths}!")
         # We enumerate packages for pyenv_path _once_ and cache the result here:
         self._packages: Optional[Dict[str, Package]] = None
 
@@ -230,6 +244,42 @@ class LocalPackageResolver(BasePackageResolver):
         # Try again with parent directory
         return None if path.parent == path else cls.determine_package_dir(path.parent)
 
+    def _from_one_env(
+        self, env_paths: List[str]
+    ) -> Iterator[Tuple[CustomMapping, str]]:
+        """Return package-name-to-import-names mapping from one Python env.
+
+        This is roughly equivalent to calling importlib_metadata's
+        packages_distributions(), except that instead of implicitly querying
+        sys.path, we query env_paths instead.
+
+        Also, we are able to return packages that map to zero import names,
+        whereas packages_distributions() cannot.
+        """
+        seen = set()  # Package names (normalized) seen earlier in env_paths
+
+        # We're reaching into the internals of importlib_metadata here, which
+        # Mypy is not overly fond of, hence lots of "type: ignore"...
+        context = DistributionFinder.Context(path=env_paths)  # type: ignore
+        for dist in MetadataPathFinder().find_distributions(context):
+            normalized_name = Package.normalize_name(dist.name)
+            parent_dir = dist.locate_file("")
+            if normalized_name in seen:
+                # We already found another instance of this package earlier in
+                # env_paths. Assume that the earlier package is what Python's
+                # import machinery will choose, and that this later package is
+                # not interesting.
+                logger.debug(f"Skip {dist.name} {dist.version} under {parent_dir}")
+                continue
+
+            logger.debug(f"Found {dist.name} {dist.version} under {parent_dir}")
+            seen.add(normalized_name)
+            imports = list(
+                _top_level_declared(dist)  # type: ignore
+                or _top_level_inferred(dist)  # type: ignore
+            )
+            yield {dist.name: imports}, str(parent_dir)
+
     @property
     @calculated_once
     def packages(self) -> Dict[str, Package]:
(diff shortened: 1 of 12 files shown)


tests/test_cmdline.py
```diff
@@ -710,6 +710,25 @@ def test_check_detailed__simple_project_in_fake_venv__resolves_imports_vs_deps(
...
 
+def test_check_detailed__simple_project_w_2_fake_venv__resolves_imports_vs_deps(
+    fake_venv, project_with_code_and_requirements_txt
+):
+    tmp_path = project_with_code_and_requirements_txt(
+        imports=["some_module", "other_module", "requests"],
+        declares=["pandas", "some_module"],
+    )
+    venv_dir1, _ = fake_venv({"some_module": {"some_module"}})
+    venv_dir2, _ = fake_venv({"pandas": {"requests"}, "some_module": {"other_module"}})
```
Nour-Mws, line 721: Nitpick alert: I found the package names and their exposed import names confusing to read. Is there a reason to not:
- either use actual package names even if this is a fake env?
- or use completely made up names for all packages and imports?

Seeing `pandas` expose `requests` is a tiny bit jarring.

tests/test_settings.py
```diff
@@ -275,6 +275,14 @@ class SettingsTestVector:
             custom_mapping_file={Path("mapping.toml")},
         ),
     ),
+    SettingsTestVector(
+        "config_file_with_pyenvs_and_cli__cli_pyenvs_overrides_config",
+        config=dict(pyenvs=["foo", "bar"]),
+        cmdline=dict(pyenvs=["baz", "xyzzy"]),
```
Nour-Mws, line 281: Are we also defaulting to override config with cli arguments for other options that might have multiple values?

tests/test_cmdline.py
```diff
@@ -904,6 +904,30 @@ def test_cmdline_on_ignored_undeclared_option(
...
+                # ignore_undeclared = []
+                # ignore_unused = []
+                # deps_parser_choice = ...
+                # install_deps = false
+                # verbosity = 0
+                # [tool.fawltydeps.custom_mapping]
+                """
+            ).splitlines(),
+            id="generate_toml_config_with_multiple_pyenvs",
+        ),
```
Nour-Mws, line 930: A general question on the code below. Apologies in advance as this hasn't been actually touched by this PR:
```python
    tmp_path = project_with_code_and_requirements_txt(
        imports=["requests"],
        declares=["pandas"],
    )
    setup_fawltydeps_config(config)
    output, errors, returncode = run_fawltydeps_subprocess(
        "--config-file=pyproject.toml", *args, cwd=tmp_path
    )
```
This seems to assume that the `tmp_path` returned by `project_with_code_and_requirements_txt()` and the one used by `setup_fawltydeps_config()` are the same. Is that the case? And if so, why?

fawltydeps/packages.py
```diff
@@ -181,21 +192,26 @@ def lookup_packages(self, package_names: Set[str]) -> Dict[str, Package]:
...
-            self.pyenv_path = None
+        self.package_dirs: Set[Path] = set()  # empty => use sys.path instead
+        if pyenv_paths:
+            package_dirs = {
+                (path, self.determine_package_dir(path)) for path in pyenv_paths
+            }
+            for pyenv_path, package_dir in package_dirs:
+                if package_dir is None:
+                    logger.warning(f"Could not find a Python env at {pyenv_path}!")
+            self.package_dirs = {dir for _, dir in package_dirs if dir is not None}
```
Nour-Mws, line 212: I found this code a touch too indirect and hard to follow. Would the following make sense?
```suggestion
      for path in pyenv_paths:
          package_dir = self.determine_package_dir(path)
          if package_dir is None:
              logger.warning(f"Could not find a Python env at {path}!")
          else:
              self.package_dirs.add(package_dir)
```

Nour-Mws (commented):
Looks great as usual!
There's nothing that blocks this PR per se. Just a couple small questions that I hope will be straightforward. I am also fine either way for the suggestions I made.

fawltydeps/packages.py
jherland: Yes, that reads much better. Thanks! Will adopt, along with a changing the two lines below:
```python
            if not self.package_dirs:
                raise ValueError(f"Could not find any Python env in {pyenv_paths}!")
```
into
```python
        if pyenv_paths and not self.package_dirs:
            raise ValueError(f"Could not find any Python env in {pyenv_paths}!")
```
as an empty `self.package_dirs` is acceptable, but only when `pyenv_paths` was already empty.

tests/test_cmdline.py
jherland: Agreed. I believe what I was thinking about when I wrote this was to demonstrate (along with grabbing packages from multiple pyenvs) that you could have imports and declared deps that don't seem to line up at all, but that - given the right `Package` object - actually lines up perfectly in the end. However, using `pandas` and `requests` to demonstrate this is just stupid on my part. 🤪 Will fix!

tests/test_cmdline.py
jherland: Yes, in the end all of this boils down to the [`tmp_path` fixture from pytest](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture): If I use multiple fixtures in a test, and each of those uses `tmp_path`, then it's all the _same_ `tmp_path`; it is instantiated _once per test_:

```python
@pytest.fixture
def fixtureA(tmp_path):
    return tmp_path

@pytest.fixture
def fixtureB(tmp_path):
    return tmp_path

def test_case(fixtureA, fixtureB):
    assert fixtureA == fixtureB
```

Both `project_with_code_and_requirements_txt()` and `setup_fawltydeps_config()` are fixtures that rely on the pytest `tmp_path` fixture (via their use of `write_tmp_files()`. And the path returned from `project_with_code_and_requirements_txt()` is the one returned from `write_tmp_files()`, which is indeed the `tmp_path` fixture/value itself.

Thus the `tmp_path` variable in the test below has the exact same value as if we had just used the `tmp_path` fixture directly in the test case.

The test case, as written, is a bit indirect and convoluted, for sure, but I don't think it's _misleading_?

This is maybe clearer, although a bit more verbose?

```python
    tmp_path = project_with_code_and_requirements_txt(
        imports=["requests"],
        declares=["pandas"],
    )
    config_path = setup_fawltydeps_config(config)
    output, errors, returncode = run_fawltydeps_subprocess(
        f"--config-file={config_path.relative_to(tmp_path)}",
        *args,
        cwd=tmp_path,
    )
```

tests/test_settings.py
jherland: Yes, here's what happens for e.g. `.code` (using environment vs. CLI, but the same holds for config file vs. CLI):
```sh
$ fawltydeps_code='["fawltydeps/main.py", "fawltydeps/packages.py"]' \
  fawltydeps --list-imports --detailed
fawltydeps/main.py:20: pydantic
fawltydeps/packages.py:20: importlib_metadata
fawltydeps/packages.py:37: tomli

$ fawltydeps_code='["fawltydeps/main.py", "fawltydeps/packages.py"]' \
  fawltydeps --list-imports --detailed --code fawltydeps/types.py fawltydeps/utils.py
fawltydeps/types.py:15: typing_extensions
fawltydeps/utils.py:9: importlib_metadata
```

When I first wrote the `Settings` class, I added the following TODO (and it's still there):
```
    TODO? Currently overrides happen _completely_, for example specifying
    --ignore-undeclared on the command-line will _replace_ an ignore_undeclared
    directive in pyproject.toml. We may want to consider allowing some
    directives to _combine_ (although that will carry further complications).
```

The "further complications" I allude to here are more complex than one might imagine at first:
- We would probably need to introspect over the `Settings` members and/or their type and define _different_ ways of combining them:
  - Most sets (`code`/`deps`/`pyenvs`/`custom_mapping_file`/`ignore_*`) combine by _union_
  - `custom_mapping` combines by _merging dicts_ (a la `accumulate_mappings()`)
  - How do we want `actions` to combine? One could argue for either _replace_ or _union_, but probably _replace_ makes most sense.
  - `output_format` should likely _replace_?
  - Same with `deps_parser_choice`?
  - The `verbosity`number should probably `sum()` across levels?
- When we decide to take the _union_ of multi-valued members, then how do we deal with wanting to _remove_ something that was added at an earlier stage? For example: the config file says `code = ["foo.py", "bar.py"]`, but I want FD to look at code from `bar.py` and `baz.py`; I can add `baz.py` at the command-line, but how do I _remove_ `foo.py`?
- We would maybe need to add more options (`--remove-code=foo.py` or `--no-code=foo.py`) or more complex option parsing (`--code "+baz.py" "-foo.py"`) to deal with this.
- How to deal with different options that "cooperate" in unexpected ways? E.g. `deps = ["pyproject.toml"]` in the config file, but `--deps my_reqs.pip --deps-parser-choice requirements.txt` on the command line would probably combine into the following nonsense in `Settings`:
    ```
    deps = {"pyproject.toml", "my_reqs.pip"}
    deps_parser_choice = ParserChoice.REQUIREMENTS_TXT
    ```

When I first wrote this code in #167, I decided to _not_ open this Pandora's box of complexity, and to leave everything as strictly overriding/replacing for now.

After revisiting it now, I still think this is the sanity-preserving choice! 😅

fawltydeps/packages.py
jherland: Fixed in ed8b82a

tests/test_settings.py
Nour-Mws: I agree with that analysis. Could you post it in an issue as it's very useful and would otherwise be lost.

tests/test_cmdline.py
Nour-Mws: #TIL!
The new code you posted is as you said more verbose without necessarily making it clearer that we're using the same `tmp_path` as the fixture is called once.

Nour-Mws (approved):
Approving :tada: 
I added comments to two conversations and resolved them.

tests/test_settings.py
jherland: https://github.com/tweag/FawltyDeps/issues/328 ✅

PR #323, opened 2023-05-08
`TemporaryPipInstallResolver`: Handle `pip install` errors gracefully

### LINKED ISSUE #322: Handle `pip install` errors in `TemporaryPipInstallResolver`
## Problem
If `LocalPackageResolver` is used and installation of some packages locally fails, an uncaught `CalledProcessError` is raised. 

## Expected outcome
Handle this error properly and give the user a list of the problematic packages.

## Background and discussion

The original discussion sparked in https://github.com/tweag/FawltyDeps/pull/320#discussion_r1186052929

@jherland 
I would hope that we could produce a `logger.warning()` or `.error()` from `TemporaryPipInstallResolver` when the `pip install` fails for whatever reason, as I suspect that failure is often more relevant to the issue at hand.

@mknorps 

Here we will produce a [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError):
```python 
            subprocess.run(
                [f"{venv_dir}/bin/pip", "install", "--no-deps"] + requirements,
            )
```
and in fact we do not handle it anywhere :see_no_evil: . What do you think about moving handling of this error to another issues? It does not seem so in need for now.

@jherland 

It seems that we might need to think a bit more about how to best handle errors from `pip install`. Here's what I found after digging for a little bit:
- We currently try to install _all_ of `requirements` with a single `pip install` command. This is ~necessary for performance reasons, I believe (to prevent running one subprocess per requirement/dependency).
- If the `pip install` fails, we currently don't know whether that means _nothing_ was installed, or if _some_ of the requirements might have been installed (but others failed).
- I can think of scenarios where _nothing_ would be installed (e.g. network issues) and other scenarios where almost all deps were installed (e.g. one of the deps doesn't exist due to e.g. a typo).
- We _could_ play games with `pip` to try to figure out which dependencies were or were not installed.
  - e.g. using the new `--report` option to get a [detailed JSON report](https://pip.pypa.io/en/stable/reference/installation-report/), or
  - using `pip inspect` afterwards to get a similar JSON report, or
  - using the simpler (and backwards-compatible) `pip list` to get a list of installed dependencies.
- _However_, we already rely on `LocalPackageResolver` to look at the virtualenv and return the installed packages for us, so we're probably better off ignoring any `pip install` failure and leave `LocalPackageResolver` to tell us which packages were installed or not.
- Maybe the safest approach is something like this?:
  1. First try to install all dependencies with a single `pip install` command. If this succeeds, then skip directly to step 3.
  2. If the above fails, fall back to `pip install` one dependency at a time. (This might be very slow, but will hopefully be skipped most of the time.)
  3. Use the existing `LocalPackageResolver` code to resolve what can be resolved, and leave the rest as unresolved.

jherland:
Until now, we have not handled the `CalledProcessError` that is raised when `pip install` fails to install the given requirements in `TemporaryPipInstallResolver.temp_installed_requirements()`.

This commit adds two new flags to `.temp_installed_requirements()`:

 - `ignore_errors`: This disables the `CalledProcessError` and logs a warning message instead.
 - `retry_each`: In combination with the above, when `pip install` fails, this will retry `pip install` on each individual requirement in order. The effect of this will be to install the subset of the given requirements that we _can_ install, leaving only the problematic requirements uninstalled.

Using both of these flags enables `TemporaryPipInstallResolver` to do a "best effort" installation of dependencies: Inside the `.temp_installed_requirements()` context, the `LocalPackageResolver` will be able to resolve those dependencies that were successfully installed, and the dependencies we failed to install will be handed off to the next level of the resolver stack.

As of writing this is always the `IdentityMapping`, but this is about to go away, and instead a proper `UnresolvedDependenciesError` will be raised. Together with the warning messages from the failed `pip install` commands, this should inform the user of exactly which packages/dependencies we failed to install, and why.

On the `tests/` side, move our existing `TemporaryPipInstallResolver` test into a new test module: `test_install_deps`, and add two more tests:

 - to test that we no longer propagate a `CalledProcessError` when `pip install` fails (but instead log a warning message), and
 - to test that we are able to install one dependency with `pip install` even if another dependency fails.

Fixes #322 (Handle `pip install` errors in `TemporaryPipInstallResolver`)

### DIFF
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -317,14 +317,25 @@ class TemporaryPipInstallResolver(BasePackageResolver):
     def installed_requirements(
         venv_dir: Path, requirements: List[str]
     ) -> Iterator[Path]:
-        """Install the given requirements into venv_dir with `pip install`."""
+        """Install the given requirements into venv_dir with `pip install`.
+
+        We try to install as many of the given requirements as possible. Failed
+        requirements will be logged with warning messages, but no matter how
+        many failures we get, we will still enter the caller's context. It is
+        up to the caller to handle any requirements that we failed to install.
+        """
         marker_file = venv_dir / ".installed"
         if not marker_file.is_file():
             venv.create(venv_dir, clear=True, with_pip=True)
-        subprocess.run(
-            [f"{venv_dir}/bin/pip", "install", "--no-deps"] + requirements,
-            check=True,  # fail if any of the commands fail
-        )
+        argv = [f"{venv_dir}/bin/pip", "install", "--no-deps"]
+        proc = subprocess.run(argv + requirements, check=False)
+        if proc.returncode:  # pip install failed
+            logger.warning("Command failed: %s", argv + requirements)
+            logger.info("Retrying each requirement individually...")
+            for req in requirements:
+                proc = subprocess.run(argv + [req], check=False)
+                if proc.returncode:  # pip install failed
+                    logger.warning("Failed to install %s", repr(req))
         marker_file.touch()
         yield venv_dir
 
@@ -336,15 +347,16 @@ class TemporaryPipInstallResolver(BasePackageResolver):
         Provide a path to the temporary venv into the caller's context in which
         the given requirements have been `pip install`ed. Automatically remove
         the venv at the end of the context.
+
+        Installation is done on a "best effort" basis as documented by
+        .installed_requirements() above. The caller is expected to handle any
+        requirements that we failed to install.
         """
         with tempfile.TemporaryDirectory() as tmpdir:
             with cls.installed_requirements(Path(tmpdir), requirements) as venv_dir:
                 yield venv_dir
 
-    def lookup_packages(
-        self,
-        package_names: Set[str],
-    ) -> Dict[str, Package]:
+    def lookup_packages(self, package_names: Set[str]) -> Dict[str, Package]:
         """Convert package names into Package objects via temporary pip install.
 
         Use the temp_installed_requirements() above to `pip install` the given
--- a/tests/test_install_deps.py
+++ b/tests/test_install_deps.py
@@ -1,0 +1,63 @@
+"""Verify behavior of TemporaryPipInstallResolver."""
+import logging
+
+from fawltydeps.packages import (
+    IdentityMapping,
+    Package,
+    TemporaryPipInstallResolver,
+    resolve_dependencies,
+)
+
+
+def test_resolve_dependencies_install_deps__via_local_cache(local_pypi):
+    debug_info = "Provided by temporary `pip install`"
+    actual = resolve_dependencies(
+        ["leftpadx", "click"], pyenv_path=None, install_deps=True
+    )
+    assert actual == {
+        "leftpadx": Package(
+            "leftpadx", {"leftpad"}, TemporaryPipInstallResolver, debug_info
+        ),
+        "click": Package("click", {"click"}, TemporaryPipInstallResolver, debug_info),
+    }
+
+
+def test_resolve_dependencies_install_deps__handle_pip_install_failure(
+    caplog, local_pypi
+):
+    # TemporaryPipInstallResolver will handle the inevitable pip install error
+    # and return to resolve_dependencies() with the missing package unresolved.
+    # For now, IdentityMapping "saves the day" and supplies a Package object.
+    # Soon, this should result in an unresolved package error instead.
+    caplog.set_level(logging.WARNING)
+    actual = resolve_dependencies(
+        ["does_not_exist"], pyenv_path=None, install_deps=True
+    )
+    assert actual == {
+        "does_not_exist": Package(
+            "does_not_exist", {"does_not_exist"}, IdentityMapping
+        ),
+    }
+    assert all(word in caplog.text for word in ["pip", "install", "does_not_exist"])
+
+
+def test_resolve_dependencies_install_deps__pip_install_some_packages(
+    caplog, local_pypi
+):
+    debug_info = "Provided by temporary `pip install`"
+    caplog.set_level(logging.WARNING)
+    actual = resolve_dependencies(
+        ["click", "does_not_exist", "leftpadx"], pyenv_path=None, install_deps=True
+    )
+    # pip install is able to install "leftpadx", but "package_does_not_exist"
+    # falls through to IdentityMapping.
+    assert actual == {
+        "click": Package("click", {"click"}, TemporaryPipInstallResolver, debug_info),
+        "does_not_exist": Package(
+            "does_not_exist", {"does_not_exist"}, IdentityMapping
+        ),
+        "leftpadx": Package(
+            "leftpadx", {"leftpad"}, TemporaryPipInstallResolver, debug_info
+        ),
+    }
+    assert all(word in caplog.text for word in ["pip", "install", "does_not_exist"])
--- a/noxfile.py
+++ b/noxfile.py
@@ -94,6 +94,7 @@ def lint(session):
         "too-many-arguments",
         "too-many-instance-attributes",
         "too-many-lines",
+        "unused-argument",
         "use-implicit-booleaness-not-comparison",
     ]
     session.run(
--- a/tests/test_local_env.py
+++ b/tests/test_local_env.py
@@ -8,7 +8,6 @@ from fawltydeps.packages import (
     IdentityMapping,
     LocalPackageResolver,
     Package,
-    TemporaryPipInstallResolver,
     resolve_dependencies,
 )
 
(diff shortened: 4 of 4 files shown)


fawltydeps/packages.py
```diff
@@ -310,20 +310,45 @@ class TemporaryPipInstallResolver(BasePackageResolver):
...
+                logger.warning("Command failed: %s", argv)
+                if retry_each:
+                    logger.info("Retrying each requirement individually...")
+                    for req in requirements:
+                        proc = subprocess.run(
+                            [f"{venv_dir}/bin/pip", "install", "--no-deps", req],
+                            check=False,
+                        )
+                        if proc.returncode:  # pip install failed
+                            logger.warning("Command failed: %s", argv)
```
mknorps, line 350:
```suggestion
                        if proc.returncode:  # pip install failed
                            logger.warning("Installation failed for requirement: %s", req)
```

fawltydeps/packages.py
```diff
@@ -310,20 +310,45 @@ class TemporaryPipInstallResolver(BasePackageResolver):
...
         """
         with tempfile.TemporaryDirectory() as tmpdir:
             venv_dir = Path(tmpdir)
             venv.create(venv_dir, clear=True, with_pip=True)
-            subprocess.run(
-                [f"{venv_dir}/bin/pip", "install", "--no-deps"] + requirements,
-                check=True,  # fail if any of the commands fail
-            )
+            argv = [f"{venv_dir}/bin/pip", "install", "--no-deps"] + requirements
+            proc = subprocess.run(argv, check=not ignore_errors)
```
mknorps, line 339: I do not understand why a `CalledProcessError` is not raised, could you explain it? According to documentation, https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module:
 > If check is true, and the process exits with a non-zero exit code, a CalledProcessError exception will be raised. Attributes of that exception hold the arguments, the exit code, and stdout and stderr if they were captured.

mknorps: Oh, I know why! This case is never tested in tests, as `ignore_errors` is set to `True` :D

In that case we can either drop `ignore_errors` parameter and have:
```python
            proc = subprocess.run(argv, check=False)
```
Or test this case properly in tests, because it will result in some more guardian code around `CalledProcessError`.

I would opt for option 1, as I think that what we do here is an equivalent of handling this error.

fawltydeps/packages.py
```diff
@@ -310,20 +310,45 @@ class TemporaryPipInstallResolver(BasePackageResolver):
...
         """Create a temporary venv and install the given requirements into it.
 
         Provide a path to the temporary venv into the caller's context in which
         the given requirements have been `pip install`ed. Automatically remove
         the venv at the end of the context.
+
+        By default, if `pip install` fails, we will propagate the resulting
+        CalledProcessError from here. If 'ignore_errors' is enabled we instead
+        log a warning and continue into the caller's context. It is then up to
+        the caller to handle the missing requirements in the temporary venv.
```
mknorps, line 327: If we remove `ignore_errors` option, this desription should explain that `pip install` will stop on a first failed dependency instalation and not instal anything further.

tests/test_install_deps.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+
+
+def test_resolve_dependencies_install_deps__pip_install_some_packages(
+    caplog, local_pip_cache
+):
+    debug_info = "Provided by temporary `pip install`"
+    caplog.set_level(logging.WARNING)
+    actual = resolve_dependencies(
+        ["leftpadx", "package_does_not_exist"], pyenv_path=None, install_deps=True
```
mknorps, line 66:
```suggestion
        ["leftpadx", "package_does_not_exist", "click"], pyenv_path=None, install_deps=True
```
Add a legit package after the fake one, to make sure `pip install` did `retry_each`.

tests/test_install_deps.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+    # pip install is able to install "leftpadx", but "package_does_not_exist"
+    # falls through to IdentityMapping.
+    assert actual == {
+        "leftpadx": Package(
+            "leftpadx", {"leftpad"}, TemporaryPipInstallResolver, debug_info
+        ),
+        "package_does_not_exist": Package(
+            "package_does_not_exist", {"package_does_not_exist"}, IdentityMapping
+        ),
```
mknorps, line 76:
```suggestion
        "leftpadx": Package(
            "leftpadx", {"leftpad"}, TemporaryPipInstallResolver, debug_info
        ),
        "package_does_not_exist": Package(
            "package_does_not_exist", {"package_does_not_exist"}, IdentityMapping
        ),
        "click": Package(
            "click", {"click"}, TemporaryPipInstallResolver, debug_info
        ),
```

mknorps (changes requested):
I like the way of handling the installation errors with the process return code :+1: 

There are some improvement suggestions in the review, of which I consider `subprocess` `check` option handling the most important.

fawltydeps/packages.py
jherland: Yes, I wrote the patch to be minimally invasive in terms of preserving current behavior unless the extra flags are passed. But since there is only one caller, and it always passes both flags as `True`, we can drop the flags, and hardcode the new behavior instead. Will fix.

fawltydeps/packages.py
jherland: Ack. Will update the docstring when I hardcode the new behavior.

tests/test_install_deps.py
jherland: Agreed. However, `TemporaryPipInstallResolver.lookup_packages()` _sorts_ the package names before passing them to `.temp_installed_requirements()` so in order for this to be properly tested, we need package names that sort alphabetically before and after the missing package Will fix.

jherland:
I believe I addressed all your comments in updated commit 13ec273.

fawltydeps/packages.py
```diff
@@ -317,14 +317,27 @@ class TemporaryPipInstallResolver(BasePackageResolver):
     def installed_requirements(
         venv_dir: Path, requirements: List[str]
     ) -> Iterator[Path]:
-        """Install the given requirements into venv_dir with `pip install`."""
+        """Install the given requirements into venv_dir with `pip install`.
+
+        If `pip install` fails, we will log a warning, then retry by attempting
+        to `pip install` each given requirement in order (logging any errors
+        that occur), before we finally enter the caller's context.
```
mknorps, line 324:
```suggestion
```
Nit: Do you think this comment is needed? It duplicates exactly what the code does, which may cause some later code vs comments discrepancy if we change something.

fawltydeps/packages.py
```diff
@@ -317,14 +317,27 @@ class TemporaryPipInstallResolver(BasePackageResolver):
...
+
+        It is up to the caller to detect/handle any requirements that we failed
+        to install.
+        """
```
mknorps, line 328:
```suggestion
        It is up to the caller to handle any requirements that we failed
        to install.
        """
```
Nit: detection is now done via looping over all `requirements`.

mknorps (approved):
This one is good to :rock:, joking - :rocket:  ;)

I added two small nitpicks, that you may or may not consider.

fawltydeps/packages.py
jherland: I would agree if this was a _comment_ in the code, but when it's a _docstring_ I think it is somewhat different:

The docstring is often visible in places where the code is not (or less) visible (see below), and I believe the behavior described in these three lines is valuable, even for a casual user of this method (that would not want to read the code), as it describes postconditions that the caller must be aware of (i.e. do we enter the context on failure? Do we abort of the first failing package, or do we persist/retry?).

Even if the code and docstring later would disagree (which of course should be fixed as soon as it is discovered), I think it is valuable to capture the _intended behavior_ of this method in as simple English as possible. Of course, the above text is not particularly good in that regard, and I think it could be simplified further:

```suggestion
        We try to install as many of the given requirements as possible. Failed
        requirements will be logged with warning messages, but no matter how
        many failures we get, we will still enter the caller's context.
```

---

To my point of docstrings being _more visible_ than code: For example, in the Python REPL you can use `help()` to view the signature + docstring of something to learn how to call it:

```sh
$ python
Python 3.10.9 (main, Dec  6 2022, 18:44:57) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from fawltydeps.packages import TemporaryPipInstallResolver
>>> help(TemporaryPipInstallResolver.installed_requirements)
Help on function installed_requirements in module fawltydeps.packages:

installed_requirements(venv_dir: pathlib.Path, requirements: List[str]) -> Iterator[pathlib.Path]
    Install the given requirements into venv_dir with `pip install`.
    
    If `pip install` fails, we will log a warning, then retry by attempting
    to `pip install` each given requirement in order (logging any errors
    that occur), before we finally enter the caller's context.
    
    It is up to the caller to handle any requirements that we failed
    to install.
```

Another example is when a Python library is documented via an automatically generated website; the docstrings are included, but the actual code is not included (or, at best, referenced in a link).

jherland: Cleaned up the docstring in 41d9a42, it now reads:
```python
        """Install the given requirements into venv_dir with `pip install`.

        We try to install as many of the given requirements as possible. Failed
        requirements will be logged with warning messages, but no matter how
        many failures we get, we will still enter the caller's context. It is
        up to the caller to handle any requirements that we failed to install.
        """
```

Issue #226, opened 2023-03-14
Exclude directories from analysis


colindean:
**Is your feature request related to a problem? Please describe.**

When I tried to run fawltydeps for the first time, it caught some stuff in my project's `dist` and `build` directories. E.g., `dist` contains a requirements.txt generated from my Poetry config. 

**Describe the solution you'd like**

These directories could be ignored entirely, likely by default but also by configuration.

**Describe alternatives you've considered**

I could run `make clean` to delete these directories before running fawltydeps, but that'd cost extra time later on my workstation. It'd also necessitate running fawltydeps in CI before anything is generated.

jherland:
Thanks for reporting!

I'm wondering if it would be "safe" to always ignore `dist/` and `build/` or if there cases where that would backfire. Maybe safer to take our ignore list from some other source, like `.gitignore`. Would that work in your case?

For now I added this issue to our roadmap (#195).

colindean:
I think it would be safe to use the .gitignore in this case. Thanks for considering this!

jherland:
Note to selves: When we get around to implementing a better mechanism for ignoring files, please take https://github.com/tweag/FawltyDeps/pull/326#discussion_r1259238169 into consideration; in particular: consider if there are cases where the search for code/deps/pyenv sources _differ_ in what paths should be ignored.

For example: it might make sense for code and deps to ignore hidden (dot) dirs, while the pyenv search _should_ descend into these dirs?

md384:
I second this issues and would like to configuration argument that would accept globs (or just simple paths) to ignore to allow a little more flexibility.

dorranh:
+1 for this feature. I've encountered the same issue when running FawltyDeps in a project which had a quick, ad-hov virtual environment created in the project directory for experimentation purposes.

jherland:
With the [v0.15.0 release](https://github.com/tweag/FawltyDeps/releases/tag/v0.15.0), we now support `--exclude` to ignore files and directories. For now, `--exclude` defaults to the `.*` pattern, meaning that "hidden" files/directories are excluded by default.

We don't (yet) exclude `build/` and `dist/` by default, so you'd have to specify `--exclude build/ dist/` on the command-line (or `exclude = ["build/", "dist/"]` in the FawltyDeps section of your `pyproject.toml`)

[...]

PR #324, opened 2023-05-08
New code for traversing directories

jherland:
(Depends on #318)

Commits:
- `utils`: Add `DirectoryTraversal`, a more powerful alternative to `walk_dir()`
- `traverse_project`: Refactor `find_sources()` to `DirectoryTraversal`
- Move `walk_dir()` from `fawltydeps.utils` to `tests.utils`

### DIFF
--- a/fawltydeps/traverse_project.py
+++ b/fawltydeps/traverse_project.py
@@ -1,18 +1,18 @@
 """Traverse a project to identify appropriate inputs to FawltyDeps."""
 import logging
 from pathlib import Path
-from typing import AbstractSet, Dict, Iterator, Optional, Set, Type
+from typing import AbstractSet, Iterator, Optional, Type, Union
 
 from fawltydeps.extract_declared_dependencies import validate_deps_source
 from fawltydeps.extract_imports import validate_code_source
 from fawltydeps.settings import Settings
 from fawltydeps.types import CodeSource, DepsSource, Source, UnparseablePathException
-from fawltydeps.utils import walk_dir
+from fawltydeps.utils import DirectoryTraversal
 
 logger = logging.getLogger(__name__)
 
 
-def find_sources(
+def find_sources(  # pylint: disable=too-many-branches
     settings: Settings,
     source_types: AbstractSet[Type[Source]] = frozenset([CodeSource, DepsSource]),
 ) -> Iterator[Source]:
@@ -22,18 +22,23 @@ def find_sources(
     and yield the corresponding *Source objects found.
     """
 
-    # Collect any directories we will need to traverse.
-    dirs_to_traverse: Dict[Path, Set[Type[Source]]] = {}
+    logger.debug("find_sources() Looking for sources under:")
+    logger.debug(f"    code:   {settings.code}")
+    logger.debug(f"    deps:   {settings.deps}")
+
+    traversal: DirectoryTraversal[Union[Type[Source], Path]] = DirectoryTraversal()
 
     for path_or_special in settings.code if CodeSource in source_types else []:
         # exceptions raised by validate_code_source() are propagated here
         validated: Optional[Source] = validate_code_source(path_or_special)
         if validated is not None:  # parse-able file given directly
+            logger.debug(f"find_sources() Found {validated}")
             yield validated
         else:  # must traverse directory
             # sanity check: convince mypy that SpecialPath is already handled
             assert isinstance(path_or_special, Path)
-            dirs_to_traverse.setdefault(path_or_special, set()).add(CodeSource)
+            traversal.add(path_or_special, CodeSource)
+            traversal.add(path_or_special, path_or_special)  # also record base dir
 
     for path in settings.deps if DepsSource in source_types else []:
         # exceptions raised by validate_deps_source() are propagated here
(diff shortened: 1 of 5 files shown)


tests/test_dir_traversal.py
```diff
@@ -0,0 +1,278 @@
...
+                ExpectedTraverseStep("sub1", ["rel_link_sub2"], [], []),
+                ExpectedTraverseStep("sub1/rel_link_sub2", ["abs_link_sub1"], [], []),
+            ],
+            [
+                ExpectedTraverseStep(".", ["sub1", "sub2"], [], []),
+                ExpectedTraverseStep("sub2", ["abs_link_sub1"], [], []),
+                ExpectedTraverseStep("sub2/abs_link_sub1", ["rel_link_sub2"], [], []),
+            ],
+        ],
+    ),
```
Nour-Mws, line 192: There's no `add` here. How is the traversal's `to_traverse` set initialized?

fawltydeps/traverse_project.py
```diff
@@ -22,43 +22,58 @@ def find_sources(
...
         validated: Optional[Source] = validate_code_source(path_or_special)
         if validated is not None:  # parse-able file given directly
+            logger.debug(f"find_sources() Found {validated}")
             yield validated
         else:  # must traverse directory
             # sanity check: convince mypy that SpecialPath is already handled
             assert isinstance(path_or_special, Path)
-            dirs_to_traverse.setdefault(path_or_special, set()).add(CodeSource)
+            traversal.add(path_or_special, CodeSource)
+            traversal.add(path_or_special, path_or_special)  # also record base dir
```
Nour-Mws, line 41: I don't follow what this line is doing.

Nour-Mws (commented):
I added some non major comments and questions. I'll need a second quick look after that but this looks good :1st_place_medal:

fawltydeps/traverse_project.py
jherland: We "attach" two pieces of information to the traversal of this directory:
- `CodeSource`: i.e. we want to find/create `CodeSource` objects under this directory
- `path_or_special`: This directory is the _base directory_ we want to use when creating `CodeSource` objects from python files under this directory. This "attached" information is extracted (or "detached") on line 62 below, and is then passed as the `base_dir` argument to the `validate_code_source` call. From there it is stored in the `CodeSource.base_dir` member, and finally passed to `make_isort_config()` to make `isort` correctly discern first- and third-party imports.

The reason why we need to attach this here is that with the new `DirectoryTraversal` code the _base directory_ is no longer implicitly known _while_ we traverse.

Before we had:
```python
    for dir_path, types in dirs_to_traverse.items():
        assert len(types) > 0
        for file in walk_dir(dir_path):
            ...
```
and `dir_path` was assumed to be the base directory. If `dirs_to_traverse` contained both `foo/` and `foo/bar/`, we would end up traversing `foo/bar` _twice_, first with `foo/` as base dir, and then with `foo/bar/`.

The new code fixes that issue, but at the cost of the traversal itself no longer knowing exactly which of `foo/` or `foo/bar/` would be the appropriate base dir. We fix this by "attaching" the correct base dir to the traversal object.

fawltydeps/utils.py
```diff
@@ -23,26 +38,154 @@ def version() -> str:
...
+    def traverse(self) -> Iterator[Tuple[Path, Set[Path], Set[Path], List[T]]]:
+        """Perform the traversal of the added directories.
+
+        For each directory traverse, yield a 4-element tuple consisting of:
+        - The path to the current directory being traversed.
+        - An unordered iterator of all (immediate) subdirectories in the current
+          directory. This is NOT a mutable list, as you might expect from
+          os.walk(); instead, to prevent traversing into a subdirectory, you can
+          call .ignore() with the relevant subdirectory.
+        - An unordered iterator of all files in the current directory
```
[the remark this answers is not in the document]
jherland: Oops, I forgot to update the docstring after a late refactoring. These _could_ have been lazy iterators instead of sets, but then the caller would have to take extra care to make sure they were iterated at most _once_. I'd rather keep things simpler although arguably too much is computed in the case where the caller ends up not referencing these sets.

Will fix the docstring.

tests/test_dir_traversal.py
```diff
@@ -0,0 +1,278 @@
...
+            ExpectedTraverseStep("sub", [], [], [123, 456]),
+        ],
+    ),
+    DirectoryTraversalVector(
+        "one_subdir__attach_data_on_parent__traverse_only_subdir_with_no_data",
+        given=[Dir("sub")],
+        add=[(".", 123), ("sub", None)],
+        ignore=["."],
+        expect=[ExpectedTraverseStep("sub", [], [], [])],
+    ),
```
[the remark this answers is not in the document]
jherland: Yes, I will update the test ID to better explain what's happening here.

As we discussed in the meeting, there are situations where you ignore a directory, but then still want to traverse a subdirectory:
- As of #326  (documented [here](https://github.com/tweag/FawltyDeps/pull/326/files#diff-9caafe56cb1b84736f49984af7489e0ddb419bb9440f98d4c08bca02996871e2R33-R47)), when we find a Python environment, we want to ignore any code or deps inside that Python environment (as they most likely contain code/deps from _dependencies_ and are not relevant to the main project). However, given a sufficiently specific CLI argument, we should _still_ look for sources that the user wants us to find. For example:
   `fawltydeps --pyenv=. --deps=.venv/lib/python3.10/site-packages/foo/`
   The `--pyenv=.` will end up finding a  Python environment at `.venv` which would then cause us to ignore any code or deps sources under `.venv/`. However, the `--deps=.venv/lib/python3.10/site-packages/foo/` is even more specific, so should still find something like `.venv/lib/python3.10/site-packages/foo/requirements.txt`.
- When we get around to fixing #226 by adding an `--ignore` option, you could easily construct similar examples like:
   `fawltydeps --code=. --ignore=tests/sample_projects --code=tests/sample_projects/blog_post_example/`

tests/test_dir_traversal.py
Nour-Mws: No that's fine as it is. I missed that part.

fawltydeps/traverse_project.py
```diff
@@ -22,43 +22,58 @@ def find_sources(
...
+                traversal.ignore(subdir)
 
-    for dir_path, types in dirs_to_traverse.items():
+        types = {t for t in extras if t in source_types}
         assert len(types) > 0
-        for file in walk_dir(dir_path):  # traverse directories only _once_
-            if CodeSource in types:
+        if CodeSource in types:
+            # Retrieve base_dir from closest ancestor, i.e. last Path in extras
+            base_dir = next((x for x in reversed(extras) if isinstance(x, Path)), None)
```
Nour-Mws, line 63: Double checking my understanding here:
If the user passed 2 code directories, where one is the child of the other, e.g. `bar/` and `bar/foo/`, this line's goal is to select one of those as the `base_dir`. And the way these extras are added as the directory is traversed ensures that the child directory will appear later in the list of extras than the parent directory?

fawltydeps/traverse_project.py
Nour-Mws: That makes much more sense now! Thanks for the explanation!

Nour-Mws (approved):
I had a second pass and this looks good. Feel free to resolve and merge as soon as the dependency is resolved :rocket:

fawltydeps/traverse_project.py
jherland: Correct. The `extras` list of attached data comes from the `accumulate_attached_data()` helper inside `DirectoryTraversal.traverse()`, and this one guarantees the order of the attached data (parent dirs before child dirs). The line above reverses this, so that - indeed - we do get the "closest" `base_dir`.

fawltydeps/utils.py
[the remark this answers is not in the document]
jherland: Fixed docstring in c3fd4d6.

PR #325, opened 2023-05-08
Prepare test suite for traversing Python environments

jherland:
(Depends on #324)

There are more preparatory steps on the way to integrating Python environments logic into the `traverse_project` module.

Commits:
- `test_sample_projects`: Accept new 'pyenvs' directive in `expected.toml`
- `sample_projects`: Add `pyenv_galore` 'project'
- `test_traverse_project`: Fix incorrect `.project` member for a test vector
- `test_traverse_project`: Prepare for testing with 'pyenvs'

Issue #249, opened 2023-03-17
Using _multiple_ Python environments to resolve deps -> packages -> imports?


jherland:
The basic question here is: How can we maximize FawltyDeps' chances of finding a relevant `Package` that provides a useful dep name -> import names mapping, and minimize the cases where we fall back to the identity mapping?

Currently we take one `--pyenv` option (or default to packages found via `sys.path`, i.e. the environment where FD is installed).

Ideas:
1. A natural extension is to accept multiple `--pyenv` values (like we recently did for `--code` and `deps`), to allow multiple environments to be queried. [^1]
2. Asking the user to pass in multiple `--pyenv`s is not very friendly. We could instead walk the directory structure of the project (limited by `basepaths` and/or `--pyenv`) and automatically discover all Python environments within, and add them to our analysis. [^2]
3. A comment on #236 suggests also looking at `$PYTHONPATH`. Some research is needed to figure how this interacts with `sys.path`.
4. There are other places where Python packages are often installed, `pip install --user` places packages in a user-specific directory. That might be a good place to discover packages that are undeclared because they were simply part of a user's "ambient" environment.
5. Same goes for system-wide installed packages, whether they are installed by `pip`, og by the distro.
6. Elsewhere? What about tools like `pipx`?

Where do we stop? How much is too much?

Rather, what are the _disadvantages_ of adding more places to look for packages?
- What are the chances of encountering an environment that is willfully misleading?
- What are the _consequences_ of using a misleading environment?

[^1]: We can fairly straightforwardly extend our `resolve_dependencies()` to have a stack of multiple `LocalPackageLookup` objects, one for each environment, and then query them in order until one of them returns a matching `Package` object. The identity mapping can then be represented as a `IdentityPackageLookup` object automatically placed at the bottom of this stack. The way `LocalPackageLookup` is currently written, we delay enumerating packages (i.e. walking the filesystem) until we actually need to look something up. Therefore, as long we put the most relevant virtualenvs near the top of the stack, we can probably avoid enumerating packages for the lower levels of the stack, and we can then have a fairly deep stack without necessarily paying the cost (in terms of enumerating less relevant virtualenvs).

[^2]: If we auto-discover virtualenvs while walking the project directories, we should consider _removing_ these from the `code` and `deps`, as they should probably not be considered part of the project itself, and thus shouldn't be scanned for imports/deps by FD.

janydoe:
[...]

PR #326, opened 2023-05-08
Teach FawltyDeps to automatically discover Python environments inside the project

jherland:
(Depends on #325)

This is at the heart of solving #249, and IMHO one of the major pieces remaining in the Mapping Strategy milestone.

With this, FawltyDeps is able to:
- automatically find Python environments inside the project directory (`basepath`)
- automatically _disregard_ files inside the Python environment for `.code` and `.deps`

Commits:
- `traverse_project`: Add support for finding/validating Python environments
- `main`: Use `PyEnvSource` objects when resolving dependencies
- Update documentation related to Python environments

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,22 +138,18 @@ When FawltyDeps looks for undeclared and unused dependencies, it needs to match
 `import` statements in your code with corresponding package dependencies
 declared in your project configuration.
 
-To solve this, FawltyDeps adopts several strategies: mapping provided by the user,
-identity mapping, and most powerful of all - using your python environment.
+To solve this, FawltyDeps uses a sequence of resolvers (aka. mapping strategies)
+to determine which Python packages that provide which import names:
 
 #### Python environment mapping
 
-FawltyDeps looks at the packages installed in your _current
-Python environment_ and what import names each of them provide in order to
-correctly match your dependencies against your imports.
+First of all FawltyDeps looks for _Python environments_ (virtualenvs or similar
+directories like `.venv` or `__pypackages__`.) inside your project (i.e. under
+`basepath`, if given, or the current directory).
 
-The _current Python environment_ in this case is the environment in which
-FawltyDeps itself is installed. This works well when you, for example,
-`pip install fawltydeps` into the same virtualenv as your project dependencies.
-
-If you instead want FawltyDeps to look into a _different_ Python environment for
-mapping dependencies to import names, you can use the `--pyenv` option,
-for example:
+You can use the `--pyenv` option (or the `pyenvs` configuration directive)
+to point FawltyDeps at a specific Python environment located within your project
+or elsewhere. For example:
 
 ```sh
 fawltydeps --code my_package/ --deps pyproject.toml --pyenv .venv/
@@ -166,17 +162,23 @@ This will tell FawltyDeps:
 - to use the Python environment at `.venv/` to map dependency names in
   `pyproject.toml` into import names used in your code under `my_package/`
 
+If `--pyenv` is not used, and no Python environments are found within your project,
+FawltyDeps will fall back to looking at your _current Python environment_:
+This is the environment in which FawltyDeps itself is installed.
+This works well when you, for example, `pip install fawltydeps` into the same
+virtualenv as your project dependencies.
+
 You can use `--pyenv` multiple times to have FawltyDeps look for packages in
-multiple Python environments. When mapping a dependency into import names,
-FawltyDeps will then use the union of all imports provided by all matching
-packages across those Python environments as valid import names for that
-dependency.
+multiple Python environments. In this case (or when multiple Python environments
+are found inside your project) FawltyDeps will use the union (superset) of all
+imports provided by all matching packages across those Python environments as
+valid import names for that dependency.
 
 #### Identity mapping
 
-When FawltyDeps is unable to find an installed package that corresponds to a
-declared dependency, FawltyDeps will fall back to an "identity mapping", where
-it _assumes_ that the dependency provides a single import of the same name,
+When unable to find an installed package that corresponds to a declared
+dependency, FawltyDeps will fall back to an "identity mapping", where it
+_assumes_ that the dependency provides a single import of the same name,
 i.e. it will expect that when you depend on `some_package`, then that should
 correspond to `import some_package` statements in your code.
 
@@ -188,8 +190,8 @@ project dependencies](#why-must-fawltydeps-run-in-the-same-python-environment-as
 
 #### User-defined mapping
 
-You may define your mapping by providing a toml file with package to imports mapping,
-`my_mapping.toml`:
+As a final solution and/or an ultimate override, you may define your own mapping
+of dependency names to import names, by providing a TOML file like this:
 
 ```toml
 my-package = ["mpkg"]
@@ -203,9 +205,11 @@ To use your mapping, run:
 fawltydeps --custom-mapping-file my_mapping.toml
 ```
 
-FawltyDeps will parse `my_mapping.toml` file and use extracted mapping for matching dependencies to imports.
+FawltyDeps will parse your `my_mapping.toml` file and use the extracted mapping
+for matching dependencies to imports.
 
-You may also place the custom mapping in the `pyproject.toml` file of your project, inside a `[tool.fawltydeps.custom_mapping]` section, like this:
+You may also place the custom mapping in the `pyproject.toml` file of your
+project, inside a `[tool.fawltydeps.custom_mapping]` section, like this:
 
 ```toml
 [tool.fawltydeps.custom_mapping]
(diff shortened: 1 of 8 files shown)


fawltydeps/traverse_project.py
```diff
@@ -1,30 +1,56 @@
...
+    source_types: AbstractSet[Type[Source]] = frozenset(
+        [CodeSource, DepsSource, PyEnvSource]
+    ),
 ) -> Iterator[Source]:
     """Traverse files and directories and yield Sources to be parsed.
 
     Traverse the files and directories configured by the given Settings object,
     and yield the corresponding *Source objects found.
+
+    Some rules/principles:
```
mknorps, line 33: <3

fawltydeps/traverse_project.py
```diff
@@ -51,13 +77,29 @@ def find_sources(  # pylint: disable=too-many-branches
         else:  # must traverse directory
             traversal.add(path, DepsSource)
 
+    for path in settings.pyenvs if PyEnvSource in source_types else []:
```
mknorps, line 80: Nit:
Following rule of 3, handling of source in from the `settings` object could be abstracted out. 
On the other hand we do not anticipate more source types. I will leave it to your discretion :)

mknorps: On yet another hand (;)) it may be easier to test it when it is a separate function.

fawltydeps/traverse_project.py
```diff
@@ -51,13 +77,29 @@ def find_sources(  # pylint: disable=too-many-branches
...
+            traversal.add(path, PyEnvSource)
+
     for _cur_dir, subdirs, files, extras in traversal.traverse():
         for subdir in subdirs:  # don't recurse into dot dirs
             if subdir.name.startswith("."):
                 traversal.ignore(subdir)
 
         types = {t for t in extras if t in source_types}
         assert len(types) > 0
+        if PyEnvSource in types:
```
mknorps, line 97: In one `for` above we ignore "dot" subdir directories.  We would not like to ignore them for virtual environments. Althought, the probability that someone put virtual env in a subdir is rather low.

mknorps (commented):
As usual Johan, your PRs are elegant, well written and are a great contribution to the project <3

I am posting a comment, not approval for now, as I have to go through tests and documentation. The comments are mostly nits or me making sure that I follow your line of thoughts :)

fawltydeps/cli_parser.py
```diff
@@ -163,8 +163,9 @@ def populate_parser_paths_options(parser: argparse._ActionsContainer) -> None:
         metavar="PYENV_DIR",
         help=(
             "Where to find Python environments that have project dependencies"
-            " installed. When empty (the default), fall back to the Python"
-            " environment where FawltyDeps is installed."
+            " installed. Defaults to looking for Python environments under the"
+            " current directory, or failing that, using the Python environment"
```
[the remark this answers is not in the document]
jherland: Basepath. To be technically pedantic about it:

"Where to find Python environments that have project dependencies installed. Defaults to looking for Python environments under the given basepath(s), which again defaults to the current directory. If no Python environments are found under basepath(s), the Python environment where FawltyDeps is installed will be used."

We need some refactoring of the help messages here, because this is too long/complicated IMHO, and the `--code` and `--deps` options have the same problem...

fawltydeps/types.py
```diff
@@ -83,7 +83,33 @@ def __post_init__(self) -> None:
...
+    system-wide Python installation (under /usr, /usr/local, or similar),
+    in a virtualenv, in a poetry2nix environment, or a similar mechanism that
+    uses this layout. Alternatively, it can be a __pypackages__/X.Y/lib
+    directory within a project that uses PEP582.
+    """
+
+    path: Path
+
+    def __post_init__(self) -> None:
+        assert self.path.is_dir()  # sanity check
```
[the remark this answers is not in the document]
jherland: Do you mean that we should rename the class members themselves (`.path` -> `.dir` or `.file`)?

I'm usually fond of `.path` since it also helps to remind me that this is a `Path` object. There is technically no problem with using `.dir` and `.file` as member names, but since they are also built-in functions in Python (`dir()` and `file()`), I'm usually a little hesitant to use them...

Otherwise I agree that we should be as clear as possible about this in the documentation.

fawltydeps/cli_parser.py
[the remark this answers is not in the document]
mknorps: Maybe like that:
> Path to search for Python environments that have project dependencies installed. Default search order:\n
> - basepath(s) (by default the current directory)
> - Python environment where FawltyDeps is installed.

fawltydeps/packages.py
```diff
@@ -491,3 +492,19 @@ def resolve_dependencies(
         raise UnresolvedDependenciesError(names=unresolved)
 
     return ret
+
+
+def validate_pyenv_source(path: Path) -> Optional[Set[PyEnvSource]]:
```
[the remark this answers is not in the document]
mknorps: OK. We can park this for later :)

fawltydeps/traverse_project.py
mknorps: And what if the logic was a bit different.

Let's divide folders to two categories:
- regular
- hidden (ignored with traversal)
We have 3 types of entities we search for:
- `PyEnvSource`
- `CodeSource`
- `DepsSource`

We have a following martix to search:
|   |  `PyEnvSource` |  `CodeSource` |  `DepsSource` |
|---|---|---|---|
| regular |  x | x | x |
| hidden  |  x |   |   |

We could first check regular folders and then dot folders only for `PyEnvSource`. For that `traversal.ignore` should be modified to keep the list of directories ignored due to the fact that they started with a dot.

Additionally, we also know that some of the hidden folders will not contain virtual envs, and 
```
            if subdir.name.startswith("."):
```
could become
```
IGNORED_DIRS = [".git", ".node",  ]
            if any(subdir.name.startswith(ignored) for ignored in IGNORED_DIRS):
```
and we would never need to recurse to those directories. Still we could use mechanism of revisiting ignored dirs for `PyEnvSource`.

mknorps: To be clear, this is not a blocker. I do not think there is a lot of such corner cases. We could move this Idea to an issue and solve it in a separate PR. The current logic is good enough :)

README.md
```diff
@@ -214,9 +218,10 @@ scikit-learn = ["sklearn"]
 multiple-modules = ["module1", "module2"]
 ```
 
-Caution when using your mapping is advised. The user-defined mapping takes precedence over all other resolving strategies.
-If the mapping file has some stale mapping entries, they will not be resolved by
-Python environment resolver (which in general is the most accurate).
+Caution when using your mapping is advised: The user-defined mapping takes
+precedence over the other resolvers documented above. For example, if the
+mapping file has some stale/incorrect mapping entries, they will _not_ be
+resolved by the Python environment resolver (which is geenrally more accurate).
```
mknorps, line 224:
```suggestion
resolved by the Python environment resolver (which is usually more accurate).
```

README.md
```diff
@@ -276,10 +281,11 @@ Here is a complete list of configuration directives we support:
   Defaults to the current directory, i.e. like `code = ["."]`.
 - `deps`: Files or directories containing the declared dependencies.
   Defaults to the current directory, i.e. like `deps = ["."]`.
-- `pyenvs`: Python environments (directories like `.venv`, `__pypackages__`, or
-  similar) to use for resolving project dependencies into provided import names.
-  Defaults to an empty list, i.e. like `pyenvs = []`, which is interpreted as
-  using the Python environment where FawltyDeps is installed (aka. `sys.path`).
+- `pyenvs`: Where to look for Python environments (directories like `.venv`,
+  `__pypackages__`, or similar) to be uses for resolving project dependencies
```
mknorps, line 285:
```suggestion
  `__pypackages__`, or similar) to be used for resolving project dependencies
```

mknorps (approved):
I think this is good to go 🚀 🎸

fawltydeps/traverse_project.py
jherland: Agreed. I posted https://github.com/tweag/FawltyDeps/issues/226#issuecomment-1635812164 for now, to not forget about this when we get around to refactoring our ignore logic.

fawltydeps/types.py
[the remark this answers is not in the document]
jherland: Improved the documentation of these members in 737e14ffa0531fbd404246c96de49b8c2fa99e82

fawltydeps/main.py
```diff
@@ -73,8 +74,8 @@ def sources(self) -> Set[Source]:
         source_types: Dict[Action, Set[Type[Source]]] = {
```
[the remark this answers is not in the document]
jherland: Added some more details to the `Analysis` docstring in 92f3f0de7e7ca040ac5d0b166287a2f4079d7f77. This is the kind of text that can be very helpful for a newcomer (I hope), but also can quickly go out of date when future changes are introduced. Let's see how this goes.

Issue #334, opened 2023-06-28
isort 5.12 not supported


hogru:
**Describe the bug**
Well, somewhere in between a bug and a feature request. `FawltyDeps` has a dependency of `isort = ">=5.10,<5.12.0"`, but the current version of `isort` is v5.12. Hence, `FawltyDeps` can not be installed in environments with a `isort = "^5.12.0"` dependency.

**To Reproduce**
`poetry add fawlty` in a project with isort = "^5.12.0" in its `pyproject.toml` file.

**Expected behavior**
Support `isort` v5.12 and change the corresponding dependency in `pyproject.toml`.

**Environment**
 - OS name + version: macOS 13.4.1
 - Version of the code: `FawltyDeps` v0.11, `poetry` v1.5.1

**Additional context**
None

jherland:
This limit was put in place in 828a5eaa345d80df5174a3464cc58fea5187923f, due to isort v5.12 dropping support for Python v3.7, which we (at that point, at least) wanted to keep supporting.

I'll take a closer look to see if (a) we can find some way around this where the isort `<5.12.0` constraint is only applied when on Python v3.7, or (b) reconsider whether we drop Python v3.7 support ourselves, now that [it was EOLed a couple of days ago](https://devguide.python.org/versions/).

hogru:
Thank you @jherland for the swift response. Understood, and looking forward to checking my dependencies once this can be implemented. Thank you for the nice idea/package.

jherland:
Thank you for reporting! I just posted PR #335 to fix this. Might take a while to get reviewed/merged/released, as our team members are mostly busy with other projects these days.

hogru:
That was fast! 😃  Thank you, @jherland and @mknorps. Will check once released.

PR #335, opened 2023-06-30
Allow isort >= 5.12 when running on Python > 3.7

jherland:
Fixes issue #334.


mknorps (approved):
Checked for both cases (3.7 and 3.8+). Works well, thank you @jherland 🎉

PR #336, opened 2023-07-03
Upgrade our dev env to use Python v3.11 by default

jherland:
Python v3.11 has been stable for a while, and there's no good reason for
us to stick with v3.10 in our default development environment.

This series modernizes our `shell.nix` and helps prepare for introducing
Python v3.12 support at a later date.

Also switch the `format` action in GitHub CI from Python v3.10 to v3.11.

Commits:
- `shell.nix`: Use `fetchTarball` instead of `fetchGit`
- `shell.nix`: Update our development environment
- `pyproject.toml`: Update `codespell` to support being run with Python v3.11
- `shell.nix`: Switch to Python v3.11 as the default in our dev shell
- `.github/workflows/format.yaml`: Switch format check in CI: 3.10 -> 3.11


Nour-Mws (approved):
The changes makes sense, as does the Nix syntax for a casual reader (me).

Issue #266, opened 2023-03-23
Add mapping options to CLI (e.g. `--install-deps`)


Nour-Mws:
This option is already usable via the toml config or env variables as per #252 but is still not available through the CLI.
This issue is to be tackled after some more progress and visibility have been obtained on #256.

Collecting here pieces of feedback from @jherland  when I attempted to add the option in #252.

> I don't think the `default=False` is needed when `action="store_true"` is used. Furthermore, I think it actually defeats the configuration cascade (which really depends on all relevant CLI options having `default=argparse.SUPPRESS`). What we want is that this behavior:
> - `--install-deps` on the command-line causes `args.install_deps == True` to be returned from argparse
> - No `--install-deps` on the command-line causes `args.install_deps` to NOT EXIST, allowing the `install_deps` option from the config file to take effect.
> 
> Additionally, I think the help message should emphasize that the deps are installed in a temporary/separate venv:
> 
> ```suggestion
>     parser.add_argument(
>         "--install-deps",
>         action="store_true",
>         help=(
>             "Allow FawltyDeps to `pip install` declared dependencies into a"
>             " separate temporary virtualenv to discover the imports they expose"
>         ),
>     )
> ```
> 

_Originally posted by @jherland in https://github.com/tweag/FawltyDeps/pull/252#discussion_r1145518875_

> This change highlights the problem that I talked about above: Setting `install_deps = True` in the config file, and then running without `--install-deps` on the command line _should_ let the config file decide, but this illustrates that it doesn't: the `default=False` will always override whatever comes from the config file...
> 
> However, I realize there is an inherent problem with boolean command-line flags here: If you can only _enable_ some behavior with a command-line option, then - if the behavior has already been enabled in the config file -  how could you then _disable_ this behavior at the command-line? I think the answer here is to provide a corresponding (mutually exclusive) `--no-install-deps` option that sets `install_deps` to `False`.
> 
> In other words we should have:
> - `--install-deps`: Set `install_deps` to `True`
> - `--no-install-deps`: Set `install_deps` to `False`
> - (nothing): Don't change `install_deps` at all. Use the value from the config file, falling back to the ultimate hardcoded default (`False`)
> 

_Originally posted by @jherland in https://github.com/tweag/FawltyDeps/pull/252#discussion_r1145522180_

PR #341, opened 2023-07-25
ADD CLI option for --install-deps

mknorps:
In this PR, `--install-deps` option is exposet to the user. It is almost the last piece of puzzle to add installing dependencies in a temporal virtual environment as a dependenceny mapping resolver.

We are left with describing it in the documentation #282 .

I tested the `--install-deps` option in CLi and if the virtual environment contains a dependency it is not installed in the temporal environment 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,6 +299,8 @@ Here is a complete list of configuration directives we support:
   declared dependencies. Must be one of `"requirements.txt"`, `"setup.py"`,
   `"setup.cfg"`, `"pyproject.toml"`, or leave it unset (i.e. the default) for
   auto-detection (based on filename).
+- `install-deps`: Install Python dependencies gathered with FawltyDeps to
+  a temporary virtual environment. Dependencies will be installed from PyPI.
 - `verbosity`: An integer controlling the default log level of FawltyDeps:
   - `-2`: Only `CRITICAL`-level log messages are shown.
   - `-1`: `ERROR`-level log messages and above are shown.
--- a/README.md
+++ b/README.md
@@ -299,8 +299,9 @@ Here is a complete list of configuration directives we support:
   declared dependencies. Must be one of `"requirements.txt"`, `"setup.py"`,
   `"setup.cfg"`, `"pyproject.toml"`, or leave it unset (i.e. the default) for
   auto-detection (based on filename).
-- `install-deps`: Install Python dependencies gathered with FawltyDeps to
-  a temporary virtual environment. Dependencies will be installed from PyPI.
+- `install-deps`: Automatically install Python dependencies gathered with
+  FawltyDeps into a temporary virtual environment. This will use `pip install`,
+  which downloads packages from PyPI by default.
 - `verbosity`: An integer controlling the default log level of FawltyDeps:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -299,6 +299,8 @@ Here is a complete list of configuration directives we support:
   declared dependencies. Must be one of `"requirements.txt"`, `"setup.py"`,
   `"setup.cfg"`, `"pyproject.toml"`, or leave it unset (i.e. the default) for
   auto-detection (based on filename).
+- `install-deps`: Install Python dependencies gathered with FawltyDeps to
+  a temporary virtual environment. Dependencies will be installed from PyPI.
```
jherland, line 303: I'd like this part to be a little more precise (and verbose) about how we install and where packages come from. Specifically, we only install packages from PyPI because that is where `pip install` goes by default. Therefore, I think we're better off mentioning `pip install` specifically here (so that users can infer that `pip`'s configuration/environment variables are what they should look at if default `pip` behavior is not suitable):

```suggestion
- `install-deps`: Automatically install Python dependencies gathered with
  FawltyDeps into a temporary virtual environment. This will use `pip install`,
  which downloads packages from PyPI by default.
```

fawltydeps/cli_parser.py
```diff
@@ -171,6 +171,15 @@ def populate_parser_paths_options(parser: argparse._ActionsContainer) -> None:
...
         ),
     )
+    parser.add_argument(
+        "--install-deps",
+        dest="install_deps",
+        action="store_true",
+        help=(
+            "Allow FawltyDeps to `pip install` declared dependencies into a"
+            " separate temporary virtualenv to discover the imports they expose."
+        ),
```
jherland, line 181: I really like this help text. It's short and explains exactly what happens in a way that helps point the user in the right direction if they need more help. E.g. a user that that needs to customize how/where packages are downloaded now knows that `pip install` is what they need to customize. 👏🏻

tests/test_cmdline_options.py
```diff
@@ -60,7 +60,7 @@
 """
 )
 venvs = [str(project_with_no_issues / ".venv")]
-other = ["--generate-toml-file", "--version"]
+other = ["--generate-toml-config", "--version", "--install-deps"]
```
jherland, line 63: I'm struggling to see what (if anything) this is used for? Can we remove the entire line?

jherland (approved):
Some minor nitpicking on the README, and one inconsequential thing I did not understand in `test_cmdline_options`.

But this is good to go! :rocket:

tests/test_cmdline_options.py
mknorps: This is not used now anywhere, it is a reminder for our future selves to add `other` to the tested CLI options.

fawltydeps/cli_parser.py
mknorps: I agree :D I copied that from https://github.com/tweag/FawltyDeps/issues/266 😆

jherland: 😳 😆

PR #342, opened 2023-07-25
Documentation polishes:  Update the order of options in the Configuration section of the documentation.

mknorps:
Update the order of options in the Configuration section of the documentation. to be the same as the message from the CLI.

Move 'custom_mapping_file' Settings entry to the bottom, to be closer to the custom mapping section in the generated TOML.


jherland (approved):
Looks great! 🚀

PR #344, opened 2023-07-28
Fix Settings.pyenvs to _actually_ use basepath by default

jherland:
This is a rather embarrasing omission from PR #326 (Teach FawltyDeps to automatically discover Python environments inside the project):

Although that PR did include code to traverse within a given `--pyenv` path to find Python environments, the PR "forgot" to actually switch the _default_ behavior of `--pyenv`:

The default value of `settings.pyenvs` remained an empty set, and there were no changes to have `basepath` influence the value of `settings.pyenvs`.

This PR fixes that:

- When neither `--pyenv` nor `basepath` is given, `settings.pyenvs` should default to the current directory.
- When `--pyenv` is not given, but `basepath` is given, `basepath` should override the default `settings.pyenvs`.
- When `--pyenv` is given, it overrides `basepath`.

In short `settings.pyenvs` should behave exactly like `.code` and `.deps`.

---

Commits:
- `test_settings`: Prepare for including `--pyenv` in `basepath`-related tests
- Fix `Settings.pyenvs` to _actually_ use `basepath` by default

### DIFF
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -5,9 +5,21 @@ import random
 import string
 import sys
 from dataclasses import dataclass, field
-from itertools import chain, permutations, product
+from itertools import chain, combinations, permutations, product
 from pathlib import Path
-from typing import Any, Dict, Iterable, List, Optional, Set, Tuple, Type, TypeVar, Union
+from typing import (
+    Any,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+)
 
 import pytest
 from hypothesis import given, strategies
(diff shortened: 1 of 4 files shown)


tests/test_settings.py
```diff
@@ -82,24 +94,55 @@ def test_code_deps_and_base_unequal__raises_error(code_deps_base):
...
+    "--deps": "deps",
+}
+
+Item = TypeVar("Item")
+
+
+def subsets(
+    items: Set[Item],
+    min_size: int = 0,
+    max_size: int = sys.maxsize,
```
Nour-Mws, line 109: Is this set to `sys.maxsize` and then readapted in the first line of the function because:
- It's an optional argument so it needs a default value
- We can not set the default value to `max(items)` in the function call?

tests/test_settings.py
```diff
@@ -81,22 +81,23 @@ def _inner(**kwargs: str):
 
 safe_string = strategies.text(alphabet=string.ascii_letters + string.digits, min_size=1)
 nonempty_string_set = strategies.sets(safe_string, min_size=1)
-three_different_string_groups = strategies.tuples(
-    nonempty_string_set, nonempty_string_set, nonempty_string_set
-).filter(lambda ss: ss[0] != ss[1] and ss[0] != ss[2] and ss[1] != ss[2])
+four_different_string_groups = strategies.tuples(
+    *([nonempty_string_set] * 4),
```
Nour-Mws, line 85: Are we sure this is not reusing the same [variable] pointer and duplicating it 4 times, thus resulting in an empty set after the filtering is executed?

Nour-Mws (approved):
Looks good. As usual, it was great to read this well organized and logically evolving PR.
Note that I only checked that the changes make sense and did not verify that no other tests/code should be changed for this to be complete.

tests/test_settings.py
jherland: 💯 correct. In Python the default arguments (`= 0`, and `= sys.maxsize` above) are evaluated only _once_ when the function is first parsed/created by the Python interpreter. So the default argument has no way to reflect/update based on the value of an argument in a later call.

The other, maybe more common, way to write this would be something like:

```python
def subsets(
    items: Set[Item],
    min_size: int = 0,
    max_size: int = None,
) -> Iterator[Set[Item]]:
    max_size = len(items) if max_size is None else min(max_size, len(items))
```

(Aside: This is also why you should typically never use _mutable_ default arguments in Python, because the lifetime of the default-valued argument is non-obvious: the _same_ object is reused across all calls, and changes made in earlier function calls will be reflected in later function calls.)

PR #345, opened 2023-07-31
Bump version to 0.12.1

jherland:
Also, add a FAQ to `README.md` about using FD as a pre-commit hook.


Nour-Mws (approved):
Good to go. The CI is failing but I assume this is transient.

PR #346, opened 2023-07-31
Add `--list-sources` to CLI

jherland:
Make the traversal logic in FawltyDeps more visible/debuggable to the end user:
- Add the `*Source` objects to the JSON output
- Add the `--list-sources` CLI action with a human-readable version of the same

---

Commits:
- `main`: Add `Analysis.sources` to JSON output
- Add `--list-sources` CLI argument
- `main`: Reformat `print_human_readable()` to fix pylint's 'Too many branches'

### DIFF
--- a/fawltydeps/main.py
+++ b/fawltydeps/main.py
@@ -82,7 +82,7 @@ class Analysis:  # pylint: disable=too-many-instance-attributes
 
         # The following members are calculated once, on-demand, by the
         # @property @calculated_once methods below:
-        self._source: Optional[Set[Source]] = None
+        self._sources: Optional[Set[Source]] = None
         self._imports: Optional[List[ParsedImport]] = None
         self._declared_deps: Optional[List[DeclaredDependency]] = None
         self._resolved_deps: Optional[Dict[str, Package]] = None
@@ -199,6 +199,7 @@ class Analysis:  # pylint: disable=too-many-instance-attributes
             frozenset: partial(sorted, key=str),
             set: partial(sorted, key=str),
             type(BasePackageResolver): lambda klass: klass.__name__,
+            type(Source): lambda klass: klass.__name__,
         }
         encoder = partial(custom_pydantic_encoder, custom_type_encoders)
         json_dict = {
@@ -206,6 +207,7 @@ class Analysis:  # pylint: disable=too-many-instance-attributes
             # Using properties with an underscore do not trigger computations.
             # They are populated only if the computations were already required
             # by settings.actions.
+            "sources": self._sources,
             "imports": self._imports,
             "declared_deps": self._declared_deps,
             "resolved_deps": self._resolved_deps,
--- a/fawltydeps/main.py
+++ b/fawltydeps/main.py
@@ -15,7 +15,7 @@ import logging
 import sys
 from functools import partial
 from operator import attrgetter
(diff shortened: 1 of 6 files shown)


fawltydeps/main.py
```diff
@@ -220,56 +220,67 @@ def print_json(self, out: TextIO) -> None:
...
+        if self.is_enabled(Action.LIST_SOURCES):
+            output(render_sources())
+        if self.is_enabled(Action.LIST_IMPORTS):
+            output(render_imports())
+        if self.is_enabled(Action.LIST_DEPS):
+            output(render_declared_deps())
+        if self.is_enabled(Action.REPORT_UNDECLARED) and self.undeclared_deps:
+            output(render_undeclared())
+        if self.is_enabled(Action.REPORT_UNUSED) and self.unused_deps:
+            output(render_unused())
```
Nour-Mws, line 283: This might be more readable but I don't see why Pylint is happier with it. The number of `if`-branches stays the same :thinking:

fawltydeps/types.py
```diff
@@ -47,7 +48,25 @@ def __str__(self) -> str:
...
+
+    source_type: Type["Source"] = field(init=False)
+
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "source_type", self.__class__)
+
+    @abstractmethod
+    def render(self, detailed: bool) -> str:
+        """Return a human-readable string representation of this source."""
+        raise NotImplementedError
```
Nour-Mws, line 65: Why is this an `abstractmethod`?

tests/test_cmdline.py
```diff
@@ -557,7 +704,11 @@ def test_check_json__simple_project__can_report_both_undeclared_and_unused(
         "version": version(),
     }
     output, returncode = run_fawltydeps_function(
-        "--check", "--json", f"--code={tmp_path}", f"--deps={tmp_path}"
+        "--check",
+        "--json",
+        f"--code={tmp_path}",
+        f"--deps={tmp_path}",
+        f"--pyenv={tmp_path}",
```
Nour-Mws, line 711: Why add `--pyenv` here?

Nour-Mws (approved):
Looks great! Thanks for the hard work.

fawltydeps/main.py
jherland: Yes, but they're spread over different (nested) helper functions, which I suspect appeases Pylint in the end. Specifically, the `if detailed: ... else: ...` blocks are now neatly organized into the helper functions, and I have to say I actually find the current organization quite readable (once you get over the 6 nested functions inside this method 😅 )

fawltydeps/types.py
jherland: To force/require subclasses to implement it. In practice, I believe the `@abstractmethod` provides for a better error message if we somehow end up calling it at runtime.

tests/test_cmdline.py
jherland: I want to actually verify that what a `PyEnvSource` looks like in the expected output (lines 667-670).

Without adding `--pyenv`, I would have to rely on this Python environment to be auto-discovered, which only happens if `tmp_path` is passed as `basepath`, or if we had instrumented `run_fawltydeps_function()` to run with `tmp_path` as the current directory.

PR #348, opened 2023-08-07
Run fawltydeps against itself in CI

### LINKED ISSUE #123: Dog-food FawltyDeps in the CI
n/a

Run FawtlyDeps against itself in the GitHub Actions CI.

n/a

**Additional context**
n/a

zz1874:
Closed #123 

This PR adds a functionality of running fawltydeps against itself in the GitHub Actions CI.
While running fawltydeps, it reads in the python environment from `~/.cache/pypoetry/virtualenvs` and ignores format related dependencies (`black`, `codespell`, `colorama`, `hypothesis`, `mypy`, `nox`, `pylint`, `pytest`, and `types-setuptools`).

Also, this PR declares `setuptools` as a dependency since `pkg_resources` has been imported to the main code.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,4 +1,5 @@
 import hashlib
+import sys
 from pathlib import Path
 from typing import Iterable
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -84,7 +84,8 @@ def integration_tests(session):
 @nox.session(python=python_versions)
 def self_test(session):
     install_groups(session)
-    venv_path = session.run("poetry", "env", "info", "--path", silent=True).strip()
+    # For --pyenv, use Nox's virtualenv, or fall back to current virtualenv
+    venv_path = getattr(session.virtualenv, "location", sys.prefix)
     session.run("fawltydeps", f"--pyenv={venv_path}")
 
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -82,6 +82,13 @@ def integration_tests(session):
 
 
 @nox.session(python=python_versions)
+def self_test(session):
+    install_groups(session)
+    venv_path = session.run("poetry", "env", "info", "--path", silent=True).strip()
+    session.run("fawltydeps", f"--pyenv={venv_path}")
+
+
+@nox.session(python=python_versions)
 def lint(session):
     install_groups(session, include=["lint"])
     session.run("mypy")
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,4 +1,4 @@
-name: Tests 
+name: Tests
 
 on: [workflow_call]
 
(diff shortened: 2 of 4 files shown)


jherland (changes requested):
Thanks for looking into this. I like the direction of these changes, and they get the job done. 👍🏻 

There's one comment below, that I'd like you to reconsider before we merge this.

Another thing to think about (either in this PR or a future PR) is to not only run this FawltyDeps self-check in GitHub CI, but also make it available in our dev shell as a Nox action. This involves adding a new `self_test` function to `noxfile.py`, and changing `.github/workflows/tests.yaml` to instead:

```yaml
run: poetry run nox --no-venv --no-install --non-interactive -s self_test -- -vv
```

The advantage of this (and the reason we've done this for the other CI actions) is to make it easy to run the CI actions locally. This helps make CI failures less common (since we can run the same tests locally before we push) and it also make CI failures easier to debug (because we can easily run the same command localy: `nox -s self_test`).

zz1874:
Thanks for the review @jherland ! 
Totally agree that turning the CI check into a nox action is a smart move, as we can test it out on our own machines. So, I've gone ahead and made that change, and now the CI check runs smoothly using the new `self_test` function. Your input was awesome!

noxfile.py
```diff
@@ -81,6 +81,13 @@ def integration_tests(session):
     session.run("pytest", "-x", "-m", "integration", "--durations=10", *session.posargs)
 
 
+@nox.session(python=python_versions)
+def self_test(session):
+    install_groups(session)
+    venv_path = session.run("poetry", "env", "info", "--path", silent=True).strip()
```
jherland, line 87: This looks good, but since Nox usually manages a separate virtualenv per session (that's what the `install_groups(session)` sets up), I'd prefer if the `--pyenv` we pass to FawltyDeps is the one that _Nox_ has set up [^1], and not the bigger Poetry dev env. We can get this via `session.virtualenv.location`.

That said, this only applies when we run Nox locally. In GitHub CI we run `nox` with `--no-venv`[^2], and when that is the case `session.virtualenv` no longer has a `.location` member. In this case, Nox will simply run the commands (`fawltydeps` in our case) in the _same_ virtualenv that Nox itself is running in, so we can fall back to `sys.prefix` (which will point at the surrounding Poetry environment).

In short:
```suggestion
    # For --pyenv, use Nox's virtualenv, or fall back to current virtualenv
    venv_path = getattr(session.virtualenv, "location", sys.prefix)
```

[^1]: For one, the virtualenv created by Nox usually only contains the minimal dependencies (as opposed to the entire dev shell managed by Poetry), and it's good to check that this minimal env still covers what we need for FawltyDeps to run cleanly.

[^2]: The reason for using `--no-venv` in GitHub CI is somewhat complicated, related to GitHub caching, and explained in more detail in 2a57243ebb8a30640292652fe91e679723c8ef6f.

jherland (approved):
Great! Only one thing remaining, that I only realized while reading this latest version, so please don't feel bad for not catching it! 😅

PR #349, opened 2023-08-08
Make identity mapping and install-deps mutually exclusive

### LINKED ISSUE #299: Review behavior: exclusive mapping via `install_deps` or `identity_mapping`
The current behavior (taking into account the PRs currently under review) is that all dependencies not mapped through:
- user-defined mapping
- a local environment

will be matched by exactly one of the following mappings in this order:
- a local env pip-installed from a packages repository if `install_deps` is true
- identity mapping if `identity_mapping` is true

If none of these two mappings is turned on, these deps will be unresolved.

In other terms, it's not possible for the resolver to go through more than one mapping in the following set: {local pip-installed venv, identity mapping, unresolved}.

If `install_deps` is true and pip installing packages unresolved via the first two mappings (local env, user-defined) fails, FawltyDeps will error out and will not attempt to resolve uninstalled deps via other means.

Does this behavior make sense in all cases or can we envisage cases where it should be amended?

Nour-Mws:
This PR constitutes the last missing piece of the mapping puzzle.

Identity mapping is suppressed if and only if install-deps is chosen.
Prior to this PR, any dependencies unresolved by `pip install` were passed to identity mapping. It was practically impossible to have any unresolved dependencies and therefore to have the `UnresolvedDependencyError` raised. This error is now raised in the only case where the `TemporaryPipResolver` fails to install a given dependency. In all other cases, identity mapping still resolves any dependencies left unresolved by prior resolvers in the stack.

Closes #299

### DIFF
--- a/tests/test_install_deps.py
+++ b/tests/test_install_deps.py
@@ -41,6 +41,22 @@ def test_resolve_dependencies_install_deps__raises_unresolved_error_on_pip_insta
     assert all(word in caplog.text for word in ["pip", "install", "does_not_exist"])
 
 
+def test_resolve_dependencies_install_deps__unresolved_error_only_warns_failing_packages(
+    caplog, local_pypi
+):
+    # When we fail to install _some_ - but not all - packages, the error message
+    # should only mention the packages that we failed to install.
+    caplog.set_level(logging.WARNING)
+    deps = {"click", "does_not_exist", "leftpadx"}
+
+    with pytest.raises(UnresolvedDependenciesError):
+        resolve_dependencies(deps, setup_resolvers(install_deps=True))
+
+    assert "Failed to install 'does_not_exist'" in caplog.text
+    assert "Failed to install 'click'" not in caplog.text
+    assert "Failed to install 'leftpadx'" not in caplog.text
+
+
 def test_resolve_dependencies_install_deps_on_mixed_packages__raises_unresolved_error(
     caplog, local_pypi
 ):
--- a/tests/test_install_deps.py
+++ b/tests/test_install_deps.py
@@ -1,13 +1,15 @@
 """Verify behavior of TemporaryPipInstallResolver."""
 import logging
 
+import pytest
+
 from fawltydeps.packages import (
-    IdentityMapping,
     Package,
     TemporaryPipInstallResolver,
     resolve_dependencies,
     setup_resolvers,
 )
+from fawltydeps.types import UnresolvedDependenciesError
 
 
 def test_resolve_dependencies_install_deps__via_local_cache(local_pypi):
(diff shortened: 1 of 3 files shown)


tests/test_install_deps.py
```diff
@@ -23,42 +25,33 @@ def test_resolve_dependencies_install_deps__via_local_cache(local_pypi):
...
-    assert actual == {
-        "click": Package("click", {"click"}, TemporaryPipInstallResolver, debug_info),
-        "does_not_exist": Package(
-            "does_not_exist", {"does_not_exist"}, IdentityMapping
-        ),
-        "leftpadx": Package(
-            "leftpadx", {"leftpad"}, TemporaryPipInstallResolver, debug_info
-        ),
-    }
-    assert all(word in caplog.text for word in ["pip", "install", "does_not_exist"])
```
jherland, line 64: I realize while reviewing this that the work we did to make `TemporaryPipInstallResolver` able to install as many packages as possible on a _best effort_ basis (when some packages fail to install) is currently not very useful: We always end up with `UnresolvedDependenciesError` and aborting in those cases. (That said the list of packages passed into `UnresolvedDependenciesError` is more accurate.)

I assure you that this is _not_ something I want to change now! I'm just starting to feel a small itch about maybe having the option to allow `TemporaryPipInstallResolver` and `IdentityMapping` at the same time. Let's see what our users say after we've released the mapping strategy milestone.

jherland (approved):
Great work! 🚀

tests/test_install_deps.py
Nour-Mws: Yes that is an unfortunate side-effect of raising an error on unresolved dependencies inside `resolve_dependencies()` itself. I wasn't able to assert that some dependencies were resolved with certain resolvers. I guess we can still assert that the error message refers to exactly the unresolved dependencies (which is in itself very useful data for the user to have)?

As for allowing identity mapping, I'm not convinced. If a dependency is not pip-installable, the best course of action to maintain "correctness" is either for the user to provide it in a local environment accessible to FawltyDeps or to provide it via a custom mapping.

jherland: Agreed, I wonder if there is room for a test in between `test_resolve_dependencies_install_deps__raises_unresolved_error_on_pip_install_failure` and `test_resolve_dependencies_install_deps_on_mixed_packages__raises_unresolved_error`, to verify that when we install `{"click", "does_not_exist", "leftpadx"}`, the error message makes it clear that only `does_not_exist` failed, and not the two other packages.

I think your reasoning around the identity mapping makes sense as well. Let's wait and see what users think about the new mapping strategy before we make any more changes here.

tests/test_install_deps.py
```diff
@@ -23,42 +25,33 @@ def test_resolve_dependencies_install_deps__via_local_cache(local_pypi):
...
-    }
+
+    with pytest.raises(UnresolvedDependenciesError):
+        resolve_dependencies(["does_not_exist"], setup_resolvers(install_deps=True))
+
     assert all(word in caplog.text for word in ["pip", "install", "does_not_exist"])
 
 
-def test_resolve_dependencies_install_deps__pip_install_some_packages(
+def test_resolve_dependencies_install_deps_on_mixed_packages__raises_unresolved_error(
```
jherland, line 44: In my comment about `UnresolvedDependenciesError`, here is the kind of test I was thinking about. This verifies that we preserve the nuance of what packages actually failed in our error messaging:

```suggestion
def test_resolve_dependencies_install_deps__unresolved_error_only_warns_failing_packages(
    caplog, local_pypi
):
    # When we fail to install _some_ - but not all - packages, the error message
    # should only mention the packages that we failed to install.
    caplog.set_level(logging.WARNING)
    deps = {"click", "does_not_exist", "leftpadx"}

    with pytest.raises(UnresolvedDependenciesError):
        resolve_dependencies(deps, setup_resolvers(install_deps=True))

    assert "Failed to install 'does_not_exist'" in caplog.text
    assert "Failed to install 'click'" not in caplog.text
    assert "Failed to install 'leftpadx'" not in caplog.text


def test_resolve_dependencies_install_deps_on_mixed_packages__raises_unresolved_error(
```

Nour-Mws: I somehow missed this comment. Just committed the change. Will merge once the CI passes.

PR #350, opened 2023-08-15
Verify fallback to current venv when no other env is found

jherland:
Add a couple of tests to ensure that we fall back to using the current
Python environment (aka. sys.path) whenever we fail to find any
other/better Python environments to use for resolving dependencies.

(While going over the blog post, I wanted to ensure that the desired
behavior we're describing is in fact reflected in the current version
of FD, and I didn't find existing tests that covered this directly.)

### DIFF
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -14,6 +14,7 @@ from textwrap import dedent
 from typing import List
 
 import pytest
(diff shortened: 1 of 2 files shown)


Nour-Mws (approved):
Looks good!
I had to take a deep dive into the code to check out the tests behavior, hence the late review.

PR #351, opened 2023-08-16
Add Support for Dynamic Dependencies in pyproject.toml

### LINKED ISSUE #347: Not supporting PEP-517 and new python build based packaging
It seems that FawltyDeps is unable to correctly detect dependencies from modern projects that are using `build` for packaging, and is missing to spot them.

Example: https://github.com/ansible/ansible-lint/blob/main/pyproject.toml#L12 which is using setuptools-scm with dynamic dependencies as build backend. In this case none of the correctly declared dependencies is detected and the tool reports all dependencies as undeclared, which is not true.

zz1874:
Closes #347 

This PR introduces support for dynamic dependencies declared in the `pyproject.toml` file under `tool.setuptools.dynamic` section.

In the past, our project had a feature where it did not recurse into subdirectories under the project root (`.`) to detect requirements files. This PR enables our project to detect requirements files located in any directory as long as they are declared under `tool.setuptools.dynamic` section.


jherland (commented):
(Not a full review, but I had some free minutes, and wanted to take a quick look...)

I love how this is coming together, and how little code you have to change to make this work. 😃 

One question:
Should we check/require the `project.dynamic` directive before looking at the `tool.setuptools.dynamic`? I'm looking at https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html to understand how this should work, and it seems to me that `tool.setuptools.dynamic` should not be looked at when the corrsponding key in `project.dynamic` is missing?

zz1874:
@jherland Thanks for the review! 

> Should we check/require the project.dynamic directive before looking at the `tool.setuptools.dynamic`?

Yes I've also spotted this issue. What I was thinking is that it takes the "same" effort to look at either `project.dynamic` or `tool.setuptools.dynamic`. But of course we can look at `project.dynamic` first as an initial condition.

fawltydeps/extract_declared_dependencies.py
```diff
@@ -288,6 +288,28 @@ def parse_pyproject_toml(path: Path) -> Iterator[DeclaredDependency]:
...
+            for optional_deps_item in [
+                optional_deps_section.get("file", {})
+                for optional_deps_section in dynamic_section.get(
+                    "optional-dependencies", {}
+                ).values()
+            ]
+            for file_list in optional_deps_item
+        ]
+        dynamic_files = dynamic_deps_files + dynamic_optional_deps_files
+        if isinstance(dynamic_files, list):
```
Nour-Mws, line 305: Why is this test necessary?

fawltydeps/extract_declared_dependencies.py
```diff
@@ -288,6 +288,28 @@ def parse_pyproject_toml(path: Path) -> Iterator[DeclaredDependency]:
...
+            for req_file in dynamic_files:
+                req_file_path = Path(req_file)
```
Nour-Mws, line 307: From this [example](https://github.com/ansible/ansible-lint/blob/f6961ffcc70748753110efb84ec6ca6dcd5fa128/pyproject.toml#L271-L275), it seems like the provided path is a relative one. Do we need to compute the full path of the file by joining with the basepath?

fawltydeps/extract_declared_dependencies.py
```diff
@@ -288,6 +288,28 @@ def parse_pyproject_toml(path: Path) -> Iterator[DeclaredDependency]:
...
+                if req_file_path.exists():
+                    yield from parse_requirements_txt(req_file_path)
+                else:
+                    logger.debug("%s does not exist. Skipping", req_file_path)
```
Nour-Mws, line 311: Hmm.. If the file is not found, then we have a problem. I am tempted to error out in this case.
@jherland thoughts?

Nour-Mws (commented):
Looks good so far! I think the tests will help us see better if anything is missing.
As I said offline, being able to add these requirements files to the traversal object would have been a neat solution but we unfortunately don't have a direct way to append to the list of dependencies files to traverse from this function.

jherland:
@zz1874:
> > Should we check/require the `project.dynamic` directive before looking at the `tool.setuptools.dynamic`?
> 
> Yes I've also spotted this issue. What I was thinking is that it takes the "same" effort to look at either `project.dynamic` or `tool.setuptools.dynamic`. But of course we can look at `project.dynamic` first as an initial condition.

I believe we should look at `project.dynamic` as an initial condition, yes. According to https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata (which is the closest thing I have found to an authoritative spec here), the `tool.setuptools.dynamic` only correspond to what is found/requested in `project.dynamic`. If `project.dynamic` is missing I think we should NOT look for corresponding entries in `tool.setuptools.dynamic` at all.

@Nour-Mws:
> As I said offline, being able to add these requirements files to the traversal object would have been a neat solution but we unfortunately don't have a direct way to append to the list of dependencies files to traverse from this function.

I agree that we currently do not have a good way to feed this information back into the traversal. But I don't know if it would be very useful to update the traversal, since the traversal phase is already done/finished _before_ we parse this `pyproject.toml` and discover these dynamic fields?

Currently, since we don't _parse_ files during the traversal phase, we would not expect these extra requirements files to show up in the `--list-sources` output (unless they coincidentally are _also_ found by the traversal).

If we really want them to show up in `--list-sources`, we would have to move (some of) our `pyproject.toml` parsing into the traversal stage, and I'm not sure it's ultimately worth it?

fawltydeps/extract_declared_dependencies.py
jherland: Hmm. For most projects I expect the `pyproject.toml` to be located in the root dir of the project, so it shouldn't matter (as FD would typically run from the same dir), but, yes, I expect there are scenarios where we need to be more careful.

AFAICS, [setuptools](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata) does not specify the base to be used for these relative paths, but I think it's fairly safe to assume that it should be the directory that contains the `pyproject.toml` file itself. Hence the last line here should probably be something like:

```python
                req_file_path = path.with_name(req_file)
```
or
```python
                req_file_path = path / req_file
```

Definitely something we should have a test case for....

fawltydeps/extract_declared_dependencies.py
jherland: Hmm. Maybe error out, or at least print a very visible warning/error message before continuing.

How do we currently handle other serious (but recoverable) errors in dep file parsing? E.g. a missing field in pyproject.toml or similar?

Nour-Mws:
>Currently, since we don't parse files during the traversal phase, we would not expect these extra requirements files to show up in the --list-sources output (unless they coincidentally are also found by the traversal).

>If we really want them to show up in --list-sources, we would have to move (some of) our pyproject.toml parsing into the traversal stage, and I'm not sure it's ultimately worth it?

@jherland, I don't think it's worth the hassle really (i.e. parsing pyproject.toml in the traversal stage). On principle, yes, I think these requirements files should appear in the sources. You're getting the requirements in these files by parsing them, and not directly from pyproject.toml so I would argue they are a direct source. But I don't think this is worth spending any time on.

tests/conftest.py
```diff
@@ -175,7 +178,25 @@ def format_setup_cfg(deps: Deps, no_extras: ExtraDeps) -> str:
             + "\n".join(f"    {d}" for d in deps)
         )
 
-    def format_pyproject_toml(deps: Deps, extras: ExtraDeps) -> str:
+    def format_pyproject_toml(deps: Deps, extras: ExtraDeps, is_dynamic: bool) -> str:
```
Nour-Mws, line 181: I feel uneasy about re-using the `deps` and `extras` argument to pass the dynamic requirements files as, although these are technically `List[str]` and `Dict[str, Deps]` respectively, they are not dependencies but rather paths. Would it be possible to define new arguments to hold these requirements files?
Incidentally, that would also make it possible to define both static and dynamic dependencies in the `pyproject.toml` file (see my comment below).

fawltydeps/extract_declared_dependencies.py
```diff
@@ -288,27 +288,32 @@ def parse_pyproject_toml(path: Path) -> Iterator[DeclaredDependency]:
...
-                for optional_deps_section in dynamic_section.get(
-                    "optional-dependencies", {}
-                ).values()
+    if "dynamic" in parsed_contents.get("project", {}) and (
+        "dependencies" in parsed_contents["project"]["dynamic"]
+        or "optional-dependencies" in parsed_contents["project"]["dynamic"]
+    ):
+        if "dynamic" in parsed_contents.get("tool", {}).get("setuptools", {}):
+            dynamic_section = parsed_contents["tool"]["setuptools"]["dynamic"]
+            dynamic_deps_files = dynamic_section.get("dependencies", {}).get("file", [])
```
Nour-Mws, line 297: I feel this is getting hard to read. Maybe adding a comment before each of these conditions to indicate what we're trying to do would help?

fawltydeps/extract_declared_dependencies.py
Nour-Mws: Closing this as it's addressed in [72aa2cd](https://github.com/tweag/FawltyDeps/pull/351/commits/72aa2cd1834e29b2bf9e0237cf2c9489b28c999b).

fawltydeps/extract_declared_dependencies.py
Nour-Mws: It seems we're erroring out when a pyproject element is mistyped:
https://github.com/tweag/FawltyDeps/blob/86d07b28990bea0d5bc7a6f354176b9cac27ec3f/fawltydeps/extract_declared_dependencies.py#L259-L267

https://github.com/tweag/FawltyDeps/blob/86d07b28990bea0d5bc7a6f354176b9cac27ec3f/fawltydeps/extract_declared_dependencies.py#L207

https://github.com/tweag/FawltyDeps/blob/86d07b28990bea0d5bc7a6f354176b9cac27ec3f/fawltydeps/extract_declared_dependencies.py#L228

Nour-Mws (changes requested):
I'd just like to say I really admire the way you've been able to get up and running on the test suite!
Great work!
I've left a couple of comments. Some are nitpicky (and I flagged them as such) but others would need more consideration.

fawltydeps/extract_declared_dependencies.py
jherland: The `logger.error(...)` on line 260 prints an error message, but the rest of the exception handler does not re-raise the exception, or otherwise abort execution. The surrounding loop keeps going, and we return normally at the end, AFAICS.

I suspect something similar can be done in this case:
- raise a `FileNotFoundError` (or maybe it is already raised by `parse_requirements_txt()`?) at the point where we try to open the requirements files
- catch this exception somewhere suitable, probably inside the `for` loop on line 306, so that we can keep trying to parse the other dynamic files?
- In the exception handler: log an error message, and then keep on running.

tests/conftest.py
```diff
@@ -191,20 +212,20 @@ def format_pyproject_toml(deps: Deps, extras: ExtraDeps) -> str:
...
     def format_deps(
-        filename: str, all_deps: Union[Deps, Tuple[Deps, ExtraDeps]]
+        filename: str, all_deps: Union[Deps, Tuple[Deps, ExtraDeps]], is_dynamic: bool
     ) -> str:
         if isinstance(all_deps, tuple):
             deps, extras = all_deps
         else:
             deps, extras = all_deps, {}
-        formatters: Dict[str, Callable[[Deps, ExtraDeps], str]] = {
+        formatters: Dict[str, Callable[[Deps, ExtraDeps, bool], str]] = {
```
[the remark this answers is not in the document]
jherland: > @jherland, was there any reason we preferred originally using assertions and having a unified interface for these formatters?

I think it comes down to the formatter selection on line 227, and the subsequent call on line 228. Since we do a single call for the chosen formatter (whichever it ends up being), they must have a single, unified interface (i.e. at the point we make the call we no longer know/care which formatter we call).

We _could_ attempt to diversify the interface, and then use `functools.partial` to prepare the different argument lists for each formatter. I think it would look something like this:

```python
from functools import partial
...
        formatters: Dict[str, Callable[[], str]] = {
            "requirements.txt": partial(format_requirements_txt, deps),  # default choice
            "setup.py": partial(format_setup_py, deps, extras),
            "setup.cfg": partial(format_setup_cfg, deps),
            "pyproject.toml": partial(format_pyproject_toml, deps extras, is_dynamic),
        }
        formatter = formatters.get(Path(filename).name, formatters["requirements.txt"])
        return formatter()
```

but I'm not sure this is really any better:
- `partial` functions are often harder to read/understand,
- I don't like the duplication of args inside the `formatters` dict,
- we lose the asserts that would catch misuse of `extras`/`is_dynamic` with formatters that don't support them.

tests/conftest.py
```diff
@@ -175,7 +178,25 @@ def format_setup_cfg(deps: Deps, no_extras: ExtraDeps) -> str:
...
+
+                    [tool.setuptools.dynamic]
+
+                    dependencies = {{ file = {deps!r} }}
+
+                    [tool.setuptools.dynamic.optional-dependencies]
+                    """
+                )
+                + "\n".join(f"{k} = {{ file = {v!r} }}" for k, v in extras.items())
+            )
```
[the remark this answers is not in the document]
jherland: I believe they are mutually exclusive.

https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata says:
> Currently, when specifying `optional-dependencies` dynamically, all of the groups must be specified dynamically; one can not specify some of them statically and some of them dynamically.

It does not say anything about `dependencies` per se, but I would _assume_ that a similar argument applies here: you _either_ have `dependencies` in `project.dynamic`, or you don't, and that decides whether you look at `tool.setuptools.dynamic.dependencies` or `project.dependencies`, respectively. I don't think you're allowed to mix them.

tests/conftest.py
```diff
@@ -216,18 +237,21 @@ def create_one_fake_project(
             Dict[str, Union[Deps, Tuple[Deps, ExtraDeps]]]
         ] = None,
         extra_file_contents: Optional[Dict[str, str]] = None,
+        is_dynamic: bool = False,
```
jherland, line 240: Thinking some more about how to handle dynamic `pyproject.toml`s: Instead of having an `is_dynamic` flag here, you could instead have a `create_dynamic_pyproject_toml` bool flag. Instead of passing this flag on to `format_deps()` and the underlying formatter function, you can add something like this around line 255:

```python
        if files_with_declared_deps is not None:
            for filename, all_deps in files_with_declared_deps.items():
                tmp_files[filename] = format_deps(filename, all_deps, is_dynamic)
            if create_dynamic_pyproject_toml:
                # Split files_with_declared_deps arbitrarily into regular and optional deps
                # and construct a pyproject.toml with dynamic sections pointing to them.
                req_files = list(files_with_declared_deps.keys())
                deps, optional_deps = req_files[:len(req_files) / 2], req_files[len(req_files) / 2:]
                tmp_files["pyproject.toml"] = f"""\
                    [project]
                    name = "MyLib"
                    dynamic = ["dependencies", "optional-dependencies"]
                    [tool.setuptools.dynamic]
                    dependencies = {{ file = {deps!r} }}
                    optional-dependencies.test = {{ file = {optional_deps!r} }}
                    """
        if extra_file_contents is not None:
            ...
```

fawltydeps/extract_declared_dependencies.py
jherland: This is one of those cases where I feel that looking things up directly and handling the resulting exceptions (aka. "ask forgiveness rather than permission") leads to nicer code (less repetition and nesting of `if`s):

```suggestion
    try:
        dynamic = parsed_contents["project"]["dynamic"]
    except KeyError:
        dynamic = []

    deps_files = []
    try:
        if "dependencies" in dynamic:
             deps_files = parsed_contents["tool"]["setuptools"]["dynamic"]["dependencies"]["file"]
    except KeyError:
        pass

    optional_deps_files = {}
    try:
        if "optional-dependencies" in dynamic:
             optional_deps = parsed_contents["tool"]["setuptools"]["dynamic"]["optional-dependencies"]
             optional_deps_files = {k: v["file"] for k, v in optional_deps.items()}
    except KeyError:
        pass

    dynamic_files = deps_files + optional_deps_files
    for req_file in dynamic_files:
        ...
```

Also, looking at the rest of this function, I wonder if it makes sense to put all the new/dynamic code into a seprarate helper, named `parse_dynamic_pyproject_contents()` or similar.

(Finally I also suspect we'd need to select between calling this helper, and calling `parse_pep621_pyproject_contents` depending on the contents of the `project.dynamic` section...)

tests/conftest.py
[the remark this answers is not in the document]
Nour-Mws: Your analysis makes sense. I didn't follow up my logic up until line 228. I am closing this. Let's stick with the unified interface.

Nour-Mws:
With PR #354 merged, the CI should now pass once this is rebased on main.

fawltydeps/extract_declared_dependencies.py
Nour-Mws: >raise a FileNotFoundError (or maybe it is already raised by parse_requirements_txt()?) at the point where we try to open the requirements files

Oh that's a good point. If that's the case then our job is done for us.
Otherwise, both options 2 and 3 are ok. As long as we log the error very visibly, there's no need to error out.

Nour-Mws: So I actually created a minimal pyproject.toml based on the ansible-lint pyproject.toml, and didn't create the actual requirements files. This is what I got when I tried building the package:
>[..]/python3.9/site-packages/setuptools/config/expand.py:134: SetuptoolsWarning: File '/home/nour/python-projects/pyproject-dynamic-deps/.config/requirements.in' cannot be found

Same for the other requirements files. The build concluded successfully and I only got these warnings with the verbose mode (`pip wheel . -w dist -v`). I still think we should show the error by default but maybe we can still continue the analysis.

tests/conftest.py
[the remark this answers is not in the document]
Nour-Mws: Continuing with the experimentation I mentioned in another example. I've created the following scenarios:
These build successfully:
- both `dependencies` and `optional-dependencies` defined in `[project].dynamic`, then in `[tool.setuptools.dynamic]`
- only `dependencies` defined in `[project].dynamic` and `[tool.setuptools.dynamic]`. Some optional-dependencies defined under `[project.optional-dependencies]`.
- only `optional-dependencies` defined in `[project].dynamic` and `[tool.setuptools.dynamic]`. `dependencies` defined under `[project.dependencies]`.

This raises an error:
- `dependencies` defined both in `[project].dynamic` (+ `[tool.setuptools.dynamic]`) and also as its own `[project.dependencies]`.

`raise suffixed_err(src, pos, "Invalid value")
pip._vendor.tomli._parser.TOMLDecodeError: Invalid value (at line 17, column 17)`

where line 17 is where `[project].dependencies` is defined.

Conclusion: 
- You can have any combination over `dependencies` or `optional-dependencies` being dynamic or static.
- As soon as one category is declared dynamic, you can no longer declare it statically (but you can still declare the other one statically if not in dynamic).

fawltydeps/extract_declared_dependencies.py
jherland: Thanks for digging into how setuptools handles this. Would be perfect to add that minimal project as a real_project (or sample_project) test?

Agree that the way to handle this is loud error message, but still continue.

fawltydeps/extract_declared_dependencies.py
zz1874: Thanks for the clarifications!
I adopted the second option here 
```
        req_file_path = path.parent / req_file
```
since `path` refers to `basepath/pyproject.toml`

tests/conftest.py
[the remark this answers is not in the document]
zz1874: Thanks @Nour-Mws ! That's super helpful! :+1: 
In this way, `parse_dynamic_pyproject_contents` and `parse_pep621_pyproject_contents` are not mutually exclusive. Will fix that!

fawltydeps/extract_declared_dependencies.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/351/commits/4242069da6847018533f95b1a4af1ee86159a4a0

fawltydeps/extract_declared_dependencies.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/351/commits/4242069da6847018533f95b1a4af1ee86159a4a0 and https://github.com/tweag/FawltyDeps/pull/351/commits/4c350f5438802703a55cbe67bc02ca452aef021d

fawltydeps/extract_declared_dependencies.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/351/commits/4242069da6847018533f95b1a4af1ee86159a4a0 and https://github.com/tweag/FawltyDeps/pull/351/commits/4c350f5438802703a55cbe67bc02ca452aef021d

tests/test_extract_declared_dependencies_success.py
```diff
@@ -525,6 +525,41 @@ def test_find_and_parse_sources__project_with_pyproject__returns_list(fake_proje
...
+                [
+                    (tmp_dir / ".subdir/requirements.txt").as_posix()
+                ],  # dynamic dependencies path
+                {
+                    "test": [(tmp_dir / ".subdir/requirements-test.txt").as_posix()]
+                },  # dynamic optional-dependencies path
+            ),
+        },
+        is_dynamic=True,
+    )
```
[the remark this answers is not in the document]
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/351/commits/500f0667c1bbb7529465f645b28f4a6c7b755935

tests/conftest.py
```diff
@@ -145,11 +145,14 @@ def fake_project(write_tmp_files, fake_venv):
     def format_python_code(imports: Imports) -> str:
         return "".join(f"import {s}\n" for s in imports)
 
-    def format_requirements_txt(deps: Deps, no_extras: ExtraDeps) -> str:
-        assert not no_extras  # not supported
+    def format_requirements_txt(
+        deps: Deps, no_extras: ExtraDeps, no_dynamic: bool
+    ) -> str:
+        assert not no_extras, no_dynamic  # not supported
```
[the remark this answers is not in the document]
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/351/commits/500f0667c1bbb7529465f645b28f4a6c7b755935
In this commit, I removed the `is_dynamic`-related variables and hardcoded the dynamic section under each test.

tests/conftest.py
```diff
@@ -160,8 +163,8 @@ def format_setup_py(deps: Deps, extras: ExtraDeps) -> str:
             )
             """
 
-    def format_setup_cfg(deps: Deps, no_extras: ExtraDeps) -> str:
-        assert not no_extras  # not supported
+    def format_setup_cfg(deps: Deps, no_extras: ExtraDeps, no_dynamic: bool) -> str:
+        assert not no_extras, no_dynamic  # not supported
```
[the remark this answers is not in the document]
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/351/commits/500f0667c1bbb7529465f645b28f4a6c7b755935

tests/conftest.py
zz1874: Having a `create_dynamic_pyproject_toml` flag is a good idea. The only reason why I decided to hardcode the dynamic section under each test at last is that we also need to test the case where the toml file has both `[project.dependencies]` and `[tool.setuptools.dynamic.dependencies]`. If we turn on this `create_dynamic_pyproject_toml` flag, it will be complicated to add some non-dynamic dependencies because the pyproject.toml in the fake project will be replaced. 
Please let me know if hardcoding the dynamic section looks good to you!

fawltydeps/extract_declared_dependencies.py
```diff
@@ -281,6 +322,9 @@ def parse_pyproject_toml(path: Path) -> Iterator[DeclaredDependency]:
 
     yield from parse_pep621_pyproject_contents(parsed_contents, source)
 
+    if "dynamic" in parsed_contents.get("project", {}):
+        yield from parse_dynamic_pyproject_contents(parsed_contents, source)
```
jherland, line 326: Just one quick thing I noticed before I do another proper review:

I believe the dynamic fields are part of PEP621? Or at least more closely related to PEP621's fields than they are to Poetry's fields? In the above lines we _first_ parse the PEP621 fields, and then, if the dynamic fields are present, we _also_ parse them? Shouldn't this be more mutually exclusive? I.e. what happens if a project has both `project.dependencies` and `"dependencies"` in `project.dynamic`? IIUC we should then not look at `project.dependencies` at all?

zz1874: I just re-read @Nour-Mws 's comment again:

> You can have any combination over dependencies or optional-dependencies being dynamic or static.
As soon as one category is declared dynamic, you can no longer declare it statically (but you can still declare the other one statically if not in dynamic).

So if a project has both `project.dependencies` and `"dependencies"` in `project.dynamic`, we should not look at `project.dependencies` at all (which I misunderstood in the last two commits :sweat:  ). To be precise, if a project has `project.dependencies` and `optional-dependencies` in `project.dynamic`, we should look at them both, so `parse_dynamic_pyproject_contents` should be embedded in `parse_pep621_pyproject_contents`?

zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/351/commits/353ab19dfc899fbd8a2b61e2fbf9100f7133fe3b

tests/conftest.py
jherland: Yes, hardcoding the `pyproject.toml` in these tests is good with me. Adding a helper to generate it is only useful if it's easier to read the tests _with_ the helper than _without_, and for the number of tests (few) and complexity (high) here, it's not worth it.

fawltydeps/extract_declared_dependencies.py
jherland: Yes, I believe the code is correct now, although I'm not 100% sure what happens when `project.dynamic` contains only one of `"dependencies"` or `"other-dependencies"`and not the other. Maybe worth adding two more test cases to make sure we handle all these combinations?

Nitpick alert: Looking at the code in `parse_pep621_pyproject_contents()` and `parse_dynamic_pyproject_contents()` I notice some duplication, specifically in looking up `parsed_contents["project"]["dynamic"]`. Maybe `parse_dynamic_pyproject_contents()` can be refactored into nested functions inside `parse_pep621_pyproject_contents()`, like we already have `parse_main()` and `parse_optional()`? Anyway, If you're so inclined I encourage you to play around with these functions to find the most readable solution.

jherland (approved):
There's one discussion where I just posted a comment, and left unresolved. Except for that, I think all other comments have been resolved, and I'm very happy to see this merged. 😄 Great work!

fawltydeps/extract_declared_dependencies.py
Nour-Mws: >when project.dynamic contains only one of "dependencies" or "other-dependencies"and not the other. Maybe worth adding two more test cases to make sure we handle all these combinations?

I second that. It would be good to test the following cases:
- `dependencies` in `dynamic` but not `optional-dependencies` (in this case `dependencies` should be read from `["setuptools"]["dynamic"][dependencies"]` and `optional-dependencies` from `["project"]["optional-dependencies"]`)
- the opposite case.

Nour-Mws:
Great improvements! Thanks!
I've added one comment [here](https://github.com/tweag/FawltyDeps/pull/351#discussion_r1305898654) to second @jherland's suggestion of adding two more tests. Once added, the two remaining conversations can be closed and this merged!

Nour-Mws (approved):
This is ready to be merged!

PR #354, opened 2023-08-22
Pin version of Poetry in CI to 1.5.1 to maintain support for Python 3.7

### LINKED ISSUE #352: Pin Poetry version in out GitHub Action
https://github.com/tweag/FawltyDeps/actions/runs/5929081642/job/16075993067?pr=351

Poetry 1.6.0 has dropped support for Python 3.7. We should pin Poetry to an older version for as long as we need to support v3.7.

Nour-Mws:
Closes #352

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -17,7 +17,10 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-      - uses: Gr1N/setup-poetry@v8
+      - name: Set up Poetry
+        uses: Gr1N/setup-poetry@v8
+        with:
+          poetry-version: "1.5.1" # support for Python 3.7 was dropped in 1.6.0
       - uses: actions/cache@v2
         with:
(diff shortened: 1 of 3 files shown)


jherland (approved):
Looks good. Thanks for fixing! 🚀 

It occurs to me that _outside_ of GitHub CI, we _also_ don't really constrain the Poetry version. Sure, if you use our Nix shell, we have pinned the nixpkgs version in `shell.nix`, which in effect also pins Poetry at whatever version is used in that nixpkgs. But if you _don't_ use Nix, then all bets are off.

I looked for a way to constrain the Poetry version from within `pyproject.toml` (it would not actually cause the right Poetry version to be used, but would at least fail loudly if the wrong version is used). However, some searching found https://github.com/python-poetry/poetry/issues/6644 (and https://github.com/python-poetry/poetry/issues/3316) which suggest that this is not a high priority for the Poetry team, especially not the part about restricting a _newer_ Poetry from being used.

So in the end, I think there is not much else we can/should do for the non-GitHub situation.

PR #363, opened 2023-09-01
Split the self_test part of the tests into another file

zz1874:
This PR is a fix of `self_test` (#123) while working on #362.

The `self_test` always failed when we added the functionality of parsing `typeshed` packages like `types-setuptools` because it wasn't installed in our virtual environment of CI. 

In our `tests.yaml`, we have
```
     - name: Install project
        run: poetry install --no-interaction --sync --with=nox,test
```
before 
```
      - name: Run fawltydeps on Python ${{ matrix.python-version }}
        run: poetry run nox --no-venv --no-install --non-interactive -s self_test -- -vv
```
while running fawltydeps against itself needs to install all optional dependencies, including `"nox", "test", "lint", "format", "dev"`.

We cannot just install the projects with all optional dependencies in `tests.yaml` because in our unit tests, we checked `click` package as an example, which will be installed when including `format`, and then the `tests/test_install_deps.py::test_resolve_dependencies_install_deps__via_local_cache` will fail because it will use `LocalPackageResolver` instead of `TemporaryPipInstallResolver`. (The detailed logs is [here](https://github.com/tweag/FawltyDeps/actions/runs/6039841344/job/16389390284))

So, after thinking of this, splitting the `self_test` part of the `tests` may be a better solution. In this way, in the self test, we install all dependencies and optional dependencies and do a *complete* test, and in the normal `tests`, we do the remaining tests with only `"nox"` and `"test"` groups installed.


Nour-Mws (commented):
Thanks for figuring this up @zz1874! This was not straightforward to debug.
I took a little pause on the inclusion of all deps groups in the nox session as I imagine `lint` would be enough. But then reasoned that since we're testing FawltyDeps as a project, it makes sense to include all deps groups in `pyproject.toml` in the environment we're using to resolve the deps.
Could you please reduce this PR to the last commit only? We can rebase PR #362 once that's done and re-run the CI there.

Nour-Mws (approved):

Issue #328, opened 2023-05-11
Reconsider how multi-valued options combine between the config file and the command-line?


jherland:
How should multi-valued options on the command-line (e.g. `--code foo.py bar.py`) combine with corresponding multi-valued options from the config file (e.g. `tool.fawltydeps.code = ["bar.py", "baz.py"]`). (Also don't forget the `fawltydeps_*` environment variables that exist between the config file and command-line in the configuration cascade...)

The current behavior is to override/replace completely for all `Settings` members that are defined at each level. No effort is taken to more "smartly" integrate the values at one level (e.g. the config file) with values at another level (e.g. the command line).

Quoting the relevant discussion on #321:

@Nour-Mws writes:
> Are we also defaulting to override config with cli arguments for other options that might have multiple values?

@jherland writes:
> Yes, here's what happens for e.g. `.code` (using environment vs. CLI, but the same holds for config file vs. CLI):
> ```sh
> $ fawltydeps_code='["fawltydeps/main.py", "fawltydeps/packages.py"]' \
>   fawltydeps --list-imports --detailed
> fawltydeps/main.py:20: pydantic
> fawltydeps/packages.py:20: importlib_metadata
> fawltydeps/packages.py:37: tomli
> 
> $ fawltydeps_code='["fawltydeps/main.py", "fawltydeps/packages.py"]' \
>   fawltydeps --list-imports --detailed --code fawltydeps/types.py fawltydeps/utils.py
> fawltydeps/types.py:15: typing_extensions
> fawltydeps/utils.py:9: importlib_metadata
> ```
> When I first wrote the `Settings` class, I added the following TODO (and it's still there):
> ```
>     TODO? Currently overrides happen _completely_, for example specifying
>     --ignore-undeclared on the command-line will _replace_ an ignore_undeclared
>     directive in pyproject.toml. We may want to consider allowing some
>     directives to _combine_ (although that will carry further complications).
> ```
> The "further complications" I allude to here are more complex than one might imagine at first:
> 
> - We would probably need to introspect over the `Settings` members and/or their type and define different ways of combining them:
>   - Most sets (`code`/`deps`/`pyenvs`/`custom_mapping_file`/`ignore_*`) combine by _union_
>   - `custom_mapping` combines by _merging dicts_ (a la `accumulate_mappings()`)
>   - How do we want `actions` to combine? One could argue for either _replace_ or _union_, but probably _replace_ makes most sense.
>   - `output_format` should likely _replace_?
[...]

PR #365, opened 2023-09-07
Exclude development tools from unused dependencies detection

### LINKED ISSUE #117: How to handle tools (not imported) that end up being reported as "unused" dependencies
From https://github.com/tweag/FawltyDeps/pull/112#discussion_r1097701997:
> Rather I think this is a (separate) issue with FD: When calculating/presenting "unused" dependencies, we should consider ignoring/skipping common tools that are typically declared as dependencies but seldom imported. Examples, just from our own project config:
> - black
> - isort
> - codespell
> - mypy
> - nox
> - pylint
> - pytest
> - fawltydeps 🎉 
>
> Maybe we could try to be smarter about it, like looking for tool configurations in pyproject.toml as proof that a tool is actually being used? But I suspect a simple allow-list might be good enough.

More thought is probably needed about how best to solve this, but we will probably want to do something about this fairly soon, as there are enough of these tools that will keep showing up as "unused dependencies"...

zz1874:
### Description
Closes #117 

This PR addresses the issue of handling tools (development dependencies) that are reported as "unused" in the project's dependencies. The solution involves introducing a `default_ignore_unused` list that includes common development tools known not to be imported directly in the project's scripts. 

There are two scenarios:
- If the user does not provide their own `ignore_unused` option, Fawltydeps will use this `default_ignore_unused` list as its default, automatically detecting and ignoring the specified development tools. 
- If the user provides their own `ignore_unused` option, it will overwrite the default one, and Fawltydeps will use the user-defined list instead.

### Changes Made
- Added a `default_ignore_unused` list to specify default development tools to be ignored.
- Updated unit tests `test_cmdline_on_ignored_undeclared_option` to include `check_unused_action_on_default_ignored_unused_dep__outputs_nothing` (automatically detect using default ignore_unused list) and `check_unused_action_on_overriden_ignored_unused_dep__outputs_nothing` (using user defined ignore_unused list)
- Updated integration tests for real projects


fawltydeps/settings.py
```diff
@@ -93,6 +93,27 @@ def parse_path_or_stdin(arg: str) -> PathOrSpecial:
...
+        "isort",
+        "mypy",
+        "nox",
+        "pylint",
+        "pyright",
+        "pytest",
+        "wheel",
+        "tox",
+    ]
+)
```
jherland, line 114: I think I prefer this constant with UPPER_CASE spelling. Also, suggest to use the `{ ... }` literal set syntax instead of `set([ ... ])`:

```suggestion
DEFAULT_IGNORE_UNUSED = {
    # Development tools not meant to be imported
    "black",
    "codespell",
    "colorama",
    "coverage",
    "fawltydeps",
    "flake8",
    "hypothesis",
    "isort",
    "mypy",
    "nox",
    "pylint",
    "pyright",
    "pytest",
    "wheel",
    "tox",
}
```

fawltydeps/settings.py
```diff
@@ -114,7 +135,7 @@ class Settings(BaseSettings):  # type: ignore
     pyenvs: Set[Path] = {Path(".")}
     custom_mapping: Optional[CustomMapping] = None
     ignore_undeclared: Set[str] = set()
-    ignore_unused: Set[str] = set()
+    ignore_unused: Set[str] = default_ignore_unused
```
jherland, line 138: To reflect the above:

```suggestion
    ignore_unused: Set[str] = DEFAULT_IGNORE_UNUSED
```

fawltydeps/settings.py
```diff
@@ -93,6 +93,27 @@ def parse_path_or_stdin(arg: str) -> PathOrSpecial:
     return Path(arg)
 
 
+default_ignore_unused = set(
+    [  # Development tools not meant to be imported
+        "black",
+        "codespell",
+        "colorama",
```
jherland, line 100: As I think was already suggested somewhere else: `colorama` does not belong in this list, as it's not a stand-alone tool (rather a transitive library dependency of many other tools).

tests/test_cmdline.py
```diff
@@ -17,6 +17,7 @@
 from importlib_metadata import files as package_files
 
 from fawltydeps.main import UNUSED_DEPS_OUTPUT_PREFIX, VERBOSE_PROMPT, Analysis, version
+from fawltydeps.settings import default_ignore_unused
```
jherland, line 20:
```suggestion
from fawltydeps.settings import DEFAULT_IGNORE_UNUSED
```

tests/test_cmdline.py
```diff
@@ -41,7 +42,7 @@ def make_json_settings_dict(**kwargs):
         "custom_mapping": None,
         "output_format": "human_summary",
         "ignore_undeclared": [],
-        "ignore_unused": [],
+        "ignore_unused": sorted(list(default_ignore_unused)),
```
jherland, line 45: No need to explicitly call `list()` when `sorted()` will handle the `set` directly (and return a list in either case).

```suggestion
        "ignore_unused": sorted(DEFAULT_IGNORE_UNUSED),
```

tests/test_cmdline.py
```diff
@@ -955,24 +956,31 @@ def test_check_json__no_pyenvs_found__falls_back_to_current_env(fake_project):
...
+            id="check_unused_action_on_default_ignored_unused_dep__outputs_nothing",
+        ),
         pytest.param(
             ["--check-unused", "--ignore-unused", "black", "mypy"],
             ["requests"],
             ["black", "mypy"],
             [Analysis.success_message(check_undeclared=False, check_unused=True)],
-            id="check_unused_action_on_ignored_unused_dep__outputs_nothing",
+            id="check_unused_action_on_overriden_ignored_unused_dep__outputs_nothing",
         ),
```
jherland, line 972: Can we use another module name than `black` or `mypy` here (at least one that is _not_ in `DEFAULT_IGNORE_UNUSED`), to verify that we  are indeed obeying the given `--ignore-unused` option here?

Otherwise a potential bug where `--ignore-unused` no longer overrides the default would not be caught by this test...

tests/test_cmdline.py
```diff
@@ -1089,7 +1097,7 @@ def test_cmdline_on_ignored_undeclared_option(
                 deps = ['foobar']
                 # pyenvs = ['.']
                 # ignore_undeclared = []
-                # ignore_unused = []
+                # ignore_unused = {sorted(list(default_ignore_unused))}
```
jherland, line 1100:
```suggestion
                # ignore_unused = {sorted(DEFAULT_IGNORE_UNUSED)}
```

tests/test_cmdline.py
```diff
@@ -1113,7 +1121,7 @@ def test_cmdline_on_ignored_undeclared_option(
                 # deps = ['.']
                 pyenvs = ['None']
                 # ignore_undeclared = []
-                # ignore_unused = []
+                # ignore_unused = {sorted(list(default_ignore_unused))}
```
jherland, line 1124:
```suggestion
                # ignore_unused = {sorted(DEFAULT_IGNORE_UNUSED)}
```

tests/test_cmdline.py
```diff
@@ -1137,7 +1145,7 @@ def test_cmdline_on_ignored_undeclared_option(
                 # deps = ['.']
                 pyenvs = ['baz', 'xyzzy']
                 # ignore_undeclared = []
-                # ignore_unused = []
+                # ignore_unused = {sorted(list(default_ignore_unused))}
```
jherland, line 1148:
```suggestion
                # ignore_unused = {sorted(DEFAULT_IGNORE_UNUSED)}
```

tests/test_cmdline.py
```diff
@@ -1161,7 +1169,7 @@ def test_cmdline_on_ignored_undeclared_option(
                 # deps = ['.']
                 # pyenvs = ['.']
                 # ignore_undeclared = []
-                # ignore_unused = []
+                # ignore_unused = {sorted(list(default_ignore_unused))}
```
jherland, line 1172:
```suggestion
                # ignore_unused = {sorted(DEFAULT_IGNORE_UNUSED)}
```

tests/test_settings.py
```diff
@@ -27,7 +27,7 @@
 from pydantic.env_settings import SettingsError  # pylint: disable=no-name-in-module
 
 from fawltydeps.main import build_parser
-from fawltydeps.settings import Action, OutputFormat, Settings
+from fawltydeps.settings import Action, OutputFormat, Settings, default_ignore_unused
```
jherland, line 30:
```suggestion
from fawltydeps.settings import Action, OutputFormat, Settings, DEFAULT_IGNORE_UNUSED
```

tests/test_settings.py
```diff
@@ -43,7 +43,7 @@
     custom_mapping=None,
     output_format=OutputFormat.HUMAN_SUMMARY,
     ignore_undeclared=set(),
-    ignore_unused=set(),
+    ignore_unused=default_ignore_unused,
```
jherland, line 46:
```suggestion
    ignore_unused=DEFAULT_IGNORE_UNUSED,
```

jherland (commented):
This looks really good, except for a couple of minor changes outlined below (most of my comments are just copies of one change).

In addition to this, I'd like to see an update to the documentation of the `--ignore-unused` option, to reflect the new default/behavior.

I'm not sure if we should also update the help text:
```
$ fawltydeps --help
...
  --ignore-unused DEP_NAME [DEP_NAME ...]
                        Dependencies to ignore when looking for unused dependencies, e.g. --ignore-unused pylint black
...
```

fawltydeps/settings.py
zz1874: :+1: 
Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/b6afa98689e4720130fad434d0ce13e2d2b44a26.

fawltydeps/settings.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/b6afa98689e4720130fad434d0ce13e2d2b44a26.

tests/test_cmdline.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/b6afa98689e4720130fad434d0ce13e2d2b44a26.

tests/test_cmdline.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/b6afa98689e4720130fad434d0ce13e2d2b44a26.

tests/test_cmdline.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/b6afa98689e4720130fad434d0ce13e2d2b44a26.

tests/test_cmdline.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/b6afa98689e4720130fad434d0ce13e2d2b44a26.

tests/test_cmdline.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/b6afa98689e4720130fad434d0ce13e2d2b44a26.

tests/test_settings.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/b6afa98689e4720130fad434d0ce13e2d2b44a26.

tests/test_settings.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/b6afa98689e4720130fad434d0ce13e2d2b44a26.

fawltydeps/settings.py
zz1874: `colorama` is a direct dependency in our pyproject.toml,
```toml
[tool.poetry.group.format.dependencies]
colorama = "^0.4.6"
...
[tool.poetry.group.dev.dependencies]
colorama = "^0.4.6"
```
In order to make our CI (specially `self_test`) pass, I added it to the list.

Also, this brings me to think, as our list of development dependencies continues to grow, should we consider implementing functionality that allows users to add other tools to the `DEFAULT_IGNORE_UNUSED` list without overwriting it entirely? This could provide more flexibility for users while maintaining our core list of dependencies.
Any ideas? @jherland @Nour-Mws @mknorps

tests/test_cmdline.py
zz1874: Nice catch!
Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/66f403e245037e93478b4e85be5ef90b58f390ba.

fawltydeps/settings.py
Nour-Mws: > colorama is a direct dependency in our pyproject.toml

@jherland it seems it has been added in this commit https://github.com/tweag/FawltyDeps/commit/b86a20eb7956691f6b0851fb494cab39835f4c1a.

> Also, this brings me to think, as our list of development dependencies continues to grow, should we consider implementing functionality that allows users to add other tools to the DEFAULT_IGNORE_UNUSED list without overwriting it entirely? This could provide more flexibility for users while maintaining our core list of dependencies.
Any ideas?

I rather prefer the user takes full responsibility of declaring this list the moment they decide to add a dependency to it. I believe having an in-between state is in the realm of diminishing returns of efforts to functionality.

fawltydeps/settings.py
```diff
@@ -100,21 +100,42 @@ def parse_path_or_stdin(arg: str) -> PathOrSpecial:
...
 DEFAULT_IGNORE_UNUSED = {
     # Development tools not meant to be imported
+    # Formatting Tools
+    "autopep8",
     "black",
+    "isort",
+    "pyformat",
+    "yapf",
+    # Linting Tools
+    "bandit",
```
Nour-Mws, line 110: `bandit` is a security tool

fawltydeps/settings.py
```diff
@@ -100,21 +100,42 @@ def parse_path_or_stdin(arg: str) -> PathOrSpecial:
...
+    "flake8",
+    "mccabe",
+    "pyflakes",
+    "pylint",
+    # Security Tools
```
Nour-Mws, line 115: I'd put `codespell` under formatting as it's not a security tool.

fawltydeps/settings.py
```diff
@@ -100,21 +100,42 @@ def parse_path_or_stdin(arg: str) -> PathOrSpecial:
...
     "tox",
+    # Utility Tools
+    "colorama",
+    "pydocstyle",
+    "pyright",
+    "rope",
+    "ruff",
+    "twine",
+    "unify",
+    "wheel",
```
Nour-Mws, line 138: - I'd put `pyright` and `mypy` under type-checking or linting.
- `ruff` is a linter
- `unify` should go under formatting
- `twine` and `wheel` under building and packaging

fawltydeps/settings.py
```diff
@@ -98,6 +98,47 @@ def parse_path_or_stdin(arg: str) -> PathOrSpecial:
...
+    # Security Tools
+    "codespell",
+    # Documentation Tools
+    "sphinx",
+    "sphinx-rtd-theme",
+    "recommonmark",
+    # Testing Tools
+    "coverage",
+    "fawltydeps",
+    "hypothesis",
```
Nour-Mws, line 124: I'm not sure about including `hypothesis`. Can it be used without being imported in the codE?

fawltydeps/settings.py
jherland: > > colorama is a direct dependency in our pyproject.toml
> 
> @jherland it seems it has been added in this commit [b86a20e](https://github.com/tweag/FawltyDeps/commit/b86a20eb7956691f6b0851fb494cab39835f4c1a).

Ah, I see, `colorama` is an _optional_ dependency of `isort` and `black`: If it is available, it is automatically used to colorize their output, but they otherwise do not depend on it.

This makes it an oddball when considering `DEFAULT_IGNORE_UNUSED`: It is _not_ a tool that is meant to be invoked from the command-line, but there are still arguments for why you'd want to depend on it without (yourself) `import`ing it...

Actually, when I reconsider this now, I think we could have done better in b86a20e: Instead of adding `colorama` as a direct dependency to FawltyDeps, we should make use of the fact that both `black` and `isort` provide optional/extra dependency groups ([here](https://github.com/psf/black/blob/e9356c1ff0083aea4416bf1d3e29748634bb4f7f/pyproject.toml#L76) and [here](https://github.com/PyCQA/isort/blob/e35a55f6ab8473003d68a8b2ffebdd08fbb2b61b/pyproject.toml#L50), respectively) that bring in `colorama` as a transitive dependency.

I think this - combined - with dropping `colorama` from `DEFAULT_IGNORE_UNUSED` is the cleanest solution.

In other words, try something like this diff on our `pyproject.toml`:
```diff
diff --git a/pyproject.toml b/pyproject.toml
index 48d6cd6..bf1450c 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,12 +74,11 @@ types-setuptools = "^65.6.0.2"
 optional = true
 
 [tool.poetry.group.format.dependencies]
-black = "^22"
-colorama = "^0.4.6"
+black = {extras = ["colorama"], version = "^22"}
 isort = [
     # isort 5.12.0 drops support for Python v3.7:
-    {version = "^5.10", python = ">=3.8"},
-    {version = ">=5.10,<5.12.0", python = "<3.8"},
+    {extras = ["colors"], version = "^5.10", python = ">=3.8"},
+    {extras = ["colors"], version = ">=5.10,<5.12.0", python = "<3.8"},
 ]
 codespell = "^2.2.4"
 
@@ -91,9 +90,8 @@ optional = true
 # something to the above groups (i.e. something that targets a specific purpose
 # (e.g. a CI action), consider whether it's also useful to have this available
 # in a developers environment
-black = "^22"
+black = {extras = ["colorama"], version = "^22"}
 codespell = "^2.2.4"
-colorama = "^0.4.6"
 hypothesis = "^6.68.2"
 mypy = "^1.0.1"
 nox = "^2022.11.21"
```

> > Also, this brings me to think, as our list of development dependencies continues to grow, should we consider implementing functionality that allows users to add other tools to the DEFAULT_IGNORE_UNUSED list without overwriting it entirely? This could provide more flexibility for users while maintaining our core list of dependencies.
> > Any ideas?
> 
> I rather prefer the user takes full responsibility of declaring this list the moment they decide to add a dependency to it. I believe having an in-between state is in the realm of diminishing returns of efforts to functionality.

TL;DR: Yes, I think I agree with @Nour-Mws here. This has been discussed somewhat before in https://github.com/tweag/FawltyDeps/issues/328, and I suggest that we continue this part of the discussion there...

fawltydeps/settings.py
jherland: I took a quick look at the introductory docs for Hypothesis, and AFAICS they always need `import hypothesis` somewhere in their examples.

Although there is a `hypothesis` command-line tool that is installed by the package, it seems to be geared towards writing/rewriting test code, and its output (e.g. from running `hypothesis write fawltydeps.utils`) is test code that does indeed `import hypothesis`.

So I think it makes sense to _remove_ `hypothesis` from our `DEFAULT_IGNORE_UNUSED` list.

fawltydeps/settings.py
```diff
@@ -98,6 +98,47 @@ def parse_path_or_stdin(arg: str) -> PathOrSpecial:
...
+    "bandit",
+    "flake8",
+    "mccabe",
+    "pyflakes",
+    "pylint",
+    # Security Tools
+    "codespell",
+    # Documentation Tools
+    "sphinx",
+    "sphinx-rtd-theme",
```
jherland, line 119: Not expected to be imported, but also not a "tool" in its own right (i.e. it does not include a command-line interface). Maybe more of a "tool plugin" in the same category as `pytest-*` packages, I guess?

I'm also guessing there are more sphinx themes than just this one that are in common use?

In general, do we want to consider a more general solution to these categories of plugin packages? Maybe allowing simple globs like `pytest-*` and `sphinx-*` in this set? (although this will add complexity to our matching algorithm in `fawltydeps/check.py`).

But I'm also very happy to leave this as-is for now. We can always reconsider later if/when we see that more of these packages need to be added.

fawltydeps/settings.py
zz1874: Thanks @jherland for digging into this! I agree that making `colorama` a transitive dependency is a neat solution.
It is addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/c263fbdeba33e49d1f5b2c38a82825d4de60d019. 
For the second point, it makes great sense that we wait to see if there is some demand for extending/combining some directives instead always/only replacing. :100:

fawltydeps/settings.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/939043f6fe59516cda60b0d13227d77f50fb114b.

fawltydeps/settings.py
zz1874: Thanks @Nour-Mws for spoting this! 
Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/939043f6fe59516cda60b0d13227d77f50fb114b.

fawltydeps/settings.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/939043f6fe59516cda60b0d13227d77f50fb114b.

fawltydeps/settings.py
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/939043f6fe59516cda60b0d13227d77f50fb114b.

fawltydeps/settings.py
```diff
@@ -98,6 +98,47 @@ def parse_path_or_stdin(arg: str) -> PathOrSpecial:
...
+    "recommonmark",
```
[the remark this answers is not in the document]
zz1874: Addressed in https://github.com/tweag/FawltyDeps/pull/365/commits/939043f6fe59516cda60b0d13227d77f50fb114b.

fawltydeps/settings.py
jherland: Your changes look good to me.

The fact that the discussion on extend/combine vs. replace has come up (at least) twice  in different contexts indicates that there is _some_ demand, but let's track/prioritize that separately in #328. 👍🏻

docs/default_ignore_unused_list.md
```diff
@@ -0,0 +1,51 @@
...
+### Building and Packaging Tools
+
+- twine
+- wheel
+
+### Utility Tools
+
+- pydocstyle
+- rope
+- unify
```
jherland, line 51: I don't like duplicating this list between code and documentation. One will invariably fall out of sync with the other. Can we link directly to the `DEFAULT_IGNORE_UNUSED` in `fawltydeps/settings.py` instead?

fawltydeps/settings.py
zz1874: Agreed. 
I've excluded `sphinx-*` in https://github.com/tweag/FawltyDeps/pull/365/commits/f3408b80582e70cb7ae66e589b9774b12ce62c4b. We may go back and allow `pytest-*` and `sphinx-*` in this set later on.

PR #368, opened 2023-09-18
Fix path read text encode

A-kirami:


jherland (commented):
Thank you very much for your contribution!

I am not sure, however, that it is safe for us to always assume that all Python source files (including `setup.py`) are always UTF-8 encoded?

What other approaches have you attempted before arriving at this solution?

Here are some other ways for you to force Python (and FawltyDeps) to use UTF-8 encoding by default:
- Set `PYTHONUTF8=1` in the environment where FawltyDeps runs (see [here](https://docs.python.org/3.7/using/cmdline.html#envvar-PYTHONUTF8) for more details).
- Pass `-X utf8` to the Python process running FawltyDeps (harder to achieve when running `fawltydeps` on the command line).

If these (or other) options are not suitable to you, we can consider adding a command-line option or configuration setting that prefers UTF-8 over the system locale, but before we go that far, I'd like to learn more about the context for your PR. 😃

A-kirami:
> Thank you very much for your contribution!
> 
> I am not sure, however, that it is safe for us to always assume that all Python source files (including `setup.py`) are always UTF-8 encoded?
> 
> What other approaches have you attempted before arriving at this solution?
> 
> Here are some other ways for you to force Python (and FawltyDeps) to use UTF-8 encoding by default:
> 
> * Set `PYTHONUTF8=1` in the environment where FawltyDeps runs (see [here](https://docs.python.org/3.7/using/cmdline.html#envvar-PYTHONUTF8) for more details).
> * Pass `-X utf8` to the Python process running FawltyDeps (harder to achieve when running `fawltydeps` on the command line).
> 
> If these (or other) options are not suitable to you, we can consider adding a command-line option or configuration setting that prefers UTF-8 over the system locale, but before we go that far, I'd like to learn more about the context for your PR. 😃

I'm using FawltyDeps on Windows, and when I run it, I encounter a Unicode encoding error. Setting `PYTHONUTF8=1` works. Thank you for the suggestion, I'll close this PR.

PR #372, opened 2023-09-26
Relax regex re. recognizing requirements.txt

jherland:
Zhihan found that the plotly project has some requirements files that
are not recognized by FawltyDeps, with names like `requirements_abc.txt`.

Using the underscore as a separator in the filename does not match our
regex, which requires a word boundary (`"\b"`) around the `"requirements"`
string in the filename.

Fix our regex: There is no reason why the underscore separator should
disqualify this as a requirements file, and simplifying the regex to
`".*requirements.*\.(txt|in)"` is unlikely to introduce false positives
(files that should _not_ be recognized as `requirements.txt` files, but
now are).

Also add unit tests precisely for parsing filenames of dependency files
as I could not find another test case that tested exactly this.


Nour-Mws (approved):
Looks good! Hopefully we won't have any false positives popping up in the future :)

PR #383, opened 2023-10-11
Support Python v3.12

jherland:
A fairly straighforward upgrade of our deps to support Python v3.12. No material changes to our code.

Commits:
- `shell.nix`: Upgrade `nixpkgs` to version that has Python v3.12.0
- `poetry.lock`: Refresh with Poetry v1.6.1
- `test_local_env`: Relax test on default venv to accomodate Python v3.12
- Upgrade Pylint for Python versions >=v3.8
- Add Python v3.12 support


shell.nix
```diff
@@ -7,8 +7,8 @@
   python37_overlay ? self: super: { python37 = pkgsWith37.python37; },
   pkgs ? import (builtins.fetchTarball {
     # Branch: nixos-23.05
-    url = "https://github.com/NixOS/nixpkgs/archive/9790f3242da2152d5aa1976e3e4b8b414f4dd206.tar.gz";
-    sha256 = "1y6zipys4803ckvnamfljb8raglgkbz1fz1fg03cxp4jqiiva5s1";
+    url = "https://github.com/NixOS/nixpkgs/archive/1fcb41f796d091e9ebcd90d28c8fd8b5b93231be.tar.gz";
```
Nour-Mws, line 10: How did you determine the commit?

Nour-Mws (approved):
A joy to review as usual :)

shell.nix
jherland: I went to https://github.com/NixOS/nixpkgs and searched (`/`) for `python312` to find where the python312 nix package is defined. That led me here: https://github.com/NixOS/nixpkgs/blob/f6caf65a4a8031f9e1073949734d91b3293034b0/pkgs/development/interpreters/python/default.nix#L93C1-L93C1

I then pressed the "Blame" button to find the commit that actually introduced the 3.12.0 release version (upgrading from the latest RC), which is the commit that introduced the `   suffix = "";` line: https://github.com/NixOS/nixpkgs/commit/1fcb41f796d091e9ebcd90d28c8fd8b5b93231be

Thus we're using the oldest nixpkgs commit that has v3.12.0. I guess I could have chosen any later commit as well.

I do notice, however, that this commit is _not_ present on the `nixos-23.05` branch, which we mention in the comment on the preceding line in our `shell.nix`. I should update that comment as well in this commit...

jherland: Fixed in [21e3b9d](https://github.com/tweag/FawltyDeps/pull/383/commits/21e3b9d0594dd728ec0a574a958bf6c949cdd145)

Issue #374, opened 2023-09-30
decoding error in joblib library


krzysiek98q:
**Describe the bug**
When running FawltyDeps on the joblib library (1.3.2), I get an error.
```python
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/kris/ten_project/.tmp/joblib/joblib/venv/lib/python3.10/site-packages/fawltydeps/__main__.py", line 8, in <module>
    sys.exit(main())
  File "/home/kris/ten_project/.tmp/joblib/joblib/venv/lib/python3.10/site-packages/fawltydeps/main.py", line 362, in main
    analysis = Analysis.create(settings, stdin)
  File "/home/kris/ten_project/.tmp/joblib/joblib/venv/lib/python3.10/site-packages/fawltydeps/main.py", line 188, in create
    ret.undeclared_deps  # pylint: disable=pointless-statement
  File "/home/kris/ten_project/.tmp/joblib/joblib/venv/lib/python3.10/site-packages/fawltydeps/utils.py", line 225, in wrapper
    calculated: T = method(self)
  File "/home/kris/ten_project/.tmp/joblib/joblib/venv/lib/python3.10/site-packages/fawltydeps/main.py", line 156, in undeclared_deps
    return calculate_undeclared(self.imports, self.resolved_deps, self.settings)
  File "/home/kris/ten_project/.tmp/joblib/joblib/venv/lib/python3.10/site-packages/fawltydeps/utils.py", line 225, in wrapper
    calculated: T = method(self)
  File "/home/kris/ten_project/.tmp/joblib/joblib/venv/lib/python3.10/site-packages/fawltydeps/main.py", line 119, in imports
    return list(
  File "/home/kris/ten_project/.tmp/joblib/joblib/venv/lib/python3.10/site-packages/fawltydeps/extract_imports.py", line 201, in parse_sources
    yield from parse_source(source, stdin)
  File "/home/kris/ten_project/.tmp/joblib/joblib/venv/lib/python3.10/site-packages/fawltydeps/extract_imports.py", line 153, in parse_python_file
    path.read_text(), source=Location(path), local_context=local_context
  File "/usr/lib/python3.10/pathlib.py", line 1135, in read_text
    return f.read()
[...]

PR #384, opened 2023-10-12
Handle Python code with embedded encoding declarations

jherland:
As #374 demonstrates, FawltyDeps does not interpret/follow encoding declarations in Python source code.

These encoding declarations have been part of Python for a long time, first introduced in [PEP 263](https://peps.python.org/pep-0263/), and currently documented here: https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations

As far as I can see, these encoding declarations do not really apply to Jupyter notebooks, for the simple reason that notebooks are stored in JSON files, and JSON is specified to always be UTF-8 encdoded. Still we need to support these in three cases:
- for Python source files passed to `extract_imports.parse_python_file()`,
- for Python source code passed directly on FawltyDeps' stdin,
- for `setup.py` files parsed by `extract_declared_dependencies.py.parse_setup_py()`

For the first and last cases, we use the [`tokenize.open()` from the stdlib](https://docs.python.org/3/library/tokenize.html#tokenize.open) to automatically determine the appropriate encoding, and for the second case we use the fact that `ast.parse()` and [the underlying `compile()` built-in function](https://docs.python.org/3/library/functions.html#compile) support passing a un-decoded `bytes` object directly. This involves ensuring that we now read stdin in binary mode instead of text mode, but otherwise the changes introduced in this PR are mostly straightforward.

Commits:
- `sample_projects`: Add example use of legacy encoding
- `tests/utils`: Add `dedent_bytes()`: `textwrap.dedent()` for byte strings
- `extract_imports.parse_code()`: Accept code as `bytes` object
- `extract_imports.parse_python_file()`: Handle legacy encodings
- Handle legacy-encoded Python code passed to stdin
- `extract_declared_dependencies.parse_setup_py()`: Handle legacy encodings

### DIFF
--- a/fawltydeps/extract_declared_dependencies.py
+++ b/fawltydeps/extract_declared_dependencies.py
@@ -5,6 +5,7 @@ import configparser
 import logging
 import re
 import sys
+import tokenize
 from dataclasses import replace
 from pathlib import Path
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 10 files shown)


Nour-Mws (approved):
Thanks for the great reviewing experience Johan! You make it easy to go through the changes and your thought process!

Issue #385, opened 2023-10-31
pre-commit hook doesn't run on just staged files


msabramo:
**Describe the bug**
The pre-commit hook doesn't run on only files that are staged and being committed. The output will flag undeclared dependencies in files that haven't even been `git add`ed.

**To Reproduce**
```shell
git clone https://github.com/msabramo/fawltydeps-demo1
cd fawltydeps-demo1
pre-commit install
cat > main.py
import requests
import pandas
^D
cat > another.py
import sqlite
^D
git add main.py
git status
# Observe that `main.py` is staged and `another.py` is untracked
git commit
```

The output of `git commit` is:
```
$ git commit
FawltyDeps...............................................................Failed
- hook id: check
- exit code: 3

These imports appear to be undeclared dependencies:
- 'pandas' imported at:
    main.py:2
- 'sqlite' imported at:
    another.py:1
```

This is incorrect because it mentions the undeclared dependency in `another.py` which hasn't even been added to the project yet.

**Expected behavior**
It should've only mentioned the undeclared dependency in `main.py`; not in `another.py`, since I'm not committing that. It doesn't even exist in git at all yet, as we've never added it.

**Environment**
 - OS name + version: Mac OS X 12.6.8
 - Version of the code: `main` branch

**Additional context**

msabramo:
#386 is my attempt at fixing this.

jherland:
Thanks for your report! I agree that the pre-commit hook should run FawltyDeps on the staged changes only. We'll look into this as soon as time permits.

PR #386, opened 2023-10-31
Separate `check-undeclared`, `check-unused` pre-commit hooks

msabramo:
The `check-undeclared` hook runs only on files being committed.

Fixes: GH-385

With https://github.com/msabramo/fawltydeps-demo1/blob/GH-385-fix-my-fork/.pre-commit-config.yaml, I get this output:

```
$ cat .pre-commit-config.yaml
repos:
  - repo: https://github.com/msabramo/FawltyDeps
    rev: 4bca742
    hooks:
      - id: check-undeclared
        args:
          - --detailed
          - --code     # --code always has to be last argument, because it must precede list of files
      - id: check-unused
        args:
          - --detailed

$ cat main.py
import requests
import pandas


$ cat another.py
import sqlite


$ cat requirements.txt
requests
pyramid

$ git status
On branch GH-385-fix-my-fork
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	new file:   main.py

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	another.py



$ git commit
FawltyDeps...............................................................Failed
- hook id: check-undeclared
- exit code: 3

These imports appear to be undeclared dependencies:
- 'pandas' imported at:
    main.py:2

FawltyDeps...............................................................Failed
- hook id: check-unused
- exit code: 4

These dependencies appear to be unused (i.e. not imported):
- 'pyramid' declared in:
    requirements.txt
```

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,5 +1,5 @@
 -   id: check-undeclared
-    name: FawltyDeps
+    name: FawltyDeps-undeclared
     description: "FawltyDeps: Check for undeclared dependencies"
     entry: fawltydeps --check-undeclared
     language: python
@@ -8,8 +8,8 @@
       - --code
     pass_filenames: true
 -   id: check-unused
-    name: FawltyDeps
-    description: "FawltyDeps: Check for undeclared dependencies"
+    name: FawltyDeps-unused
+    description: "FawltyDeps: Check for unused dependencies"
     entry: fawltydeps --check-unused
     language: python
     files: ".*\\.py|ipynb"
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,14 +3,13 @@
     description: "FawltyDeps: Check for undeclared dependencies"
     entry: fawltydeps --check-undeclared
     language: python
-    files: ".*\\.py|ipynb"
-    args:
-      - --code
-    pass_filenames: true
+    args: [--code]
+    types_or: [python, jupyter]
+
 -   id: check-unused
     name: FawltyDeps-unused
     description: "FawltyDeps: Check for unused dependencies"
     entry: fawltydeps --check-unused
     language: python
-    files: ".*\\.py|ipynb"
-    pass_filenames: false
+    args: [--deps]
+    files: "(.*requirements.*\\.(txt|in)|pyproject\\.toml|setup\\.(py|cfg))"
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,7 +1,7 @@
 -   id: check-undeclared
     name: FawltyDeps-undeclared
     description: "FawltyDeps: Check for undeclared dependencies"
-    entry: fawltydeps --check-undeclared
+    entry: fawltydeps --check-undeclared --detailed
     language: python
     args: [--code]
     types_or: [python, jupyter]
@@ -9,7 +9,7 @@
 -   id: check-unused
     name: FawltyDeps-unused
     description: "FawltyDeps: Check for unused dependencies"
-    entry: fawltydeps --check-unused
+    entry: fawltydeps --check-unused --detailed
     language: python
     args: [--deps]
     files: "(.*requirements.*\\.(txt|in)|pyproject\\.toml|setup\\.(py|cfg))"
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,6 +1,16 @@
--   id: check
+-   id: check-undeclared
     name: FawltyDeps
-    description: Python dependency checker
-    entry: fawltydeps
+    description: "FawltyDeps: Check for undeclared dependencies"
+    entry: fawltydeps --check-undeclared
     language: python
+    files: ".*\\.py|ipynb"
+    args:
+      - --code
+    pass_filenames: true
+-   id: check-unused
+    name: FawltyDeps
+    description: "FawltyDeps: Check for undeclared dependencies"
+    entry: fawltydeps --check-unused
+    language: python
+    files: ".*\\.py|ipynb"
     pass_filenames: false
--- a/README.md
+++ b/README.md
@@ -600,9 +600,8 @@ tool, you can add something like this to your project's
 ```yaml
 repos:
   - repo: https://github.com/tweag/FawltyDeps
-    rev: v0.13.2
+    rev: v0.13.3
     hooks:
-      - id: check
-        args:
-          - --detailed
+      - id: check-undeclared
+      - id: check-unused
 ```


mknorps (commented):
Thank you for your contribution!

It is a good improvement and works fine. Also, it is a good idea to split undeclared and unused dependencies. It gives more granular visibility of possible problems with the code.
I have some minor comments on naming things, but my main question is: why does it actually work now and did not work before?
I wonder what changed for the pre-commit hook when we pass Python files with  `--code` option instead of taking them "as is".
I did some experiments on the  branch https://github.com/tweag/FawltyDeps/tree/maria/pre-commit-tests, but I do not have the answer to that. Especially, since checking unused dependencies works well without specifying`files: ".*\\.py|ipynb"`.

jherland:
TL;DR: After a few rounds of thinking about this, I think this PR is really close to what we want/need here.

---

Long version:

While trying to understand the proposed solution, I'm learning a lot about pre-commit, and some philosophical differences between pre-commit and FawltyDeps...

At first, I naively thought that we could run FawltyDeps on only the staged _changes_, but this quickly breaks down: Say you have staged a subset of `.py` scripts, and a subset of `requirements.txt` files, then there is no reason for the two sets of files to correspond or even match at all. In a Python script you could have added an import that is already covered by a different (unchanged) deps file, and in a deps file you could have removed a package that was already covered by a different deps file. In other words: it simply does not make sense to run something like `fawltydeps --code <python_files_with_staged_changes> --deps <deps_files_with_staged_changes>`.

In it's default mode of operation (report undeclared and unused deps) it really only makes sense to run FawltyDeps on the _complete state_ of a project (in order to see the full picture of imports vs. deps). And in this case, the state we want to run FawltyDeps on is the _staged_ state (i.e. imagine temporarily removing all unstaged changes and untracked files from the work tree).

This is of course not what `pre-commit` does: Instead it stashes the _unstaged changes_ (i.e. temporarily removes them from the work tree), but the _untracked files_ are not removed before running FawltyDeps. Instead it relies on being able to pass `files` arguments.

Furthermore, having pre-commit run a tool on the _complete state_ of a project sort of goes against pre-commit's philosophy. As [explained here](https://github.com/pre-commit/pre-commit/issues/708#issuecomment-368245812), the point of pre-commit is to run "lightning-fast checks when committing to do spot checks, and leave full testsuites for CI".

Therefore, what we want to check here is:

1. that our Python code changes have not introduced any new undeclared deps, i.e. `--check-undeclared` and `--code <python_files_with_staged_changes>`.
2. that our deps file changes have not introduced any new unused deps, i.e. `--check-unused` and `--deps <deps_files_with_staged_changes>`.

With that in mind, the suggested split of our pre-commit hook into two parts suddenly makes a lot of sense.

Still we'd prefer if the deps files used in point 1 above reflected the staged state (i.e. no untracked files), and, conversely, that the code files in point 2 above reflected the staged state. I don't think this can be easily solved as things currently stand, but I think even without that, this PR represents a welcome improvement.

.pre-commit-hooks.yaml
```diff
@@ -1,6 +1,16 @@
...
-    entry: fawltydeps
+-   id: check-undeclared
+    name: FawltyDeps-undeclared
+    description: "FawltyDeps: Check for undeclared dependencies"
+    entry: fawltydeps --check-undeclared
     language: python
+    files: ".*\\.py|ipynb"
+    args:
+      - --code
+    pass_filenames: true
```
jherland, line 9: - Instead of `files`, I think we can use the `types_or` directive which yields a more readable result. At least this seems to work in my local testing.
- For `args`, I prefer the single-line style when there's only one item
- And I think we can skip `pass_filename` since its default value is already `true`
- Finally, a blank line between the two hooks/checks makes the YAML a bit easier to read, IMHO.

```suggestion
    args: [--code]
    types_or: [python, jupyter]

```

.pre-commit-hooks.yaml
```diff
@@ -1,6 +1,16 @@
...
+-   id: check-unused
+    name: FawltyDeps-unused
+    description: "FawltyDeps: Check for unused dependencies"
+    entry: fawltydeps --check-unused
+    language: python
+    files: ".*\\.py|ipynb"
     pass_filenames: false
```
jherland, line 16: Here, I think we want to do the converse of the above: Pass `--deps` plus the dependency declaration files that have staged changed. Something like this:

```suggestion
    args: [--deps]
    files: "(.*requirements.*\\.(txt|in)|pyproject\\.toml|setup\\.(py|cfg))"
```

jherland (changes requested):
Added some suggestions based on the thoughts I posted earlier, along with some cursory local testing. With those applied, I think this is ready to 🚀

Thanks again for the contribution!

jherland:
It also occurs to me that we should update the `README.md` accordingly in this PR:
```diff
diff --git a/README.md b/README.md
index 69334ec..d819ddc 100644
--- a/README.md
+++ b/README.md
@@ -646,9 +646,8 @@ tool, you can add something like this to your project's
 ```yaml
 repos:
   - repo: https://github.com/tweag/FawltyDeps
-    rev: v0.13.2
+    rev: v0.13.3
     hooks:
-      - id: check
-        args:
-          - --detailed
+      - id: check-undeclared
+      - id: check-unused
```

mknorps:
Side comment.

> Furthermore, having pre-commit run a tool on the complete state of a project sort of goes against pre-commit's philosophy. As https://github.com/pre-commit/pre-commit/issues/708#issuecomment-368245812, the point of pre-commit is to run "lightning-fast checks when committing to do spot checks, and leave full testsuites for CI".

Another tool that uses the whole state and not incremental updates is [mypy](https://gdevops.gitlab.io/tuto_git/tools/pre-commit/repos_hooks/mypy/mypy.html). They use a mypy mirror, but I wonder if they would not encounter the same problem with untracked files.

jherland:
> Another tool that uses the whole state and not incremental updates is [mypy](https://gdevops.gitlab.io/tuto_git/tools/pre-commit/repos_hooks/mypy/mypy.html). They use a mypy mirror, but I wonder if they would not encounter the same problem with untracked files.

Good point. A way to test this would be to stage a change that imports a new module `import untracked_file`, and then have some code with type errors in `untracked_file.py`, then run the mypy pre-commit hook and see whether it looks at `untracked_file.py` or not...

.pre-commit-hooks.yaml
```diff
@@ -1,6 +1,16 @@
...
+    files: ".*\\.py|ipynb"
+    args:
+      - --code
+    pass_filenames: true
+-   id: check-unused
+    name: FawltyDeps
+    description: "FawltyDeps: Check for undeclared dependencies"
+    entry: fawltydeps --check-unused
+    language: python
+    files: ".*\\.py|ipynb"
```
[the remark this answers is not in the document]
jherland: (This suggestion was superseded by later suggestion.)

jherland (approved):
Pushed my suggested updates, and with that I'm happy to approve and merge this. 

Thanks again for raising the initial issue and PR! 🙏🏻

msabramo:
Thanks for updating my PR and merging. I'm glad it was useful!

PR #387, opened 2023-10-31
Preliminary refactoring before `--ignore` feature

jherland:
Preliminary changes to clean up the code in preparation for work on supporting gitignore-style patterns for ignoring parts of the project during traversal.

Commits:
- `tests/utils.py`: Better output when `assert_unordered_equivalence` fails
- `DirectoryTraversal`: Allow attaching multiple data items in `.add()`
- Move `DirectoryTraversal` into its own module
- `dir_traversal`: Rename `.ignore()` to `.skip_dir()`
- `dir_traversal`: Refactor data on each traversal step into its own class
- `test_dir_traversal`: Make use of `ExpectedTraverseStep` self-documenting

### DIFF
--- a/fawltydeps/traverse_project.py
+++ b/fawltydeps/traverse_project.py
@@ -63,8 +63,8 @@ def find_sources(  # pylint: disable=too-many-branches,too-many-statements
         else:  # must traverse directory
             # sanity check: convince mypy that SpecialPath is already handled
             assert isinstance(path_or_special, Path)
-            traversal.add(path_or_special, CodeSource)
-            traversal.add(path_or_special, path_or_special)  # also record base dir
+            # record also base dir for later
+            traversal.add(path_or_special, CodeSource, path_or_special)
 
     for path in settings.deps if DepsSource in source_types else []:
         # exceptions raised by validate_deps_source() are propagated here
--- a/fawltydeps/traverse_project.py
+++ b/fawltydeps/traverse_project.py
@@ -1,7 +1,7 @@
 """Traverse a project to identify appropriate inputs to FawltyDeps."""
 import logging
 from pathlib import Path
-from typing import AbstractSet, Iterator, Optional, Set, Type, Union
+from typing import AbstractSet, Iterator, Optional, Set, Tuple, Type, Union
 
 from fawltydeps.dir_traversal import DirectoryTraversal
 from fawltydeps.extract_declared_dependencies import validate_deps_source
@@ -17,6 +17,16 @@ from fawltydeps.types import (
 )
 
 logger = logging.getLogger(__name__)
+
+
+# When setting up the traversal, we .add() directories to be traverse and attach
+# information about what we're looking for during the traversal. These are the
+# types of data we're allowed to attach:
+AttachedData = Union[
+    Tuple[Type[CodeSource], Path],  # Look for Python code, with a base_dir
+    Type[DepsSource],  # Look for files with dependency declarations
+    Type[PyEnvSource],  # Look for Python environments
+]
 
 
 def find_sources(  # pylint: disable=too-many-branches,too-many-statements
@@ -52,7 +62,7 @@ def find_sources(  # pylint: disable=too-many-branches,too-many-statements
     logger.debug(f"    deps:   {settings.deps}")
     logger.debug(f"    pyenvs: {settings.pyenvs}")
 
-    traversal: DirectoryTraversal[Union[Type[Source], Path]] = DirectoryTraversal()
+    traversal: DirectoryTraversal[AttachedData] = DirectoryTraversal()
 
     for path_or_special in settings.code if CodeSource in source_types else []:
         # exceptions raised by validate_code_source() are propagated here
@@ -64,7 +74,7 @@ def find_sources(  # pylint: disable=too-many-branches,too-many-statements
             # sanity check: convince mypy that SpecialPath is already handled
             assert isinstance(path_or_special, Path)
             # record also base dir for later
-            traversal.add(path_or_special, CodeSource, path_or_special)
+            traversal.add(path_or_special, (CodeSource, path_or_special))
 
     for path in settings.deps if DepsSource in source_types else []:
         # exceptions raised by validate_deps_source() are propagated here
@@ -92,8 +102,16 @@ def find_sources(  # pylint: disable=too-many-branches,too-many-statements
             if subdir.name.startswith("."):
                 traversal.skip_dir(subdir)
 
-        types = {t for t in step.attached if t in source_types}
+        # Extract the Source types we're looking for in this directory.
+        # Sanity checks:
+        #   - We should not traverse into a directory unless we're looking for
+        #     at least _one_ source type.
+        #   - We should not be looking for any _other_ source types than those
+        #     that were given in our `source_types` argument.
+        types = {t[0] if isinstance(t, tuple) else t for t in step.attached}
         assert len(types) > 0
+        assert all(t in source_types for t in types)
+
         if PyEnvSource in types:
             for path in step.subdirs:
                 package_dirs = validate_pyenv_source(path)
@@ -101,9 +119,10 @@ def find_sources(  # pylint: disable=too-many-branches,too-many-statements
                     yield from package_dirs
                     traversal.skip_dir(path)  # don't recurse into Python environment
         if CodeSource in types:
-            # Retrieve base_dir from closest ancestor, i.e. last Path in attached
+            # Retrieve base_dir from closest ancestor, i.e. last CodeSource in .attached:
             base_dir = next(
-                (x for x in reversed(step.attached) if isinstance(x, Path)), None
+                (t[1] for t in reversed(step.attached) if isinstance(t, tuple)),
+                None,
             )
             assert base_dir is not None  # sanity check: No CodeSource w/o base_dir
             for path in step.files:
--- a/fawltydeps/traverse_project.py
+++ b/fawltydeps/traverse_project.py
@@ -3,6 +3,7 @@ import logging
 from pathlib import Path
 from typing import AbstractSet, Iterator, Optional, Set, Type, Union
 
+from fawltydeps.dir_traversal import DirectoryTraversal
 from fawltydeps.extract_declared_dependencies import validate_deps_source
 from fawltydeps.extract_imports import validate_code_source
 from fawltydeps.packages import validate_pyenv_source
@@ -14,7 +15,6 @@ from fawltydeps.types import (
     Source,
     UnparseablePathException,
 )
-from fawltydeps.utils import DirectoryTraversal
 
 logger = logging.getLogger(__name__)
 
--- a/fawltydeps/traverse_project.py
+++ b/fawltydeps/traverse_project.py
@@ -83,14 +83,14 @@ def find_sources(  # pylint: disable=too-many-branches,too-many-statements
         if package_dirs is not None:  # Python environment dir given directly
             logger.debug(f"find_sources() Found {package_dirs}")
             yield from package_dirs
-            traversal.ignore(path)  # disable traversal of path below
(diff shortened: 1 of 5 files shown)


fawltydeps/traverse_project.py
```diff
@@ -63,8 +63,8 @@ def find_sources(  # pylint: disable=too-many-branches,too-many-statements
         else:  # must traverse directory
             # sanity check: convince mypy that SpecialPath is already handled
             assert isinstance(path_or_special, Path)
-            traversal.add(path_or_special, CodeSource)
-            traversal.add(path_or_special, path_or_special)  # also record base dir
+            # record also base dir for later
+            traversal.add(path_or_special, CodeSource, path_or_special)
```
mknorps, line 67: Could you give me more context on this line?
It looks like `attach_data` argument of `add` method may have varying type - either a `Source` descendant or a `PathOrSpecial`. Could we distinguish clearer between paths we are traversing and the data we are attaching to them (in my understanding it would be some kind of `Source`)? Maybe splitting `attached` to the part responsible for paths and other, responsible for actual types of `Sources` we parse from the path?

mknorps (approved):
The refactor looks good to me :rocket: 

I have some question for traversal implementation, but they are not blocking from merging this PR :)

tests/test_dir_traversal.py
```diff
@@ -95,8 +95,8 @@ class DirectoryTraversalVector(Generic[T]):
 
     id: str
     given: List[BaseEntry]
-    add: List[Tuple[str, Optional[T]]] = field(default_factory=lambda: [(".", None)])
-    ignore: List[str] = field(default_factory=list)
+    add: List[Tuple[str, Tuple[T, ...]]] = field(default_factory=lambda: [(".", ())])
```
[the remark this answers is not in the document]
jherland: Agree that these are not sufficiently (self-)documented. What we're setting up here is a list of calls to `traversal.add(...)`. Each item corresponds to one `.add()` call, and consists of:
- a path, relative to `tmp_path` (in the test below, we do `tmp_path / path` to turn it into an absolute path, and that allows us to use a `str` here, which cuts down on boilerplate, i.e. not having to wrap every path in `Path(...)`)
- a collection of arbitrary data to _attach_. This can be _any_ type, and the contract is that while traversing underneath the added path, this data will be provided back to the caller via the `.attached` member on the `TraversalStep` object.

I am thinking of adding a `NamedTuple` like this:

```diff
+class AddCall(NamedTuple):
+    path: Union[Path, str]
+    attach: Tuple[T, ...] = ()
+
+
 @dataclass
 class DirectoryTraversalVector(Generic[T]):
     """Test vectors for DirectoryTraversal."""
 
     id: str
     given: List[BaseEntry]
-    add: List[Tuple[str, Tuple[T, ...]]] = field(default_factory=lambda: [(".", ())])
+    add: List[AddCall] = field(default_factory=lambda: [AddCall(".")])
     skip_dirs: List[str] = field(default_factory=list)
     expect: List[ExpectedTraverseStep] = field(default_factory=list)
     expect_alternatives: Optional[List[List[ExpectedTraverseStep]]] = None
```

And each vector would then be changed like this:
```diff
     DirectoryTraversalVector(
         "one_subdir_plus__attach_data_on_both",
         given=[Dir("sub")],
-        add=[(".", (123,)), ("sub", (456,))],
+        add=[AddCall(".", (123,)), AddCall("sub", (456,))],
         expect=[
             ExpectedTraverseStep(".", subdirs=["sub"], attached=[123]),
             ExpectedTraverseStep("sub", attached=[123, 456]),
         ],
     ),
```

or even more verbose:
```diff
     DirectoryTraversalVector(
         "one_subdir_plus__attach_data_on_both",
         given=[Dir("sub")],
-        add=[(".", (123,)), ("sub", (456,))],
+        add=[
+            AddCall(path=".", attach=(123,)),
+            AddCall(path="sub", attach=(456,)),
+        ],
         expect=[
             ExpectedTraverseStep(".", subdirs=["sub"], attached=[123]),
             ExpectedTraverseStep("sub", attached=[123, 456]),
         ],
     ),
```

What do you prefer? Is this sufficient, or should we do something more/different?

fawltydeps/traverse_project.py
jherland: Yeah, so _ideally_ what I wanted to add here was just the `Source` type. In that sense I would read this:

```python
        traversal.add(path_or_special, CodeSource)
```

as "look for code sources beneath `path_or_special`".

However, when we go looking for code sources, we have an extra requirement: In order to _create_ a `CodeSource` object we need to pass in an extra `base_dir` argument. (This is needed to correctly resolve 1st- vs. 3rd-party imports in the parsed code).

So this `base_dir` is another piece of data that needs to be attached to the traversal. I thought about other ways to store/handle this information throughout the traversal, but it is fundamentally linked to which directory we're currently traversing, so in the end I found this was the best place to put it.

So the items we are attaching to an added directory are one or more of these:
- for code sources: `CodeSource` and a corresponding `base_dir`,
- for deps sources: `DepsSource`,
- for pyenv sources: `PyEnvSource`.

In summary, the attached data items that we use in `traverse_project.py` are then of type `Union[Type[Source], Path]`, as can be seen near the start of `find_sources()`:

```python
    traversal: DirectoryTraversal[Union[Type[Source], Path]] = DirectoryTraversal()
```

(so Mypy will catch if we ever try to attach something else than what is described above).

At the point where we are stepping through the traversal and extracting the attached data, we pull apart the various types again, albeit this part of the code is not very elegant:

```python
    for step in traversal.traverse():
        ...
        types = {t for t in step.attached if t in source_types}
        assert len(types) > 0
        if PyEnvSource in types:
            ...
        if CodeSource in types:
            # Retrieve base_dir from closest ancestor, i.e. last Path in attached
            base_dir = next(
                (x for x in reversed(step.attached) if isinstance(x, Path)), None
            )
            assert base_dir is not None  # sanity check: No CodeSource w/o base_dir
            ...
        if DepsSource in types:
            ...
```

tests/test_dir_traversal.py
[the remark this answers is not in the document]
mknorps: I think the last option is the most readable. Indeed, it adds verbosity to the test examples, but on the other hand one knows what is being tested at a glance.

fawltydeps/traverse_project.py
mknorps: > So the items we are attaching to an added directory are one or more of these:
> 
>     for code sources: `CodeSource` and a corresponding `base_dir`,
>     for deps sources: `DepsSource`,
>     for pyenv sources: `PyEnvSource`.

Could we use `base_dir` property of `CodeSource` object instead and not attach another `base_dir` outside of this object?

That would transform `traversal: DirectoryTraversal[Union[Type[Source], Path]] ` into `traversal: DirectoryTraversal[Type[Source]] ` and the code will be simpler.

jherland: At the point we are attaching this data, there _is_ no `CodeSource` object yet. We are attaching the `CodeSource` _type_ as an indicator that we want `CodeSource` objects to be created for python code found under the given path.

I was considering a trick where we use `functools.partial` to "halfway construct" a `CodeSource` object:

```python
    halfway = functools.partial(CodeSource, base_dir=path_or_special)
    traversal.add(path_or_special, halfway)
```

However, the type of `halfway` is _no longer_ `CodeSource` (but rather a `functools.partial` object representing the deferred call to the `CodeSource` constructor), so this is possibly even more complex/less understandable...

mknorps: > At the point we are attaching this data, there is no `CodeSource` object yet.

I see. Now that I better understand it I can summarize my concerns :) I think this is a nitpicking discussion, the code works perfectly well. From my perspective the code would be clearer if in `DirectoryTraversal` function `add` would have following signature:
```
    def add(self, dir_path: Path, *attach_data: Tuple[T,Any]) -> None:
```
or something like
```
    def add(self, dir_path: Path, sources: List[T], attach_data: Dict[T,Any]) -> None:
```
btw, current signature also does not always work, as we have ` *attach_data: T`, but we put there `CodeSource, path_or_special` of which latter is an instance of `PathOrSpecial` and not a type variable.


Back to my point. We know that to add an entry to traverse, we need to have three things (in general):
- path to the directory/file
- type of Source
- additional data needed to instantiate this source (like base path)

For me it would be clearer if we make those three arguments separate and visible in the `add` function. In the current formulation, we could as well do:
```
traversal.add(path_or_special, CodeSource, path_or_special, DepsSource, "42")
```
as `attach_data` becomes a bucket to throw "all other" data. But we know the structure of this "other data" so why not use it.

jherland: Thanks for following up on this. I agree that it would be good to have a stronger type for the attached data in this case. `traverse_project` should more strictly specify what type of data we attach here.

However, I don't want to push this type into the `DirectoryTraversal` API. I like the idea of `DirectoryTraversal` providing a general "bucket" where users can attach whatever data they need to retrieve during the traversal, and not to impose assumptions/requirements on the shape of this data. It is then up to the user (`traverse_project` in this case) to specify the `T` type variable as strongly as possible for their case, so that there is no ambiguity.

So realizing that I was a little too lazy when I wrote:
```python
    traversal: DirectoryTraversal[Union[Type[Source], Path]] = DirectoryTraversal()
```

what about this instead?:

```python
# When setting up the traversal, we .add() directories to be traverse and attach
# information about what we're looking for during the traversal. These are the
# types of data we're allowed to attach:
AttachedData = Union[
    Tuple[Type[CodeSource], Path],  # Look for Python code, with a base_dir
    Type[DepsSource],  # Look for files with dependency declarations
    Type[PyEnvSource],  # Look for Python environments
]


def find_sources(...):
    ...
    traversal: DirectoryTraversal[AttachedData] = DirectoryTraversal()
    ...
    # replace traversal.add(path_or_special, CodeSource, path_or_special)
    # with    traversal.add(path_or_special, (CodeSource, path_or_special))
    ...
    for step in traversal.traverse():
        ...
        # Extract the Source types we're looking for in this directory.
        # Sanity checks:
        #   - We should not traverse into a directory unless we're looking for
        #     at least _one_ source type.
        #   - We should not be looking for any _other_ source types than those
        #     that were given in our `source_types` argument.
        types = {
            t[0] if isinstance(t, tuple) else t
            for t in step.attached
        }
        assert len(types) > 0
        assert all(t in source_types for t in types)

        ...
        if CodeSource in types:
            # Retrieve base_dir from closest ancestor, i.e. last CodeSource in .attached:
            base_dir = next(
                (t[1] for t in reversed(step.attached) if isinstance(t, tuple)),
                None,
            )
            assert base_dir is not None  # sanity check: No CodeSource w/o base_dir
        ...
```

tests/test_dir_traversal.py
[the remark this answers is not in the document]
jherland: Fixed in [13f9373](https://github.com/tweag/FawltyDeps/pull/387/commits/13f9373ef6b2f90c610c38e409e9314611baca36)

fawltydeps/traverse_project.py
jherland: Suggestion pushed in [2bf2a8c](https://github.com/tweag/FawltyDeps/pull/387/commits/2bf2a8c73f59f80c3df253c6f4ab07b13ca7b612).

mknorps: > I like the idea of DirectoryTraversal providing a general "bucket" where users can attach whatever data they need to retrieve during the traversal, and not to impose assumptions/requirements on the shape of this data. It is then up to the user (traverse_project in this case) to specify the T type variable as strongly as possible for their case, so that there is no ambiguity.

I am not a big fan of the "bucket" idea, but I see how it can be extended easier than the overspecified types in which we may be lost ;)

I like very much `AttachedData` type, now it is easily seen what `find_sources` needs to work.

Thank you @jherland :rocket:

PR #388, opened 2023-10-31
Support exclude patterns

jherland:
Support gitignore-style patterns given directly in the configuration or on the command line. No support for indirectly reading ignore patterns from files, yet.

See #391 for a follow-up PR that vendors/integrates the `gitignore_parser` dependency instead of depending on it, as we cannot really use its "public" API as intended.

Commits:
- `test_dir_traversal`: Add `.skip_me` member to each `DirectoryTraversalVector`
- Add `gitignore_parser` as a new dependency
- `DirectoryTraversal`: Initial support for exclude patterns
- `dir_traversal`: Add `excluded_*` fields to `TraversalStep`
- `dir_traversal`: Do not exclude an explicitly `.add()`ed path
- `traverse_project`: Use new `exclude` feature to exclude "hidden" paths
- `Settings`: Add `.exclude`, paths to exclude during project traversal
- `sample_projects`: Add support for exclude patterns
- Add `--exclude` option to CLI
- Add `ExcludePattern` test helper

---

(Old description before rename from "ignore" to "exclude":)

Support gitignore-style patterns given directly in the configuration or on the command line. No support for _files_ with ignore patterns yet.

Also, contemplating whether to vendor/integrate the `gitignore_parser` dependency instead of depending on it, as we cannot really use its "public" API as intended.

Commits:
- Add `gitignore_parser` as a new dependency
- `DirectoryTraversal`: Initial support for ignore patterns
- `dir_traversal`: Add `ignored_*` fields to `TraversalStep`
- `dir_traversal`: Do not ignore an explicitly `.add()`ed path
- `traverse_project`: Use new ignore feature to ignore "hidden" paths
- `Settings`: Add `.ignores`, paths to ignore during project traversal
- `sample_projects`: Add support for ignores
- Add `--ignore` option to CLI

### DIFF
--- a/README.md
+++ b/README.md
@@ -386,6 +386,9 @@ Here is a complete list of configuration directives we support:
 - `install-deps`: Automatically install Python dependencies gathered with
   FawltyDeps into a temporary virtual environment. This will use `pip install`,
   which downloads packages from PyPI by default.
+- `exclude`: File/directory patterns to exclude/ignore when looking for code
+  (imports), dependency declarations and/or Python environments. Defaults to
+  `exclude = [".*"]`, meaning that hidden/dot paths are excluded from traversal.
 - `verbosity`: An integer controlling the default log level of FawltyDeps:
   - `-2`: Only `CRITICAL`-level log messages are shown.
   - `-1`: `ERROR`-level log messages and above are shown.
--- a/README.md
+++ b/README.md
@@ -298,6 +298,51 @@ above. For example, if the mapping file has some stale/incorrect mapping
 entries, they will _not_ be resolved by the Python environment resolver (which
 is usually more accurate).
 
+### Excluding paths
+
+If you want FawltyDeps to exclude parts of your source tree when loooking for
+code, dependency declarations, or Python environments, then you can use the
+`--exclude` option to specify path patterns to exclude, e.g. the following
+command will skip everything under `tests/`:
+
+```sh
+fawltydeps --exclude tests/
+```
+
+The format of the exclude patterns is the same as used by `.gitignore` files,
+[see here for a full description](https://git-scm.com/docs/gitignore#_pattern_format).
+(Note that some patterns that must be interpreted relative to the location of a
+`.gitignore` file are not (yet) supported by FawltyDeps. Search for "relative"
+in the linked page to see which pattern types are not yet supported.)
+
+When the `--exclude` option is not specified, its default value is `".*"`, which
+matches all paths that start with a dot (`.`), aka. "hidden" paths. In the above
+example, if you want to exclude both hidden paths, and everything under
+`tests/`, then instead use:
+
+```sh
+fawltydeps --exclude tests/ ".*"
+```
+
+The `--exclude` patterns have lower priority than any paths you pass directly
+on the command line, e.g. in this command:
+
+```sh
+fawltydeps --code my_file.py --exclude my_file.py
+```
+
+the `--code` options "wins" (i.e. imports in `my_file.py` will be found); the
+`--exclude` option only takes affect when traversing directories to look for
+more files. E.g. use this to find code inside `my_dir`, but skip Jupyter
+notebooks:
+
+```sh
+fawltydeps --code my_dir --exclude "*.ipynb"
+```
+
+(The extra quotes here are needed to prevent the shell from interpreting and
+replacing the `*` wildcard.)
+
 ### Ignoring irrelevant results
 
 There may be `import` statements in your code that should not be considered an
--- a/fawltydeps/settings.py
+++ b/fawltydeps/settings.py
@@ -162,6 +162,7 @@ class Settings(BaseSettings):
     ignore_unused: Set[str] = DEFAULT_IGNORE_UNUSED
     deps_parser_choice: Optional[ParserChoice] = None
     install_deps: bool = False
+    exclude: Set[str] = {".*"}
     verbosity: int = 0
     custom_mapping_file: Set[Path] = set()
 
--- a/tests/sample_projects/hidden_files/expected.toml
+++ b/tests/sample_projects/hidden_files/expected.toml
@@ -8,7 +8,7 @@ description = """
 """
 
 [experiments.default]
-description = "Run fawltydeps in an project where everything is hidden."
+description = "Default run where everything is hidden."
 pyenvs = [""]  # Find all Python environments in project
 
 # 3rd-party imports found in the code:
@@ -22,3 +22,57 @@ undeclared_deps = []
 
 # Declared dependencies which were never `import`ed from the code:
 unused_deps = []
+
+[experiments.exclude_nothing]
+description = "Run without excludes - everything is found."
+pyenvs = [""]  # Find all Python environments in project
+exclude = []  # No default excludes
+
+# 3rd-party imports found in the code:
+imports = [
+    "module_a",  # .hidden.code.py
+    "module_b",  # .hidden_dir/code.py
+]
+
+# Declared dependencies found in the project configuration:
+declared_deps = [
+    "package_x",  # .hidden.requirements.txt
+    "package_y",  # .hidden_dir/requirements.txt
+]
+
+# Import names in the code that do not have a matching dependency declared:
+undeclared_deps = [
+    "module_a",  # .hidden.code.py
+    "module_b",  # .hidden_dir/code.py
+]
+
+# Declared dependencies which were never `import`ed from the code:
+unused_deps = [
+    "package_x",  # .hidden.requirements.txt
+    "package_y",  # .hidden_dir/requirements.txt
+]
+
+[experiments.exclude_some_things]
+description = "Run with custom exclude pattern instead of default."
+pyenvs = [""]  # Find all Python environments in project
+exclude = [".hidden_dir/"]  # Only exclude .hidden_dir/
+
+# 3rd-party imports found in the code:
+imports = [
+    "module_a",  # .hidden.code.py
+]
+
+# Declared dependencies found in the project configuration:
+declared_deps = [
+    "package_x",  # .hidden.requirements.txt
+]
+
+# Import names in the code that do not have a matching dependency declared:
+undeclared_deps = [
+    "module_a",  # .hidden.code.py
+]
+
+# Declared dependencies which were never `import`ed from the code:
+unused_deps = [
+    "package_x",  # .hidden.requirements.txt
+]
--- a/tests/sample_projects/hidden_files/expected.toml
+++ b/tests/sample_projects/hidden_files/expected.toml
@@ -1,0 +1,24 @@
+[project]
+# General information about a simplified project: Its name, why we test it,
+# its relation to real world projects
+name = "hidden_files"
+description = """
+    A project with code/deps in hidden files and inside hidden dirs, as well as
+    pyenvs that are inside hidden dirs (one of them being a hidden dir itself)
+"""
+
+[experiments.default]
+description = "Run fawltydeps in an project where everything is hidden."
+pyenvs = [""]  # Find all Python environments in project
+
+# 3rd-party imports found in the code:
+imports = []
+
+# Declared dependencies found in the project configuration:
+declared_deps = []
+
+# Import names in the code that do not have a matching dependency declared:
+undeclared_deps = []
+
+# Declared dependencies which were never `import`ed from the code:
+unused_deps = []
--- a/tests/test_dir_traversal.py
+++ b/tests/test_dir_traversal.py
@@ -108,6 +108,12 @@ class DirectoryTraversalVector(Generic[T]):
     skip_dirs: List[str] = field(default_factory=list)
     expect: List[ExpectedTraverseStep] = field(default_factory=list)
     expect_alternatives: Optional[List[List[ExpectedTraverseStep]]] = None
+    skip_me: Optional[str] = None
+
+
+def on_windows(msg: str) -> Optional[str]:
+    """Helper used by .skip_me to skip certain tests on Windows."""
+    return msg if sys.platform.startswith("win") else None
 
 
 directory_traversal_vectors: List[DirectoryTraversalVector] = [
@@ -178,6 +184,7 @@ directory_traversal_vectors: List[DirectoryTraversalVector] = [
         "symlinks_to_self__are_not_traversed",
         given=[RelativeSymlink("rel_self", "."), AbsoluteSymlink("abs_self", ".")],
         expect=[ExpectedTraverseStep(".", subdirs=["rel_self", "abs_self"])],
+        skip_me=on_windows("Symlinks on Windows may be created only by administrators"),
     ),
     DirectoryTraversalVector(
         "symlinks_to_parent__are_not_traversed",
@@ -189,6 +196,7 @@ directory_traversal_vectors: List[DirectoryTraversalVector] = [
             ExpectedTraverseStep(".", subdirs=["sub"]),
             ExpectedTraverseStep("sub", subdirs=["rel_parent", "abs_parent"]),
         ],
+        skip_me=on_windows("Symlinks on Windows may be created only by administrators"),
     ),
     DirectoryTraversalVector(
         "mutual_symlinks__are_traversed_once",
@@ -219,6 +227,7 @@ directory_traversal_vectors: List[DirectoryTraversalVector] = [
                 ),
             ],
         ],
+        skip_me=on_windows("Symlinks on Windows may be created only by administrators"),
     ),
     DirectoryTraversalVector(
         "relative_symlink_to_dir_elsewhere__is_traversed",
@@ -231,6 +240,7 @@ directory_traversal_vectors: List[DirectoryTraversalVector] = [
             ExpectedTraverseStep("here", subdirs=["symlink"]),
             ExpectedTraverseStep(str(Path("here", "symlink")), files=["file"]),
         ],
+        skip_me=on_windows("Symlinks on Windows may be created only by administrators"),
     ),
     DirectoryTraversalVector(
         "absolute_symlink_to_dir_elsewhere__is_traversed",
@@ -243,6 +253,7 @@ directory_traversal_vectors: List[DirectoryTraversalVector] = [
             ExpectedTraverseStep("here", subdirs=["symlink"]),
             ExpectedTraverseStep(str(Path("here", "symlink")), files=["file"]),
         ],
+        skip_me=on_windows("Symlinks on Windows may be created only by administrators"),
     ),
     DirectoryTraversalVector(
         "dir_and_symlinks_to_dir__is_traversed_only_once",
@@ -265,29 +276,18 @@ directory_traversal_vectors: List[DirectoryTraversalVector] = [
                 ExpectedTraverseStep("abs_link", files=["file"]),
             ],
         ],
+        skip_me=on_windows("Symlinks on Windows may be created only by administrators"),
     ),
 ]
 
 
 @pytest.mark.parametrize(
-    "vector",
-    [
-        pytest.param(
-            v,
-            id=v.id,
-            marks=pytest.mark.skipif(
-                sys.platform.startswith("win")
-                and any(
-                    isinstance(entry, (RelativeSymlink, AbsoluteSymlink))
-                    for entry in v.given
-                ),
-                reason="Symlinks on Windows may be created only by administrators",
-            ),
-        )
-        for v in directory_traversal_vectors
-    ],
+    "vector", [pytest.param(v, id=v.id) for v in directory_traversal_vectors]
 )
 def test_DirectoryTraversal(vector: DirectoryTraversalVector, tmp_path):
+    if vector.skip_me is not None:
+        pytest.skip(vector.skip_me)
+
     for entry in vector.given:
         entry(tmp_path)
 
--- a/tests/test_dir_traversal.py
+++ b/tests/test_dir_traversal.py
@@ -3,11 +3,16 @@ import sys
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Generic, List, Optional, Tuple, TypeVar, Union
+from typing import Generic, List, Optional, Tuple, Type, TypeVar, Union
 
 import pytest
 
-from fawltydeps.dir_traversal import DirectoryTraversal, TraversalStep
+from fawltydeps.dir_traversal import (
+    DirectoryTraversal,
+    ExcludeRuleError,
+    ExcludeRuleMissing,
+    TraversalStep,
+)
 
 from .utils import assert_unordered_equivalence
 
@@ -106,6 +111,10 @@ class DirectoryTraversalVector(Generic[T]):
     given: List[BaseEntry]
     add: List[AddCall] = field(default_factory=lambda: [AddCall(path=".")])
     skip_dirs: List[str] = field(default_factory=list)
+    exclude_patterns: List[Union[str, Tuple[str, Optional[str]]]] = field(
+        default_factory=list
+    )
+    exclude_exceptions: List[Type[Exception]] = field(default_factory=list)
     expect: List[ExpectedTraverseStep] = field(default_factory=list)
     expect_alternatives: Optional[List[List[ExpectedTraverseStep]]] = None
     skip_me: Optional[str] = None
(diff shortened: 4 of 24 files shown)


tests/test_dir_traversal.py
```diff
@@ -257,6 +267,484 @@ class DirectoryTraversalVector(Generic[T]):
...
+    #
+    # The following tests are based on the pattern format rules listed in
+    # `git help ignore` (https://git-scm.com/docs/gitignore#_pattern_format).
+    #
+    # A blank line matches no files, so it can serve as a separator for
+    # readability.
+    DirectoryTraversalVector(
+        "gitignore_parsing__disregard_blank_lines",
+        given=[File("dir/file")],
+        ignore_patterns=["", ""],
```
Nour-Mws, line 316: I am not sure I follow why `", "` is a blank line

tests/test_traverse_project.py
```diff
@@ -510,6 +512,66 @@ class TraverseProjectVector:
...
+    #
+    TraverseProjectVector(
+        "empty_ignore_pattern__raises_IgnoreRuleMissing",
+        "empty",
+        code=set(),
+        deps=set(),
+        pyenvs=set(),
+        ignores={"", "\t", "    "},
+        expect_raised=IgnoreRuleMissing,
+    ),
```
Nour-Mws, line 526: The error is caused by the tab on the second line? or the general lack of ignore patterns?

tests/test_traverse_project.py
```diff
@@ -510,6 +512,66 @@ class TraverseProjectVector:
...
+        ignores={"# a comment", "# another comment"},
+        expect_raised=IgnoreRuleMissing,
+    ),
+    TraverseProjectVector(
+        "anchored_ignore_pattern_without_basedir__raises_IgnoreRuleError",
+        "empty",
+        code=set(),
+        deps=set(),
+        pyenvs=set(),
+        ignores={"foo/bar"},
```
Nour-Mws, line 542: How would this not raise the `IgnoreRuleError`?

tests/sample_projects/hidden_files/expected.toml
```diff
@@ -0,0 +1,78 @@
...
+# General information about a simplified project: Its name, why we test it,
+# its relation to real world projects
+name = "hidden_files"
+description = """
+    A project with code/deps in hidden files and inside hidden dirs, as well as
+    pyenvs that are inside hidden dirs (one of them being a hidden dir itself)
+"""
+
+[experiments.default]
+description = "Default run where everything it hidden."
```
Nour-Mws, line 11:
```suggestion
description = "Default run where everything is hidden."
```

tests/sample_projects/hidden_files/expected.toml
```diff
@@ -22,3 +22,57 @@ undeclared_deps = []
 
 # Declared dependencies which were never `import`ed from the code:
 unused_deps = []
+
+[experiments.ignore_nothing]
+description = "Run without ignores - everything is found."
+pyenvs = [""]  # Find all Python environments in project
```
Nour-Mws, line 28: Why aren't the pyenvs under `.venvs/`detected if the dot ignore pattern is overridden?

tests/sample_projects/hidden_files/expected.toml
```diff
@@ -22,3 +22,57 @@ undeclared_deps = []
...
+
+# Declared dependencies which were never `import`ed from the code:
+unused_deps = [
+    "package_x",  # .hidden.requirements.txt
+    "package_y",  # .hidden_dir/requirements.txt
+]
+
+[experiments.ignore_some_things]
+description = "Run with custom ignore pattern instead of default."
+pyenvs = [""]  # Find all Python environments in project
```
Nour-Mws, line 57: Same question as above

README.md
```diff
@@ -296,6 +296,50 @@ above. For example, if the mapping file has some stale/incorrect mapping
...
+
+The format of the ignore patterns is the same as used by `.gitignore` files,
+[see here for a full description](https://git-scm.com/docs/gitignore#_pattern_format).
+(Note that some patterns that must be interpreted relative to the location of a
+`.gitignore` file are not (yet) supported by FawltyDeps. Search for "relative"
+in the linked page to see which pattern types are not yet supported.)
+
+When the `--ignore` option is not specified, its default value is `".*"`, which
+matches all paths that start with a dot (`.`), aka. "hidden" paths. In the above
+example, if you want to ignore both hidden paths, and everything under `tests/`,
```
Nour-Mws, line 318: Which example are we referring to?

Nour-Mws (commented):
Meticulous and comprehensive as expected :sparkles: 
I am a bit confused as to some tests related to ignored venvs and I'm sure this is a lapse on my side as the tests are passing.
But I thought I'd comment and wait until I understand so that I can approve in good conscience.

README.md
jherland: `fawltydeps --ignore tests/` on line 307

tests/sample_projects/hidden_files/expected.toml
jherland: Oh, crap. This is poorly documented...

The `pyenvs = [""]` here is not the Python environments that we expect to find (i.e. the expected output), but rather the `--pyenv` values that we pass _into_ the analysis (that is, we're passing `""` which is appended onto the directory containing the sample project, meaning that we're searching through the entire sample project for Python envs). The Python environments that we end up finding are actually _not_ directly verified here at all...

tests/test_traverse_project.py
jherland: The general lack of ignore patterns. The intention here was to test that a blank line is considered empty whether or not it contains extra whitespace. These should be three different tests, I think.

tests/test_traverse_project.py
jherland: In order to make `"foo/bar"` a valid pattern in this context, we would have to specify a `base_dir` (required when the pattern is "anchored", i.e. when it contains a slash). You can see tests of this logic in `test_dir_traversal.py`, specifically look at `anchored_pattern__must_have_base_dir` and `ignore_pattern_without_slash__does_not_match_above_base_dir` for examples without/with `base_dir` respectively.

But this logic is not _yet_ available to users of `traverse_project`. At this point in the series of PRs, we can only specify ignore patterns in the CLI, and from the CLI I believe it is _unclear_ what the correct `base_dir` should be: The natural choice would be the current dir, but as soon as you pass multiple `basepath`s on the command line, and/or multiple `--code` or `--deps` paths, I think it becomes impossible to determine which `base_dir` to be used for `--ignore foo/bar`. Therefore, I've chosen to leave `base_dir` unspecified in this case, with the consequence that anchored patterns cannot be given directly to `--ignore` on the command-line.

This is reflected in our code as of this PR: `DirectoryTraversal.ignore()` does take an optional `base_dir`, but this argument is not yet used in `traverse_project.find_sources()`, and it will only start to be used when I add the logic for parsing `.gitignore` files (in the final `--ignore-from` PR that I'm still working on). When parsing `.gitignore` files, the `base_dir` is indeed well-specified: IIRC, [the `gitignore` documentation](https://git-scm.com/docs/gitignore) states that patterns in a file use the parent directory of that file as their `base_dir`.

tests/test_dir_traversal.py
jherland: There are two blank lines: `""`, and `""`...

tests/test_dir_traversal.py
```diff
@@ -257,6 +267,484 @@ class DirectoryTraversalVector(Generic[T]):
...
+        ignore_exceptions=[IgnoreRuleMissing, IgnoreRuleMissing],
```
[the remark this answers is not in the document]
jherland: We're checking that we get `IgnoreRuleMissing` raised twice, once for each empty pattern above.

[In the test below](https://github.com/tweag/FawltyDeps/pull/388/files/142b91a7559c389414ac807556daeff5d3919074#diff-3aa986ddb8b14d40d3559e3b54826bfc5b43b56d33d849bd21f82378c8723764R763) we pass each ignore pattern separately to the parser and recording each exception raised. I wanted to make sure that the first exception is not messing up some internal state that would cause problems for later ignore patterns. (When we get around to parsing gitignore files, we will encounter valid ignore patterns interspersed with empty lines, and we need to skip the empty lines and keep parsing.)

tests/test_dir_traversal.py
```diff
@@ -257,6 +267,484 @@ class DirectoryTraversalVector(Generic[T]):
...
+        expect=[
+            ExpectedTraverseStep(".", subdirs=["dir"]),
+            ExpectedTraverseStep("dir", files=["file"]),
+        ],
+    ),
+    # A line starting with # serves as a comment. Put a backslash ("\") in
+    # front of the first hash for patterns that begin with a hash.
```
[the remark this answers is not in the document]
jherland: Here is everything that the [gitignore docs](https://git-scm.com/docs/gitignore) have to say about backslashes:
> A line starting with # serves as a comment. Put a backslash ("`\`") in front of the first hash for patterns that begin with a hash.
>
> Trailing spaces are ignored unless they are quoted with backslash ("`\`").
>
> An optional prefix "`!`" which negates the pattern; any matching file excluded by a previous pattern will become included again. It is not possible to re-include a file if a parent directory of that file is excluded. Git doesn’t list excluded directories for performance reasons, so any patterns on contained files have no effect, no matter where they are defined. Put a backslash ("`\`") in front of the first "`!`" for patterns that begin with a literal "`!`", for example, "`\!important!.txt`".

As for hash characters elsewhere on the line, I just tested, and these are _not_ interpreted as comment chars, but rather as part of the pattern (so no backslash needed there). This is in line with the above docs, but maybe unexpected when comparing to another languages like e.g. shell or Python.

tests/test_dir_traversal.py
```diff
@@ -257,6 +267,484 @@ class DirectoryTraversalVector(Generic[T]):
...
+            ExpectedTraverseStep("a", files=["file"]),
+        ],
+    ),
+    DirectoryTraversalVector(
+        "anchored_pattern__must_have_base_dir",
+        given=[File("file")],
+        ignore_patterns=[("/file", None)],
+        ignore_exceptions=[IgnoreRuleError],
+        expect=[ExpectedTraverseStep(".", files=["file"])],
+    ),
```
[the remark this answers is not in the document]
jherland: Added test helper in [2f13434](https://github.com/tweag/FawltyDeps/pull/388/commits/2f13434038593f3a5a070f98f6a5a625e689a15d).

tests/test_dir_traversal.py
```diff
@@ -257,6 +267,484 @@ class DirectoryTraversalVector(Generic[T]):
...
+            ExpectedTraverseStep("dir2"),  # ignored_subdirs=["some_path"]
+        ],
+    ),
+    DirectoryTraversalVector(
+        "ignore_pattern_with_combined_slashes_and_base_dir",
+        given=[
+            Dir("b"),
+            Dir("a/b"),
+            Dir("a/a/b"),  # below pattern matches only this
+            Dir("a/a/a/b"),
```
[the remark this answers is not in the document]
jherland: Sorry, this was my horribly under-documented test code striking again... The `a/` is the base_dir for the `a/b/` pattern, not a pattern in its own right...

Hopefully clearer after 2f13434.

tests/test_dir_traversal.py
```diff
@@ -257,6 +267,484 @@ class DirectoryTraversalVector(Generic[T]):
...
+    # all files inside directory "abc", relative to the location of the
+    # .gitignore file, with infinite depth.
+    DirectoryTraversalVector(
+        "trailing_double_asterisk_after_slash_matches_everything_underneath",
+        given=[
+            File("abc/def"),
+            File("abc/a/very/long/and/deeply/nested/subdir/with/files/in/it"),
+            File("abx/yz"),  # not matched
+        ],
+        ignore_patterns=["abc/**"],
```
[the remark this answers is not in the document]
jherland: Almost. Technically speaking `abc/` ignores the `abc` directory, but `abc/**` ignores everything _inside_ that directory. From the point of view of the `DirectoryTraversal` itself, the `abc/` dir _is_ traversed (but nothing inside it). You can see this in the expected traversal step below (the second one):

```python
            ExpectedTraverseStep("abc"),  # ignored_subdirs=["a"], ignored_files=["def"]
```

This step would not occur if we ignored the `abc/` directory instead.

In practice, from the FawltyDeps traversal's POV, the difference is academic. I'm trying to construct a case where there would be a meaningful difference in behavior, but it's hard to come up with: For 'code' and 'deps', we're looking for files, so the extra directory traversal step makes no difference. For `pyenv` we are only ignoring for Python environments that are located _inside_ ignored dirs (i.e. with `.*` as the default pattern we _want_ to find `.venv/`, but we want to ignore `.hidden/.venv/`). Therefore, I believe a virtualenv located at `abc/my_venv` would be _ignored_ by the `abc/` pattern, but would be _found_ by the `abc/**` pattern.

tests/test_dir_traversal.py
```diff
@@ -257,6 +267,484 @@ class DirectoryTraversalVector(Generic[T]):
...
+    # that is directly under directory "foo".
+    DirectoryTraversalVector(
+        "ignore_pattern_double_asterisk_slash_matches_in_all_dirs_under_base",
+        given=[
+            File("foo"),  # not matched, due to base_dir a/
+            File("a/foo"),
+            File("a/b/foo"),
+            File("a/c/foo/more"),
+        ],
+        ignore_patterns=[("**/foo", "a/")],
```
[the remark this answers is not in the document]
jherland: Yes, you got it! As I've now tried to fix in 2f13434, the tuple was used when we needed to communicate an ignore pattern together with a `base_dir` (in practice: the directory of the `.gitiignore` file containing this pattern). Wrapping this in an `IgnorePattern("**/foo", base_dir="a/")` hopefully makes it more readable.

jherland:
> Meticulous and comprehensive as expected ✨
> I am a bit confused as to some tests related to ignored venvs and I'm sure this is a lapse on my side as the tests are passing.
> But I thought I'd comment and wait until I understand so that I can approve in good conscience.

Thanks for the review, and I apologize for not getting around to replying sooner.

Yes, I stupidly used `List[Union[str, Tuple[str, Optional[str]]]]` for the `DirectoryTraversalVector.ignore_patterns` member, which only added unnecessary confusion to something that is already quite complex. 2f13434 tries to improve on that.

Also, in a couple of my replies above I've tried to explain the "anchored" patterns that require the `base_dir` argument to `DirectoryTraversal.ignore()`. I regret adding (and testing) this extra-complex part here in this PR, when it is _not_ used from the POV of `traverse_project`. The reason we're not properly using this part yet, is because I don't think there _is_ a suitable `base_dir` that can be set for patterns that come from the command line. They can only be determined for patterns that come from `.gitignore` files, and support for those are only added in a future PR...

Again, I'm sorry for the unnecessary complexity that is not immediately warranted by this PR. Maybe I should remove the `base_dir` argument from `DirectoryTraversal.ignore()` for now? Or at least add a comment explaining how/why it's not yet used by `traverse_project`?

tests/sample_projects/hidden_files/expected.toml
Nour-Mws: Oops

tests/test_dir_traversal.py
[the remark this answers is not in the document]
Nour-Mws: Thanks for the explanation!

Nour-Mws (approved):
Thanks for taking the time to explain in details. I have a much better understanding of the context now and the blind spots in the tests are clearer.
It took me more than I expected to go through the replies now that a certain duration has passed since I've last reviewed this.
All good. This is ready to go!

jherland:
@mknorps: I had to make a few changes here to reconcile this PR with your recent Windows work. In short, I added commit 3a2b96f to refactor how we skip individual `DirectoryTraversalVector`s on Windows, and I additionally made the following changes (compared to what was already approved) to make the tests pass on Windows and MacOS:

```diff
diff --git a/tests/test_cmdline.py b/tests/test_cmdline.py
index 196d31f..1ba9bb2 100644
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -241,7 +241,7 @@ def test_list_imports__from_dir_with_some_ignored__prints_imports_from_unignored
         }
     )
 
-    expect = [f"{tmp_path}/file3.ipynb[1]:1: pytorch"]
+    expect = [f"{tmp_path / 'file3.ipynb'}[1]:1: pytorch"]
     output, returncode = run_fawltydeps_function(
         "--list-imports", "--detailed", f"--code={tmp_path}", "--ignore=*.py"
     )
diff --git a/tests/test_dir_traversal.py b/tests/test_dir_traversal.py
index e3e9153..fef53b2 100644
--- a/tests/test_dir_traversal.py
+++ b/tests/test_dir_traversal.py
@@ -375,6 +387,8 @@ directory_traversal_vectors: List[DirectoryTraversalVector] = [
                 ".", files=["do_not_ignore "], ignored_files=["do_ignore "]
             ),
         ],
+        skip_me=on_windows("Windows does not support paths with trailing spaces"),
+        # https://stackoverflow.com/a/67886889
     ),
     # An optional prefix "!" which negates the pattern; any matching file
     # excluded by a previous pattern will become included again. It is not
@@ -608,7 +622,7 @@ directory_traversal_vectors: List[DirectoryTraversalVector] = [
         "range_matches_any_one_char_in_range",
         given=[
             File("abcdef"),  # not matched
-            File("abcxdef"),  # not matched
+            File("abcWdef"),  # not matched
             File("abcXdef"),
             File("abcYdef"),
             File("abcZdef"),
@@ -620,7 +634,7 @@ directory_traversal_vectors: List[DirectoryTraversalVector] = [
             ExpectedTraverseStep(
                 ".",
                 subdirs=["abc"],
-                files=["abcdef", "abcxdef", "abcXYXdef"],
+                files=["abcdef", "abcWdef", "abcXYXdef"],
                 ignored_files=["abcXdef", "abcYdef", "abcZdef"],
             ),
             ExpectedTraverseStep("abc", files=["def"]),
```

fawltydeps/settings.py
```diff
@@ -162,6 +162,7 @@ class Settings(BaseSettings):
     ignore_unused: Set[str] = DEFAULT_IGNORE_UNUSED
     deps_parser_choice: Optional[ParserChoice] = None
     install_deps: bool = False
+    ignores: Set[str] = {".*"}
```
GuillaumeDesforges, line 165: Since Maria mentionned it in the dataeng weekly, I'll just add a quick comment that can be safely ignored.

Since FawltyDeps "ignore" can be about different things, namely import of file paths, as a user I may be confused by the name `--ignores`. A more explicit option name (e.g. `--ignore-files`) would be less confusing to me.

mknorps: We have 
- `--ignore-undeclared` to ignore imports that do not have declared dependencies
- `--ignore` for paths we traverse (including files).
So we have an explicit name but for a first case you mentioned. Is this OK?

tests/test_dir_traversal.py
```diff
@@ -106,8 +128,17 @@ class DirectoryTraversalVector(Generic[T]):
     given: List[BaseEntry]
     add: List[AddCall] = field(default_factory=lambda: [AddCall(path=".")])
     skip_dirs: List[str] = field(default_factory=list)
+    ignore_patterns: List[IgnorePattern] = field(default_factory=list)
+    ignore_exceptions: List[Type[Exception]] = field(default_factory=list)
+    ignore_patterns_raise_ValueError: bool = False
     expect: List[ExpectedTraverseStep] = field(default_factory=list)
     expect_alternatives: Optional[List[List[ExpectedTraverseStep]]] = None
+    skip_me: Callable[[], Optional[str]] = lambda: None
```
mknorps, line 136: Maybe we could simplify the type here to just:
```suggestion
    skip_me: Optional[str] = None
```
We use only `on_windows` function to pupolate this field for now, and maybe we do not need to over-abstarct this.
Unless there is some other reason, that skips me.

mknorps (approved):
I imagine this was a big rebase :1st_place_medal: ...

It looks good to me. I had one question/suggestion regarding the `skip_me` option.

fawltydeps/settings.py
jherland: Thanks for raising this question @GuillaumeDesforges. Although we are eager to get this new feature merged and published, this is probably the best time to make sure we get the naming right  (_before_ committing to it in a public release).

FTR, we also have:
- `--ignore-unused` to ignore dependencies that are not (meant to be) imported, such as tools, etc.
- `--ignore-from`: This does not exist yet, but it's what I'm currently working on to support getting `--ignore` pattens indirectly from a file of patterns. To be used like this: `fawltydeps --ignore-from=.gitignore`.

Although all of these options have different names, I cannot prove that this will **NOT** be confusing to some users. There is nothing inherent in the name of  `--ignore` (or `--ignore-from`) that explains that these are about ignoring _paths_. Instead, we hope/assume/believe that users will most easily associate the concept of ignoring _something_ to the paths that are given as input to FawltyDeps. (And if not, we hope that `--help` or the docs will quickly set them straight.)

We could go with something longer, like `--ignore-paths` (and `--ignore-paths-from`?), which would be clearer, but more tedious to type out.

We could also try to re-evaluate our naming and come up with a better scheme. For one: `--ignore-undeclared` and `--ignore-unused` are about ignoring facets of the analysis done by FawltyDeps, but ignoring paths is really about excluding paths from the analysis entirely. As such the path-related options are logically applied/handled at a very different time than the other two. We could consider something like `--exclude`/`--exclude-from` to denote that these are _different_ from the existing `--ignore-undeclared`/`--ignore-unused`? That said, I still find the overlapping of terms across different tools a little clunky here: i.e. that the "gitignore option" is named `--exclude-from=.gitignore` when we have other options called `--ignore-something`.

Maybe we should take a closer look at how other linter tools handle this situation, although I suspect we won't find much consensus...

tests/test_dir_traversal.py
jherland: True, I was thinking _too_ generically here (that the test vector should be able to check something _when the test is run_ to decide whether or not to skip), but (since Python is an interpreted language) the Windows-or-not question is already known when we create the test vector here. Will fix. 🙏🏻

jherland: Fixed in 584fc1d45a1b570fd414e74b4b8cf3539674978c.

fawltydeps/settings.py
mknorps: I really like this `--exclude`/`--exclude-from` idea!
It clearly separates what we ignore in traversing directories and what we ignore when showing results. 
In other words `--exclude...` is in the preprocessing step and `--ignore...` in post-processing.

Currently we have one CLI option that uses file as a helper for a setup and it is `--custom-mapping-file`. `--exclude-file` would be extremely confusing, since its purpose it to load a file that defines ignored paths in the gitignore-style. Maybe we should consider having `--custom-mapping-from`. But this would mean backward incompatibility.

jherland: A quick survey of what other linters do:
- ruff:
  - [`exclude`](https://docs.astral.sh/ruff/settings/#exclude) works on "file patterns" (i.e. adjust analysis input)
  - [`ignore`](https://docs.astral.sh/ruff/settings/#ignore) works on "rule codes" (i.e. adjust result of analysis)
- flake8:
  - [`--exclude`](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-exclude) works on files.
  - [`--ignore`](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-ignore) works on "codes".
- pylint:
  - [`--ignore` (and friends)](https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html#ignore) are used to ignore paths.
  - [`--disable`](https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html#disable) seems to be the main way to disable checkers/rules/codes.
  - Various checkers add other `--ignore-*` options, but these are for ignoring checker-specific details (not related to paths at all).
  - A couple of checkers add `--exclude-*` options, for similar purposes.
  - In general, I don't think pylint is what we should try to emulate. It's interface is a bewildering mess, IMHO.
- black:
  - [`--exclude`](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#exclude): "A regular expression that matches files and directories that should be excluded on recursive searches."
  - Black is famously lacking in options to change its logic/behavior, but there are a few options named [`--skip-*`](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#x-skip-source-first-line) to disable a few parts of its internal logic.
- Mypy:
  - [`exclude`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-exclude) is used to exclude paths
  - For ignoring/disabling internal rules, there is a plethora of options with names like `disallow_*` or `warn_*`.
  - There is not much use of `ignore`, but the few uses that are visible are certainly more related to rules ([`ignore_errors`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-ignore_errors), [`ignore_missing_imports`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-ignore_missing_imports), [`warn_unused_ignores`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-warn_unused_ignores), or [`# type: ignore`](https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html#silencing-type-errors)) than to paths/files.

I did not expect to find this much consensus, and it seems a change of opinion is in order:

I now think we should rename `--ignore` (and `--ignore-from`) to `--exclude` (and `--exclude-from`[^1]).

What do you think, @GuillaumeDesforges and @mknorps?

[^1]: Although it's worth having another look at how other tools deal with ignore patterns read from a file like `.gitignore`.

jherland: (I didn't see your last comment @mknorps before I posted the above survey, but it seems we're in agreement. 🎉)

Re. renaming `--custom-mapping-file` to `--custom-mapping-from`: Yes, it's backwards incompatible, so we would have to keep `...-file` as a deprecated alias to `...-from` for a while, before we remove it completely. Other than that, I don't think it's a problem to rename this.

mknorps: > I now think we should rename --ignore (and --ignore-from) to --exclude (and --exclude-from[1](https://github.com/tweag/FawltyDeps/pull/388#user-content-fn-1-67c636c8d29ad3ceedc5aad716e06f53)).

aye!

GuillaumeDesforges: Makes sense to me :+1: this is probably what triggered my initial comment, somehow my brain had internalized how other tools behave :D

jherland: Thanks @GuillaumeDesforges for bothering us enough to fix it before we released it! 😅

I've rebased/replaced everything from `ignore` to `exclude` now. I took extra care in the docs to make sure things are as clear and unambiguous as I could make them.

mknorps (changes requested):
There are some surprising behaviors with `--exclude` option. It looks like the problem is related to absolute vs relative paths, but it should be checked.

Two examples below:

- Example 1

    Run FawltyDeps on itself with `--exclude` option like that:
    ```
    (fawltydeps-RB6LwJVc-py3.11) ➜  FawltyDeps git:(jherland/ignore-work) fawltydeps --exclude fawltydeps
    
    No undeclared or unused dependencies detected.
    ```

    I thought I will have lots of unused dependencies.

- Example 2
  ```
  (fawltydeps-RB6LwJVc-py3.11) ➜  FawltyDeps git:(jherland/ignore-work) fawltydeps . --exclude 'fawltydeps/traverse_project.py'
  usage: fawltydeps [--check | --check-undeclared | --check-unused | --list-sources | --list-imports | --list-deps]
                [--summary | --detailed | --json] [--code PATH_OR_STDIN [PATH_OR_STDIN ...]]
                [--deps PATH [PATH ...]]
                [--deps-parser-choice {requirements.txt,setup.py,setup.cfg,pyproject.toml}]
                [--pyenv PYENV_DIR [PYENV_DIR ...]] [--exclude PATTERN [PATTERN ...]] [--install-deps]
                [--custom-mapping-file FILE_PATH [FILE_PATH ...]] [--config-file CONFIG_FILE]
                [--ignore-undeclared IMPORT_NAME [IMPORT_NAME ...]] [--ignore-unused DEP_NAME [DEP_NAME ...]]
                [-v] [-q] [--generate-toml-config] [-V] [-h]
                [basepaths ...]
  fawltydeps: error: Error while parsing exclude pattern: Anchored pattern without base_dir in 'fawltydeps/traverse_project.py'
  ```

jherland:
> * Example 1
>   Run FawltyDeps on itself with `--exclude` option like that:
>   ```
>   (fawltydeps-RB6LwJVc-py3.11) ➜  FawltyDeps git:(jherland/ignore-work) fawltydeps --exclude fawltydeps
>   
>   No undeclared or unused dependencies detected.
>   ```
>   
>   I thought I will have lots of unused dependencies.

There are certainly problems with absolute vs relative paths that are still to be resolved. But this example  has _another_ complicating factor as well: Our own configuration from pyproject.toml contains: `code = ["fawltydeps"]`, meaning that we end up with both `code = {"fawltydeps"}` and `exclude = {"fawltydeps"}` in our `Settings` object, and in that case, the `code` takes precedence over the `exclude`. This was not obvious to you, the user. I'll certainly add a warning about this special case.

jherland:
> * Example 2
>   ```
>   $ fawltydeps . --exclude 'fawltydeps/traverse_project.py'
>   usage: [...]
>   fawltydeps: error: Error while parsing exclude pattern: Anchored pattern without base_dir in 'fawltydeps/traverse_project.py'
>   ```

This is currently expected (but unfortunate). The base_dir referred to in the error message is not the same as a the basepath you supplied on the command line, but rather a separate piece of data that follows the exclude pattern: When the pattern is read from a `.gitignore` file, the base_dir is the directory holding this `.gitignore` file. (and it means that the pattern will only apply to paths _within_ this directory, e.g. it prevents a pattern in `foo/.gitignore` to be applied to paths outside `foo/`). The base_dir is also needed whenever the pattern is "anchored"[^1]. However, this not so easy when the pattern does not come from a `.gitignore` file, but comes from the command-line instead.

Here are some examples to show why this does not easily work:
- `fawltydeps --exclude foo/bar`: Here the default basepath is `./` so we could easily argue that the `foo/bar` pattern must match `./foo/bar` only.
- `fawltydeps path/to/my_project --exclude foo/bar`: Now we've changed the basepath. We _could_ argue that the pattern should be interpreted relative to the customized basepath, i.e. the `foo/bar` pattern should match `path/to/my_project/foo/bar` only.
- `fawltydeps path/to/my/scripts path/to/my/notebooks --exclude foo/bar`: Now we have _two_ basepaths. Where does the pattern apply?
  - Both `path/to/my/scripts/foo/bar` and `path/to/my/notebooks/foo/bar`?
  - None of them?
  - Auto-calculate the common parent and use that: `path/to/my/foo/bar`? Probably not...
- `fawltydeps --code path/to/my/scripts --deps path/to/my/deps --exclude foo/bar`: Now things are getting really complex: We're looking for code under `path/to/my/scripts`, deps under `path/to/my/deps`, and Python environments under `./`. But where should the pattern apply?
  - All places? `path/to/my/scripts/foo/bar`, `path/to/my/deps/foo/bar`, and `./foo/bar`?
  - None of them?
  - Something else?

I decided early on to NOT deal with this problem just yet, and thus the current behavior is that CLI-supplied patterns get no `base_dir`, and thus cannot be anchored (i.e. the only place you can put a `/` is at the end, to signal a directory-only match). But it is becoming clear to me that we rather need to fix this now, _before_ I have finished the third PR...

Right now, I am leaning towards using all paths as base_dirs, that is, for the last example above, the `foo/bar` exclude pattern technically ends up as three separate exclude patterns (with `path/to/my/scripts`, `path/to/my/deps`, and `./` as their respective base_dirs). At least I hope this would cause the least amount of surprises...

[^1]:  This comes from the `git help ignore` "specification": A pattern with a `/` at the start or in the middle is always interpreted relative to the directory in which the pattern was found.

jherland:
For the record, I'm fixing the above issues in a new PR: #415. Although I'm not rewriting much code, and could potentially squash the fixes into this PR, I would have to rebase them across PR #391, and that introduces a substantial amount of conflicts that I'd rather avoid. This PR and #391 are already reviewed and with #415 fixing these latest issues, I plan on merging all 3 to `main` once #415 is finished and reviewed.

jherland:
#415 is ready for review now. It should fix the issues you found above, @mknorps.

mknorps (approved):
With #415 this is ready to be merged :)

PR #394, opened 2023-11-09
Reintroduce current Python environment as a separate level in our resolver stack

### LINKED ISSUE #392: FawltyDeps fails running on itself
**Describe the bug**
When I install from scratch a developer's environment and run `fawltydeps` in it I get following message:
```
(fawltydeps-py3.11) ➜  FawltyDeps git:(main) fawltydeps

These imports appear to be undeclared dependencies:
- 'pkg_resources'

These dependencies appear to be unused (i.e. not imported):
- 'setuptools'
- 'types-setuptools'

```

**To Reproduce**
Run fawltydeps from `poetry shell`.

**Expected behavior**
FawltyDeps does not report any missing or unused dependencies

**Environment**
 - OS name + version: I tested on Ubuntu and Windows.
 - Version of the code:
 ```
(fawltydeps-py3.11) ➜  FawltyDeps git:(main) fawltydeps --version
FawltyDeps v0.13.2
(fawltydeps-py3.11) ➜  FawltyDeps git:(main) which fawltydeps
/home/mknorps/.cache/pypoetry/virtualenvs/fawltydeps-RB6LwJVc-py3.11/bin/fawltydeps
```

### PR DESCRIPTION
**Pull request by jherland**
Fixes #392.

We foresee no harm or significant cost by simply using the current Python environment as a separate step in FawltyDeps' resolver stack.

The step is introduced _after_ any other local environments (found inside the project), but before we either fall back to then expensive `TemporaryPipInstallResolver`, or the suboptimal `IdentityMapping`.

As explained in the third commit message, this is somewhat of a return to our roots:
> In fact, this is partially a return to our pre v0.7 behavior where we would always (then: only) find packages in `sys.path`. Since v0.7 we allowed `--pyenv` to override this, and since v0.12, we also started searching for pyenvs under `--pyenv` or `basepaths` given on the command line. The result of this is that by now, the current environment is only consulted in the fallback where no other environment is possibly found.

Commits:
- Preparation:
  - `packages`: Expose decision to use current env to resolve packages
  - `packages`: Start separating handling of project pyenvs from current env
- Fix:
  - Always include current env in the package resolver stack
- Cleanup:
  - `packages`: Split `SysPathPackageResolver` out of `LocalPackageResolver`

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,9 +161,10 @@ customization options are summarized in the table below:
 | Priority | Mapping strategy        | Options |
 |----------|----------------------|------------|
 | 1        | User-defined mapping | Provide a custom mapping in TOML format via `--custom-mapping-file` or a `[tool.fawltydeps.custom_mapping]` section in `pyproject.toml`. <br /> Default: No custom mapping|
-| 2        | Mapping from installed packages | Point to one or more environments via `--pyenv`.<br />Default: auto-discovery of Python environments under the project’s basepath. If none are found, default to the Python environment in which FawltyDeps itself is installed.|
-| 3a       | Mapping via temporary installation of packages  | Activated with the `--install-deps` option.|
-| 3b       | Identity mapping | Active by default. Deactivated when `--install-deps` is used. |
+| 2        | Mapping from installed packages found inside project | Point to one or more environments with `--pyenv`.<br />Default: auto-discovery of Python environments under the project’s basepath.|
+| 3        | Mapping from packages installed in `sys.path` | Active by default. No CLI option. This finds packages installed in the Python environment in which FawltyDeps itself runs.|
+| 4a       | Mapping via temporary installation of packages  | Activated with the `--install-deps` option.|
+| 4b       | Identity mapping | Active by default. Deactivated when `--install-deps` is used. |
 
 
 #### Local Python environment mapping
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by jherland** [comment]
For a more visual representation of the changes introduced here, our resolver stack is modified from what we described in the [recent blog post](https://www.tweag.io/blog/2023-09-21-fawltydeps-mapping-strategy/) into this:

fawltydeps_resolver_stack_2023-11-09

File: README.md [hunk 1]
```diff
@@ -161,9 +161,10 @@ customization options are summarized in the table below:
 | Priority | Mapping strategy        | Options |
 |----------|----------------------|------------|
 | 1        | User-defined mapping | Provide a custom mapping in TOML format via `--custom-mapping-file` or a `[tool.fawltydeps.custom_mapping]` section in `pyproject.toml`. <br /> Default: No custom mapping|
-| 2        | Mapping from installed packages | Point to one or more environments via `--pyenv`.<br />Default: auto-discovery of Python environments under the project’s basepath. If none are found, default to the Python environment in which FawltyDeps itself is installed.|
-| 3a       | Mapping via temporary installation of packages  | Activated with the `--install-deps` option.|
-| 3b       | Identity mapping | Active by default. Deactivated when `--install-deps` is used. |
+| 2        | Mapping from installed packages found inside project | Point to one or more environments with `--pyenv`.<br />Default: auto-discovery of Python environments under the project’s basepath.|
+| 3        | Mapping from packages installed in `sys.path` | No option, this finds packages installed in the Python environment in which FawltyDeps itself is installed.|
```
mknorps, line 165: Nit: for the reading consistency
```suggestion
| 3        | Mapping from packages installed in `sys.path` | Active by default. No CLI option. This finds packages installed in the Python environment in which FawltyDeps itself runs.|
```

fawltydeps/packages.py
```diff
@@ -499,7 +528,11 @@ def setup_resolvers(
         mapping_paths=custom_mapping_files or set(), custom_mapping=custom_mapping
     )
 
-    yield LocalPackageResolver(pyenv_srcs)
+    if pyenv_srcs:
+        yield LocalPackageResolver(pyenv_srcs)
```
mknorps, line 532: Nit: If we give empty `pyenv_src` then `LocalPackageResolver` will not produce any mapping. For the consistency sake I think it would be better to drop this condition here. We have conditional resolver usage in case of 4a and 4b, and 3 but they arer governed by explicitly provided flags.
```suggestion
    yield LocalPackageResolver(pyenv_srcs)
```

mknorps (approved):
Thank you Johan! 
I ejoyed this PR greatly. It is always a pleasure to check your PRs. You are a master of PR storytelling 💚 

Apart from some small nitpicks, which you may or may not take into account, I think this is a ready-to-go contribution and a definite improvement, especially for the newcomers who will not be surprised when using for example Poetry.
 🚀

fawltydeps/packages.py
jherland: Sure, I wanted to skip the creation of an unnecessary object, but as we have no indication of a performance problem here, I agree that clarity/consistency is more important. 👍🏻

PR #396, opened 2023-11-16
Update our Python project metadata

jherland:
Commits:
- Declare that we support Python v3.12
- Progress from Alpha to Beta quality :-)


mknorps (approved):
:green_heart:

PR #397, opened 2023-11-20
MacOS and Windows support

mknorps:
## Support for MacOS

To support MacOS it is enough to add `macos-latest` to the test matrix and allow for longer runs for tests that use `--install-deps` option. 

## Support for Windows
 To support Windows we needed to:
- change all paths to be system-agnostic (5843d1efced695bf374b82fe3f385a1bdb97a8f4)
- change paths that referred to python executable, pip executable and site packages to work on all systems (36e957a174ff5bd68fc48dcefac43447380bdbe9, 8c4ecb984d9bf5ebe6d57991aee1f2bc4f76714e, d68917162b5f58a49c678122f824f80c2755e43d)
- Adjust command line script that creates virtual environments for tests (73e1819a38b83a2e51728bf5832df53c8e047ffd)
- Skip irrelevant tests (442db9696631777e3c93f56823f27b9d4abf9525, 3b5e3b6870e5ee9d827bd9f5b00676d58452af37)
- skip real-project tests as the test setup does not work well with Windows and we may consider adjusting the tests  (03fcd441b444c735d173c024046a1a934ded26fb)
- few other minor tweaks.

## Issues
In the process of adjusting the code to support Windows, we discovered, that the `real_projects` tests setup is not suited for Windows. The fact that we use a separate virtual environment, which we construct from the command line makes it hard to work with Windows. 

You may see example of such problem [in this CI run](https://github.com/tweag/FawltyDeps/actions/runs/7263529972/job/19789047009?pr=397). In this case, project [The Algorithms-Python](https://github.com/TheAlgorithms/Python) uses conditional requirement:
```
    "tensorflow; python_version < '3.11'",
```
This falls into undeclared dependencies on Windows for Python <=3.10. This behavior should be investigated in a separate ticket.

### DIFF
--- a/fawltydeps/extract_declared_dependencies.py
+++ b/fawltydeps/extract_declared_dependencies.py
@@ -149,7 +149,7 @@ def parse_setup_cfg(path: Path) -> Iterator[DeclaredDependency]:
         # TODO: try leveraging RequirementsFile.from_string once
         #       pip-requirements-parser updates.
         # See:  https://github.com/nexB/pip-requirements-parser/pull/17
-        with NamedTemporaryFile(mode="wt") as tmp:
+        with NamedTemporaryFile(mode="wt", delete=False) as tmp:
             tmp.write(value)
             tmp.flush()
             for dep in parse_requirements_txt(Path(tmp.name)):
--- a/fawltydeps/extract_declared_dependencies.py
+++ b/fawltydeps/extract_declared_dependencies.py
@@ -152,7 +152,7 @@ def parse_setup_cfg(path: Path) -> Iterator[DeclaredDependency]:
         # See:  https://github.com/nexB/pip-requirements-parser/pull/17
 
         # https://github.com/nexB/pip-requirements-parser/pull/19#discussion_r1379279880
-        temp_file = NamedTemporaryFile( # pylint: disable=consider-using-with
+        temp_file = NamedTemporaryFile(  # pylint: disable=consider-using-with
             "wt",
             delete=False,
             # we prefer utf8 encoded strings, but ...
--- a/fawltydeps/extract_declared_dependencies.py
+++ b/fawltydeps/extract_declared_dependencies.py
@@ -7,6 +7,7 @@ import re
 import sys
 import tokenize
 from dataclasses import replace
+from os import unlink
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Callable, Iterable, Iterator, NamedTuple, Optional, Tuple
@@ -149,11 +150,24 @@ def parse_setup_cfg(path: Path) -> Iterator[DeclaredDependency]:
         # TODO: try leveraging RequirementsFile.from_string once
         #       pip-requirements-parser updates.
         # See:  https://github.com/nexB/pip-requirements-parser/pull/17
-        with NamedTemporaryFile(mode="wt", delete=False) as tmp:
-            tmp.write(value)
-            tmp.flush()
-            for dep in parse_requirements_txt(Path(tmp.name)):
+
+        # https://github.com/nexB/pip-requirements-parser/pull/19#discussion_r1379279880
+        temp_file = NamedTemporaryFile( # pylint: disable=consider-using-with
+            "wt",
+            delete=False,
+            # we prefer utf8 encoded strings, but ...
+            # - must not change newlines
+            # - must not  change encoding, fallback to system encoding for compatibility
+            newline="",
+            encoding=None,
+        )
+        temp_file.write(value)
+        temp_file.close()
+        try:
+            for dep in parse_requirements_txt(Path(temp_file.name)):
                 yield replace(dep, source=source)
+        finally:
+            unlink(Path(temp_file.name))
 
     def extract_section(section: str) -> Iterator[DeclaredDependency]:
         if section in parser:
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -433,11 +433,10 @@ class TemporaryPipInstallResolver(BasePackageResolver):
             text=True,
             check=False,
         )
-        pip_path = (
-            venv_dir / "Scripts" / "pip.exe"
-            if platform.system() == "Windows"
-            else venv_dir / "bin" / "pip"
-        )
+        if sys.platform.startswith("win"):  # Windows
+            pip_path = venv_dir / "Scripts" / "pip.exe"
+        else:  # Assume POSIX
+            pip_path = venv_dir / "bin" / "pip"
 
         argv = [
             f"{pip_path}",
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -1,6 +1,7 @@
 """Encapsulate the lookup of packages and their provided import names."""
 
 import logging
+import os
 import platform
 import subprocess
 import sys
@@ -340,6 +341,15 @@ class LocalPackageResolver(InstalledPackageResolver):
         # Workaround for projects using PEP582:
         if path.name == "__pypackages__":
             for site_packages in path.glob("?.*/lib"):
+                if site_packages.is_dir():
+                    yield site_packages
+                    found = True
+            if found:
+                return
+
+        # Check for packages on Windows
+        if platform.system() == "Windows":
+            for site_packages in path.glob(os.path.join("Lib", "site-packages")):
                 if site_packages.is_dir():
                     yield site_packages
                     found = True
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -2,7 +2,6 @@
 
 import logging
 import os
-import platform
 import subprocess
 import sys
 import tempfile
@@ -348,7 +347,7 @@ class LocalPackageResolver(InstalledPackageResolver):
                 return
 
         # Check for packages on Windows
-        if platform.system() == "Windows":
+        if sys.platform.startswith("win"):
             for site_packages in path.glob(os.path.join("Lib", "site-packages")):
                 if site_packages.is_dir():
                     yield site_packages
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -1,7 +1,6 @@
 """Encapsulate the lookup of packages and their provided import names."""
 
 import logging
-import os
 import subprocess
 import sys
 import tempfile
@@ -42,7 +41,7 @@ from fawltydeps.types import (
     UnparseablePathException,
     UnresolvedDependenciesError,
 )
-from fawltydeps.utils import calculated_once
+from fawltydeps.utils import calculated_once, site_packages
 
 if sys.version_info >= (3, 11):
     import tomllib  # pylint: disable=no-member
@@ -329,28 +328,30 @@ class LocalPackageResolver(InstalledPackageResolver):
         # Note that we must also accept lib/pythonX.Y/site-packages for python
         # versions X.Y that are different from the current Python version.
         found = False
-        if (path / "bin/python").is_file():
-            for site_packages in path.glob("lib/python?.*/site-packages"):
-                if site_packages.is_dir():
-                    yield site_packages
+
+        if sys.platform.startswith("win"):  # Check for packages on Windows
+            if (path / "Scripts" / "python.exe").is_file():
+                _site_packages = site_packages(path)
+                if _site_packages.is_dir():
+                    yield _site_packages
                     found = True
             if found:
                 return
+
+        else:  # Assume POSIX
+            if (path / "bin/python").is_file():
+                for _site_packages in path.glob("lib/python?.*/site-packages"):
+                    if _site_packages.is_dir():
+                        yield _site_packages
+                        found = True
+                if found:
+                    return
 
         # Workaround for projects using PEP582:
         if path.name == "__pypackages__":
-            for site_packages in path.glob("?.*/lib"):
-                if site_packages.is_dir():
-                    yield site_packages
-                    found = True
-            if found:
-                return
-
-        # Check for packages on Windows
-        if sys.platform.startswith("win"):
-            for site_packages in path.glob(os.path.join("Lib", "site-packages")):
-                if site_packages.is_dir():
-                    yield site_packages
+            for _site_packages in path.glob("?.*/lib"):
+                if _site_packages.is_dir():
+                    yield _site_packages
                     found = True
             if found:
                 return
--- a/fawltydeps/packages.py
+++ b/fawltydeps/packages.py
@@ -1,6 +1,7 @@
 """Encapsulate the lookup of packages and their provided import names."""
 
 import logging
+import platform
 import subprocess
 import sys
 import tempfile
@@ -422,8 +423,14 @@ class TemporaryPipInstallResolver(BasePackageResolver):
             text=True,
             check=False,
         )
+        pip_path = (
+            venv_dir / "Scripts" / "pip.exe"
+            if platform.system() == "Windows"
+            else venv_dir / "bin" / "pip"
+        )
+
         argv = [
-            f"{venv_dir}/bin/pip",
+            f"{pip_path}",
             "install",
             "--no-deps",
             "--quiet",
--- a/fawltydeps/types.py
+++ b/fawltydeps/types.py
@@ -1,7 +1,6 @@
 """Common types used across FawltyDeps."""
 
 import os
-import platform
 import sys
 from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass, field, replace
@@ -158,7 +157,7 @@ class PyEnvSource(Source):
             return  # also ok
 
         # Support Windows projects
-        if platform.system() == "Windows" and self.path.match(
+        if sys.platform.startswith("win") and self.path.match(
             os.path.join("Lib", "site-packages")
         ):
             return  # also ok
--- a/fawltydeps/types.py
+++ b/fawltydeps/types.py
@@ -1,5 +1,7 @@
 """Common types used across FawltyDeps."""
 
+import os
+import platform
 import sys
 from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass, field, replace
@@ -153,6 +155,12 @@ class PyEnvSource(Source):
                 return  # all ok
         # Also support projects using __pypackages__ from PEP582:
         elif self.path.match("__pypackages__/?.*/lib"):
+            return  # also ok
+
+        # Support Windows projects
+        if platform.system() == "Windows" and self.path.match(
+            os.path.join("Lib", "site-packages")
+        ):
             return  # also ok
 
         raise ValueError(f"{self.path} is not a valid dir for Python packages!")
--- a/fawltydeps/types.py
+++ b/fawltydeps/types.py
@@ -1,6 +1,5 @@
 """Common types used across FawltyDeps."""
 
-import os
 import sys
 from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass, field, replace
--- a/fawltydeps/types.py
+++ b/fawltydeps/types.py
@@ -100,7 +100,7 @@ class CodeSource(Source):
 
     def render(self, detailed: bool) -> str:
         if detailed and self.base_dir is not None:
-            return f"{self.path} (using {self.base_dir}/ as base for 1st-party imports)"
+            return f"{self.path} (using {self.base_dir} as base for 1st-party imports)"
         return f"{self.path}"
 
 
--- a/fawltydeps/types.py
+++ b/fawltydeps/types.py
@@ -148,18 +148,20 @@ class PyEnvSource(Source):
     def __post_init__(self) -> None:
         super().__post_init__()
         assert self.path.is_dir()  # sanity check
-        # Support vitualenvs, poetry2nix envs, system-wide installs, etc.
-        if self.path.match("lib/python?.*/site-packages"):
-            if (self.path.parent.parent.parent / "bin/python").is_file():
-                return  # all ok
-        # Also support projects using __pypackages__ from PEP582:
-        elif self.path.match("__pypackages__/?.*/lib"):
-            return  # also ok
 
         # Support Windows projects
-        if sys.platform.startswith("win") and self.path.match(
-            os.path.join("Lib", "site-packages")
-        ):
+        if sys.platform.startswith("win"):
+            if self.path.match(str(Path("Lib", "site-packages"))):
+                if (self.path.parent.parent / "Scripts" / "python.exe").is_file():
+                    return  # also ok
+        # Support vitualenvs, poetry2nix envs, system-wide installs, etc.
+        else:
+            if self.path.match("lib/python?.*/site-packages"):
+                if (self.path.parent.parent.parent / "bin/python").is_file():
+                    return  # all ok
+
+        # Also support projects using __pypackages__ from PEP582:
+        if self.path.match("__pypackages__/?.*/lib"):
             return  # also ok
 
         raise ValueError(f"{self.path} is not a valid dir for Python packages!")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,4 @@
 """Fixtures for tests"""
-import platform
 import sys
 import venv
 from pathlib import Path
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,6 +12,7 @@ import pytest
 from fawltydeps.types import TomlData
 
 from .project_helpers import TarballPackage
+from .utils import site_packages
 
 
 @pytest.fixture
@@ -53,15 +54,7 @@ def fake_venv(tmp_path):
             venv_dir.parent.mkdir(parents=True, exist_ok=True)
         venv.create(venv_dir, with_pip=False)
 
-        # Create fake packages
-        major, minor = py_version
-
-        def _env_site_packages():
-            if platform.system() == "Windows":
-                return venv_dir / "Lib" / "site-packages"
-            return venv_dir / "lib" / f"python{major}.{minor}" / "site-packages"
-
-        site_dir = _env_site_packages()
+        site_dir = site_packages(venv_dir)
         assert site_dir.is_dir()
         for package_name, import_names in fake_packages.items():
             # Create just enough files under site_dir to fool importlib_metadata
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 """Fixtures for tests"""
+import platform
 import sys
 import venv
 from pathlib import Path
@@ -54,7 +55,13 @@ def fake_venv(tmp_path):
 
         # Create fake packages
         major, minor = py_version
-        site_dir = venv_dir / f"lib/python{major}.{minor}/site-packages"
+
+        def _env_site_packages():
+            if platform.system() == "Windows":
+                return venv_dir / "Lib" / "site-packages"
+            return venv_dir / "lib" / f"python{major}.{minor}" / "site-packages"
+
+        site_dir = _env_site_packages()
         assert site_dir.is_dir()
         for package_name, import_names in fake_packages.items():
             # Create just enough files under site_dir to fool importlib_metadata
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,9 +9,9 @@ from typing import Callable, Dict, List, Optional, Set, Tuple, Union
 import pytest
 
 from fawltydeps.types import TomlData
+from fawltydeps.utils import site_packages
 
 from .project_helpers import TarballPackage
-from .utils import site_packages
 
 
 @pytest.fixture
--- a/tests/project_helpers.py
+++ b/tests/project_helpers.py
@@ -101,7 +101,7 @@ class TarballPackage:
 
     @property
     def cache_key(self) -> str:
-        return f"fawltydeps/{self.tarball_name()}"
+        return os.path.join("fawltydeps", f"{self.tarball_name()}")
 
     def tarball_path(self, cache: pytest.Cache) -> Path:
         return self.cache_dir(cache) / self.tarball_name()
@@ -166,7 +166,7 @@ class CachedExperimentVenv:
         script or the requirements to create that venv change.
         """
         # We cache venv dirs using the hash from create_venv_hash
-        cached_str = cache.get(f"fawltydeps/{self.venv_hash()}", None)
+        cached_str = cache.get(os.path.join("fawltydeps", f"{self.venv_hash()}"), None)
         if cached_str is not None and Path(cached_str, ".installed").is_file():
             return Path(cached_str)  # already cached
 
@@ -174,6 +174,7 @@ class CachedExperimentVenv:
         venv_dir = Path(cache.mkdir(f"fawltydeps_venv_{self.venv_hash()}"))
         logger.info(f"Creating venv at {venv_dir}...")
         venv_script = self.venv_script_lines(venv_dir)
+
         subprocess.run(
             " && ".join(venv_script),
             check=True,  # fail if any of the commands fail
@@ -181,7 +182,7 @@ class CachedExperimentVenv:
         )
         # Make sure the venv has been installed
         assert (venv_dir / ".installed").is_file()
-        cache.set(f"fawltydeps/{self.venv_hash()}", str(venv_dir))
+        cache.set(os.path.join("fawltydeps", f"{self.venv_hash()}"), str(venv_dir))
         return venv_dir
 
 
--- a/tests/project_helpers.py
+++ b/tests/project_helpers.py
@@ -102,7 +102,7 @@ class TarballPackage:
 
     @property
     def cache_key(self) -> str:
-        return os.path.join("fawltydeps", f"{self.tarball_name()}")
+        return str(Path("fawltydeps", self.tarball_name()))
 
     def tarball_path(self, cache: pytest.Cache) -> Path:
         return self.cache_dir(cache) / self.tarball_name()
@@ -186,7 +186,7 @@ class CachedExperimentVenv:
         script or the requirements to create that venv change.
         """
         # We cache venv dirs using the hash from create_venv_hash
-        cached_str = cache.get(os.path.join("fawltydeps", f"{self.venv_hash()}"), None)
+        cached_str = cache.get(str(Path("fawltydeps", self.venv_hash())), None)
         if cached_str is not None and Path(cached_str, ".installed").is_file():
             return Path(cached_str)  # already cached
 
@@ -202,7 +202,7 @@ class CachedExperimentVenv:
         )
         # Make sure the venv has been installed
         assert (venv_dir / ".installed").is_file()
-        cache.set(os.path.join("fawltydeps", f"{self.venv_hash()}"), str(venv_dir))
+        cache.set(str(Path("fawltydeps", self.venv_hash())), str(venv_dir))
         return venv_dir
 
 
--- a/tests/project_helpers.py
+++ b/tests/project_helpers.py
@@ -1,6 +1,7 @@
 """Common helpers shared between test_real_project and test_sample_projects."""
 import hashlib
 import logging
+import os
 import shlex
 import subprocess
 import sys
@@ -150,7 +151,7 @@ class CachedExperimentVenv:
         The Python version currently used to run the tests is used to compute
         the hash and create the venv.
         """
-        dummy_script = self.venv_script_lines(Path("/dev/null"))
+        dummy_script = self.venv_script_lines(Path(os.devnull))
         py_version = f"{sys.version_info.major},{sys.version_info.minor}"
         script_and_version_bytes = ("".join(dummy_script) + py_version).encode()
         return hashlib.sha256(script_and_version_bytes).hexdigest()
--- a/tests/project_helpers.py
+++ b/tests/project_helpers.py
@@ -2,7 +2,6 @@
 import hashlib
 import logging
 import os
-import platform
 import shlex
 import subprocess
 import sys
@@ -128,7 +127,7 @@ class CachedExperimentVenv:
     requirements: List[str]  # PEP 508 requirements, passed to 'pip install'
 
     def venv_script_lines(self, venv_path: Path) -> List[str]:
-        if platform.system() == "Windows":
+        if sys.platform.startswith("win"):
             pip_path = venv_path / "Scripts" / "pip.exe"
             python_path = venv_path / "Scripts" / "python.exe"
             return (
--- a/tests/project_helpers.py
+++ b/tests/project_helpers.py
@@ -296,7 +296,6 @@ class BaseExperiment(ABC):
 
     def get_venv_dir(self, cache: pytest.Cache) -> Path:
         """Get this venv's dir and create it if necessary."""
-        print(f"EXPERIMENT REQUIREMENTS: {self.requirements}")
         return CachedExperimentVenv(self.requirements)(cache)
 
 
--- a/tests/project_helpers.py
+++ b/tests/project_helpers.py
@@ -2,6 +2,7 @@
 import hashlib
 import logging
 import os
+import platform
 import shlex
 import subprocess
 import sys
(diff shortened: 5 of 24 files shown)


fawltydeps/extract_declared_dependencies.py
```diff
@@ -149,7 +149,7 @@ def parse_value(value: str) -> Iterator[DeclaredDependency]:
         # TODO: try leveraging RequirementsFile.from_string once
         #       pip-requirements-parser updates.
         # See:  https://github.com/nexB/pip-requirements-parser/pull/17
-        with NamedTemporaryFile(mode="wt") as tmp:
+        with NamedTemporaryFile(mode="wt", delete=False) as tmp:
```
jherland, line 152: I don't think we want to pass `delete=False` on non-Windows? Even on Windows, we should maybe try to find a solution that does not leave behind this temporary file?

I looked up [the docs](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) on `NamedTemporaryFile`, and it has a lot to say about how the `delete` flag (and the new `delete_on_close` flag added in v3.12) works, but none of it applied directly to us, I believe: We only close the file once (when exiting the `with`-block), and we certainly don't attempt to re-open it after closing (hence it does not matter if the file is deletes on close or deleted on context exit).

However, it also has this to say about deleting temporary files on Windows, which I believe might be related to our situation:

> On Windows, if `delete_on_close` is false, and the file is created in a directory for which the user lacks delete access, then the [`os.unlink()`](https://docs.python.org/3/library/os.html#os.unlink) call on exit of the context manager will fail with a [`PermissionError`](https://docs.python.org/3/library/exceptions.html#PermissionError). This cannot happen when `delete_on_close` is true because delete access is requested by the open, which fails immediately if the requested access is not granted.

AFAIU the problem we're trying to work around here is that even though we are allowed to _create_ the temporary file (and write into it), we somehow need Administrator access on Windows in order to delete it.

It would be nice to play around with Python 3.12 on Windows to see if the new `delete_on_close` flag can somehow solve the issue we're seeing here. Still, we wouldn't be able to use that here any time soon...

I noticed the `TODO` in the preceding comment, and tried to chase down the upstream issue to see if maybe we could sidestep the temporary file creation altogether. That rabbit hole led me to [this comment](https://github.com/nexB/pip-requirements-parser/pull/19#discussion_r1379279880), showing an alternative solution that together with `delete=False` also tries to manually remove up the temp file afterwards. I don't know if this would solve our problem. In any case that PR is still open upstream, so we'd have to replicate the same fix here rather that rely on the upstream solution...

fawltydeps/packages.py
```diff
@@ -345,6 +347,15 @@ def find_package_dirs(cls, path: Path) -> Iterator[Path]:
...
 
+        # Check for packages on Windows
+        if platform.system() == "Windows":
+            for site_packages in path.glob(os.path.join("Lib", "site-packages")):
+                if site_packages.is_dir():
+                    yield site_packages
+                    found = True
+            if found:
+                return
+
```
jherland, line 358: There are [many different ways](https://stackoverflow.com/questions/1325581/how-do-i-check-if-im-running-on-windows-in-python) of checking if we're running on Windows. I think I'm partial to using [`sys.platform`](https://docs.python.org/3/library/sys.html#sys.platform) rather than `platform.system()` for the following reasons:
- We're already importing `sys` ~everywhere, so fewer imports are needed
- `sys.platform` differentiates between `"win32"` and `"cygwin"` which is useful here (since Unix-style paths are expected in a cygwin environment).

Also, I notice there are no wildcards (`?`, `*`) passed to `path.glob()`. Maybe we don't need to call `glob` at all?

Finally, I'd like to locate the Windows site-packages check somewhat closer to the corresponding Unix-style check.

Maybe something like this (starting around line 332)?

```python
        found = False

        if sys.platform.startswith("win"):  # Windows
            site_packages = path / "Lib" / "site-packages"
            if site_packages.is_dir():
                yield site_packages
                found = True
        else:  # Assume POSIX
            if (path / "bin/python").is_file():
                for site_packages in path.glob("lib/python?.*/site-packages"):
                    if site_packages.is_dir():
                        yield site_packages
                        found = True
        if found:
            return
```

fawltydeps/packages.py
```diff
@@ -422,8 +433,14 @@ def installed_requirements(
             text=True,
             check=False,
         )
+        pip_path = (
+            venv_dir / "Scripts" / "pip.exe"
+            if platform.system() == "Windows"
+            else venv_dir / "bin" / "pip"
+        )
```
jherland, line 440: Sometimes the multiline `if-else` expression are easier to read as an old-fashioned `if ...: ... else:` statement:
```suggestion
        if sys.platform.startswith("win"):  # Windows
            pip_path = venv_dir / "Scripts" / "pip.exe"
        else:  # Assume POSIX
            pip_path = venv_dir / "bin" / "pip"
```

fawltydeps/types.py
```diff
@@ -155,6 +157,12 @@ def __post_init__(self) -> None:
         elif self.path.match("__pypackages__/?.*/lib"):
             return  # also ok
 
+        # Support Windows projects
+        if platform.system() == "Windows" and self.path.match(
+            os.path.join("Lib", "site-packages")
+        ):
+            return  # also ok
+
```
jherland, line 165: Same as in an earlier comment, I'd prefer to co-locate the Windows code with the corresponding POSIX code (from line 152):

```python
        # Support vitualenvs, poetry2nix envs, system-wide installs, etc.
        if sys.platform.startswiith("win"):
            if self.path.match("Lib\\site-packages"):
                return  # all ok
        else:
            if self.path.match("lib/python?.*/site-packages"):
                if (self.path.parent.parent.parent / "bin/python").is_file():
                    return  # all ok

        # Also support projects using __pypackages__ from PEP582:
        if self.path.match("__pypackages__/?.*/lib"):
            return  # also ok
```

tests/conftest.py
```diff
@@ -54,7 +55,13 @@ def create_one_fake_venv(
...
         # Create fake packages
         major, minor = py_version
-        site_dir = venv_dir / f"lib/python{major}.{minor}/site-packages"
+
+        def _env_site_packages():
+            if platform.system() == "Windows":
+                return venv_dir / "Lib" / "site-packages"
+            return venv_dir / "lib" / f"python{major}.{minor}" / "site-packages"
+
+        site_dir = _env_site_packages()
```
jherland, line 64: Isn't this easier?:

```suggestion
        if sys.platform.startswiith("win"):
            site_dir = venv_dir / "Lib" / "site-packages"
        else:
            major, minor = py_version
            site_dir = venv_dir / "lib" / f"python{major}.{minor}" / "site-packages"
```

tests/project_helpers.py
```diff
@@ -100,7 +102,7 @@ def get(self, cache: pytest.Cache) -> Path:
 
     @property
     def cache_key(self) -> str:
-        return f"fawltydeps/{self.tarball_name()}"
+        return os.path.join("fawltydeps", f"{self.tarball_name()}")
```
jherland, line 105: With `pathlib`, I'm trying to deprecate `os.path` as much as possible... 😅 

```suggestion
        return Path("fawltydeps", self.tarball_name())
```
plus `from pathlib import Path` at the top.

tests/project_helpers.py
```diff
@@ -165,22 +186,23 @@ def __call__(self, cache: pytest.Cache) -> Path:
         script or the requirements to create that venv change.
         """
         # We cache venv dirs using the hash from create_venv_hash
-        cached_str = cache.get(f"fawltydeps/{self.venv_hash()}", None)
+        cached_str = cache.get(os.path.join("fawltydeps", f"{self.venv_hash()}"), None)
```
jherland, line 189:
```suggestion
        cached_str = cache.get(Path("fawltydeps", self.venv_hash()), None)
```

tests/project_helpers.py
```diff
@@ -165,22 +186,23 @@ def __call__(self, cache: pytest.Cache) -> Path:
...
+
         subprocess.run(
             " && ".join(venv_script),
             check=True,  # fail if any of the commands fail
             shell=True,  # pass multiple shell commands to the subprocess
         )
         # Make sure the venv has been installed
         assert (venv_dir / ".installed").is_file()
-        cache.set(f"fawltydeps/{self.venv_hash()}", str(venv_dir))
+        cache.set(os.path.join("fawltydeps", f"{self.venv_hash()}"), str(venv_dir))
```
jherland, line 205:
```suggestion
        cache.set(str(Path("fawltydeps", self.venv_hash())), str(venv_dir))
```

fawltydeps/packages.py
jherland: BTW, I notice further below that a Windows virtualenv contains a
```
python_path = venv_path / "Scripts" / "python.exe"
```
Do we want to check for this, just like we're checking `if (path / "bin/python").is_file():` in the POSIX case?

tests/project_helpers.py
```diff
@@ -126,18 +128,37 @@ class CachedExperimentVenv:
...
+                + [
+                    f'{python_path} -m pip install --no-deps "{req}"'
+                    for req in self.requirements
+                ]
+                + [
+                    f"type nul > {venv_path / '.installed'}",
+                ]
+            )
+
+        pip_path = venv_path / "bin" / "pip"
```
jherland, line 149: Since we're adding a section for Windows here that returns on its own, I guess we can leave the POSIX arm below unchanged? In other words, make the whole function this?:

```python
    def venv_script_lines(self, venv_path: Path) -> List[str]:
        if sys.platform.startswith("win"):  # Windows
            python_path = venv_path / "Scripts" / "python.exe"
            return (
                [
                    f"rd /s /q {venv_path}",
                    f"{sys.executable} -m venv {venv_path}",
                    f"{python_path} -m pip install --upgrade pip",
                ]
                + [
                    f'{python_path} -m pip install --no-deps "{req}"'
                    for req in self.requirements
                ]
                + [
                    f"type nul > {venv_path / '.installed'}",
                ]
            )
        else:  # Assume POSIX
            return (
                [
                    f"rm -rf {venv_path}",
                    f"python3 -m venv {venv_path}",
                    f"{venv_path}/bin/pip install --upgrade pip",
                ]
                + [
                    f"{venv_path}/bin/pip install --no-deps {shlex.quote(req)}"
                    for req in self.requirements
                ]
                + [
                    f"touch {venv_path}/.installed",
                ]
            )
```

But as soon as I suggest that, I'm starting to see some more cracks forming here:

Taking a step back and examining what this function is actually used for: This is not just returning an awkward shell script to be passed to `subprocess.run(" && ".join(venv_script), ...)` below. It is _also_ used to create the string we return from `venv_hash()`, and this string should remain stable as long as the Python version, the `venv_path` (artificially fixed) and the script itself does not change. However, we're now introducing `sys.executable` into this script, and I don't know if this is always stable between successive CI runs (a CI runner could have the same Python installed in a different location).

Still, I'm not sure it's worth doing too much about this: The failure mode would be for the hash to update needlessly which would cause Windows tests to not benefit from the caching we're implementing here. Probably not too bad, all things considered.

**To be clear: Since we're disabling `real_projects` on Windows for now, anyway, I'm happy for you to push all of these related suggestions to a later issue/PR.** 😄

---

...because there are also some other minor things that I would like to change:
- We should be running `{venv_path}/bin/python -m pip install ...` instead of `{venv_path}/bin/pip install ...` in the POSIX version as well.
- We should be quoting paths in case they contain spaces (paths with spaces are certainly more common on Windows)
- I also dislike the clumsy concatenation of lists above, and wonder if we could just `yield` instead...

Here's an attempt to rewrite this function into something that is hopefully more robust, while keeping the overall design (which arguably should be rewritten altogether, at some point):

```python
    def venv_script_lines(
        self, venv_path: Path, python_exe: Optional[str] = None
    ) -> Iterator[str]:
        if python_exe is None:  # Default to current Python executable
            python_exe = sys.executable

        if sys.platform.startswith("win"):  # Windows
            remove_cmd = "rd /s /q"
            touch_file_cmd = "type nul >"
            venv_python = venv_path / "Scripts" / "python.exe"
        else:  # Assume POSIX
            remove_cmd = "rm -rf"
            touch_file_cmd = f"touch"
            venv_python = venv_path / "bin" / "python"

        def sh_quote(arg: Union[Path, str]) -> str:
            if sys.platform.startswith("win"):  # Windows
                return f'"{arg}"'
            else:  # Assume POSIX
                return shlex.quote(str(arg))

        yield f"{remove_cmd} {sh_quote(venv_path)}"
        yield f"{sh_quote(python_exe)} -m venv {sh_quote(venv_path)}"
        yield f"{sh_quote(venv_python)} -m pip install --upgrade pip"
        for req in self.requirements:
            yield f"{sh_quote(venv_python)} -m pip install --no-deps {sh_quote(req)}"
        yield f"{touch_file_cmd} {sh_quote(venv_path / '.installed')}"
```

tests/project_helpers.py
```diff
@@ -150,7 +171,7 @@ def venv_hash(self) -> str:
         The Python version currently used to run the tests is used to compute
         the hash and create the venv.
         """
-        dummy_script = self.venv_script_lines(Path("/dev/null"))
+        dummy_script = self.venv_script_lines(Path(os.devnull))
```
jherland, line 174: With the above rewrite this should become:

```suggestion
        # Fake venv_path and python_exe to avoid unnecessary cache misses
        dummy_script = list(self.venv_script_lines(Path(os.devnull), "python"))
```

tests/project_helpers.py
```diff
@@ -165,22 +186,23 @@ def __call__(self, cache: pytest.Cache) -> Path:
...
-        cached_str = cache.get(f"fawltydeps/{self.venv_hash()}", None)
+        cached_str = cache.get(os.path.join("fawltydeps", f"{self.venv_hash()}"), None)
         if cached_str is not None and Path(cached_str, ".installed").is_file():
             return Path(cached_str)  # already cached
 
         # Must run the script to set up the venv
         venv_dir = Path(cache.mkdir(f"fawltydeps_venv_{self.venv_hash()}"))
         logger.info(f"Creating venv at {venv_dir}...")
         venv_script = self.venv_script_lines(venv_dir)
+
```
jherland, line 197: With the rewritten `venv_script_lines()`, this becomes:
```suggestion
        venv_script = list(self.venv_script_lines(venv_dir))

```

tests/project_helpers.py
```diff
@@ -275,6 +297,7 @@ def from_toml(cls, name: str, data: TomlData) -> "BaseExperiment":
 
     def get_venv_dir(self, cache: pytest.Cache) -> Path:
         """Get this venv's dir and create it if necessary."""
+        print(f"EXPERIMENT REQUIREMENTS: {self.requirements}")
```
jherland, line 300: Intentional, or temporary debug to be removed?

tests/test_cmdline.py
```diff
@@ -429,8 +437,8 @@ def test_list_sources__in_varied_project__lists_all_files(fake_project):
     tmp_path = fake_project(
         files_with_imports={
             "code.py": ["foo"],
-            "subdir/other.py": ["foo"],
-            "subdir/notebook.ipynb": ["foo"],
+            os.path.join("subdir", "other.py"): ["foo"],
+            os.path.join("subdir", "notebook.ipynb"): ["foo"],
```
jherland, line 441: Really only a matter of taste, but since `pathlib` has become available, I've grown allergic to `os.path` 😅 

```suggestion
            str(Path("subdir", "other.py")): ["foo"],
            str(Path("subdir", "notebook.ipynb")): ["foo"],
```

tests/test_cmdline.py
```diff
@@ -446,13 +454,15 @@ def test_list_sources__in_varied_project__lists_all_files(fake_project):
...
+            os.path.join("subdir", "other.py"),
+            os.path.join("subdir", "notebook.ipynb"),
             "requirements.txt",
             "pyproject.toml",
             "setup.py",
             "setup.cfg",
-            f"my_venv/lib/python{major}.{minor}/site-packages",
+            os.path.join("my_venv", "Lib", "site-packages")
+            if platform.system() == "Windows"
+            else f"my_venv/lib/python{major}.{minor}/site-packages",
```
jherland, line 465: I think I find it easier to read when the Windows-condition is not buried inside an if-else expression inside a list comprehension:

```python
    output, returncode = run_fawltydeps_function("--list-sources", f"{tmp_path}")

    if sys.platform.startswith("win"):  # Windows
        site_packages = Path("my_venv", "Lib", "site-packages")
    else:  # Assume POSIX
        major, minor = sys.version_info[:2]
        site_packages = Path("my_venv", "lib", f"python{major}.{minor}", "site-packages")
    expect = [
        str(tmp_path / filename)
        for filename in [
            "code.py",
            os.path.join("subdir", "other.py"),
            os.path.join("subdir", "notebook.ipynb"),
            "requirements.txt",
            "pyproject.toml",
            "setup.py",
            "setup.cfg",
            str(site_packages),
        ]
    ]
    assert_unordered_equivalence(output.splitlines()[:-2], expect)
    assert returncode == 0
```

tests/test_cmdline.py
```diff
@@ -475,19 +485,19 @@ def test_list_sources_detailed__in_varied_project__lists_all_files(fake_project)
...
             "setup.py",  # This is both a CodeSource and an DepsSource!
-            "subdir/notebook.ipynb",
-            "subdir/other.py",
+            os.path.join("subdir", "notebook.ipynb"),
+            os.path.join("subdir", "other.py"),
         ]
     ]
     expect_deps_lines = [
-        f"  {tmp_path / filename} (parsed as a {filename} file)"
+        f"  { tmp_path / filename} (parsed as a {filename} file)"
```
jherland, line 500: Intentional?

tests/test_cmdline.py
```diff
@@ -497,8 +507,19 @@ def test_list_sources_detailed__in_varied_project__lists_all_files(fake_project)
...
+            else "  "
+            + str(
+                tmp_path
+                / "my_venv"
+                / "lib"
+                / f"python{major}.{minor}"
+                / "site-packages"
+            )
+        )
+        + " (as a source of Python packages)",
```
jherland, line 522: Ok, time to factor our a helper function, I think:

```python
def site_packages(venv_dir: Path) -> Path:
    if sys.platform.startswith("win"):  # Windows
        return venv_dir / "Lib" / "site-packages"
    else:  # Assume POSIX
        major, minor = sys.version_info[:2]
        return venv_dir / f"lib/python{major}.{minor}/site-packages"
```
Put this near the top of the file (or in `tests/utils.py` to share across multiple tests), and then:

```suggestion
        f"  {site_packages(tmp_path / 'my_venv')} (as a source of Python packages)",
```

tests/test_cmdline.py
```diff
@@ -520,11 +541,18 @@ def test_list_sources_detailed__from_both_python_file_and_stdin(fake_project):
...
+            "  <stdin>",
+        ],
+        [
+            "Sources of Python code:",
+            "  <stdin>",
+            f"  {tmp_path / 'code.py'} (using {tmp_path} as base for 1st-party imports)",
+        ],
     ]
-    assert output.splitlines() == expect
+    assert output.splitlines() == expect[0] or output.splitlines() == expect[1]
```
jherland, line 555:
```suggestion
    assert output.splitlines() in expect
```

tests/test_cmdline.py
```diff
@@ -562,7 +590,7 @@ class ProjectTestVector:
         expect_output=[
             "These imports appear to be undeclared dependencies:",
             "- 'requests' imported at:",
-            "    {path}/code.py:1",
+            f"    {os.path.join('{path}', 'code.py')}:1",
```
jherland, line 593:
```suggestion
            f"    {Path('{path}', 'code.py')}:1",
```

tests/test_cmdline.py
```diff
@@ -574,7 +602,7 @@ class ProjectTestVector:
         expect_output=[
             "These dependencies appear to be unused (i.e. not imported):",
             "- 'pandas' declared in:",
-            "    {path}/requirements.txt",
+            f"    {os.path.join('{path}', 'requirements.txt')}",
```
jherland, line 605:
```suggestion
            f"    {Path('{path}', 'requirements.txt')}",
```

tests/test_cmdline.py
```diff
@@ -586,11 +614,11 @@ class ProjectTestVector:
         expect_output=[
             "These imports appear to be undeclared dependencies:",
             "- 'requests' imported at:",
-            "    {path}/code.py:1",
+            f"    {os.path.join('{path}', 'code.py')}:1",
```
jherland, line 617:
```suggestion
            f"    {Path('{path}', 'code.py')}:1",
```

tests/test_cmdline.py
```diff
@@ -586,11 +614,11 @@ class ProjectTestVector:
...
             "",
             "These dependencies appear to be unused (i.e. not imported):",
             "- 'pandas' declared in:",
-            "    {path}/requirements.txt",
+            f"    {os.path.join('{path}', 'requirements.txt')}",
```
jherland, line 621:
```suggestion
            f"    {Path('{path}', 'requirements.txt')}",
```

tests/test_cmdline.py
```diff
@@ -610,7 +638,8 @@ class ProjectTestVector:
         ],
         expect_logs=[
             "INFO:fawltydeps.extract_imports:Finding Python files under {path}",
-            "INFO:fawltydeps.extract_imports:Parsing Python file {path}/code.py",
+            "INFO:fawltydeps.extract_imports:Parsing Python file "
+            f"{os.path.join('{path}', 'code.py')}",
```
jherland, line 642:
```suggestion
            f"{Path('{path}', 'code.py')}",
```

tests/test_cmdline.py
```diff
@@ -624,11 +653,11 @@ class ProjectTestVector:
         expect_output=[
             "These imports appear to be undeclared dependencies:",
             "- 'requests' imported at:",
-            "    {path}/code.py:1",
+            f"    {os.path.join('{path}', 'code.py')}:1",
```
jherland, line 656:
```suggestion
            f"    {Path('{path}', 'code.py')}:1",
```

tests/test_cmdline.py
```diff
@@ -624,11 +653,11 @@ class ProjectTestVector:
...
             "",
             "These dependencies appear to be unused (i.e. not imported):",
             "- 'pandas' declared in:",
-            "    {path}/requirements.txt",
+            f"    {os.path.join('{path}', 'requirements.txt')}",
```
jherland, line 660:
```suggestion
            f"    {Path('{path}', 'requirements.txt')}",
```

tests/test_cmdline.py
```diff
@@ -678,29 +707,33 @@ def test_check_json__simple_project__can_report_both_undeclared_and_unused(
...
                 "parser_choice": "requirements.txt",
             },
             {
                 "source_type": "PyEnvSource",
-                "path": f"{tmp_path}/my_venv/lib/python{major}.{minor}/site-packages",
+                "path": (
+                    f"{tmp_path / 'my_venv' / 'Lib' / 'site-packages'}"
+                    if platform.system() == "Windows"
+                    else f"{tmp_path}/my_venv/lib/python{major}.{minor}/site-packages"
+                ),
```
jherland, line 724: With the `site_packages()` helper I suggested earlier:

```suggestion
                "path": f"{site_packages(tmp_path / 'my_venv')}",
```

tests/test_deps_parser_determination.py
```diff
@@ -29,20 +30,20 @@
...
+            (os.path.join("sub", "anything_else"), None),
+            (os.path.join("abs", "requirements.txt"), ParserChoice.REQUIREMENTS_TXT),
+            (os.path.join("abs", "setup.py"), ParserChoice.SETUP_PY),
+            (os.path.join("abs", "setup.cfg"), ParserChoice.SETUP_CFG),
+            (os.path.join("abs", "pyproject.toml"), ParserChoice.PYPROJECT_TOML),
+            (os.path.join("abs", "anything_else"), None),
+            (os.path.join("requirements.txt", "wat"), None),
+            (os.path.join("setup.py", "wat"), None),
+            (os.path.join("setup.cfg", "wat"), None),
+            (os.path.join("pyproject.toml", "wat"), None),
```
jherland, line 46:
```suggestion
            (str(Path("sub", "requirements.txt")), ParserChoice.REQUIREMENTS_TXT),
            (str(Path("sub", "setup.py")), ParserChoice.SETUP_PY),
            (str(Path("sub", "setup.cfg")), ParserChoice.SETUP_CFG),
            (str(Path("sub", "pyproject.toml")), ParserChoice.PYPROJECT_TOML),
            (str(Path("sub", "anything_else")), None),
            (str(Path("abs", "requirements.txt")), ParserChoice.REQUIREMENTS_TXT),
            (str(Path("abs", "setup.py")), ParserChoice.SETUP_PY),
            (str(Path("abs", "setup.cfg")), ParserChoice.SETUP_CFG),
            (str(Path("abs", "pyproject.toml")), ParserChoice.PYPROJECT_TOML),
            (str(Path("abs", "anything_else")), None),
            (str(Path("requirements.txt", "wat")), None),
            (str(Path("setup.py", "wat")), None),
            (str(Path("setup.cfg", "wat")), None),
            (str(Path("pyproject.toml", "wat")), None),
```

tests/test_dir_traversal.py
```diff
@@ -166,9 +168,11 @@ class DirectoryTraversalVector(Generic[T]):
...
             ExpectedTraverseStep(".", subdirs=["a"]),
             ExpectedTraverseStep("a", subdirs=["b"], attached=[456]),
-            ExpectedTraverseStep("a/b", subdirs=["c"], attached=[456]),
-            ExpectedTraverseStep("a/b/c", subdirs=["d"], attached=[456, 123]),
-            ExpectedTraverseStep("a/b/c/d", attached=[456, 123]),
+            ExpectedTraverseStep(os.path.join("a", "b"), subdirs=["c"], attached=[456]),
+            ExpectedTraverseStep(
+                os.path.join("a", "b", "c"), subdirs=["d"], attached=[456, 123]
+            ),
+            ExpectedTraverseStep(os.path.join("a", "b", "c", "d"), attached=[456, 123]),
```
jherland, line 175:
```suggestion
            ExpectedTraverseStep(str(Path("a", "b")), subdirs=["c"], attached=[456]),
            ExpectedTraverseStep(
                str(Path("a", "b", "c")), subdirs=["d"], attached=[456, 123]
            ),
            ExpectedTraverseStep(str(Path("a", "b", "c", "d")), attached=[456, 123]),
```

tests/test_dir_traversal.py
```diff
@@ -261,7 +273,22 @@ class DirectoryTraversalVector(Generic[T]):
...
+            v,
+            id=v.id,
+            marks=pytest.mark.skipif(
+                platform.system() == "Windows"
+                and any(
+                    isinstance(entry, (RelativeSymlink, AbsoluteSymlink))
+                    for entry in v.given
+                ),
+                reason="Symlinks on Windows may be created only by administrators",
+            ),
```
jherland, line 288: Nice! This condition is quite complex, but I'm happy with how readable this still is.

tests/test_local_env.py
```diff
@@ -128,7 +164,11 @@ def test_local_env__default_venv__contains_pip(tmp_path):
     # "pip" package is installed, and that it provides a "pip" import name.
     venv.create(tmp_path, with_pip=True)
     lpl = LocalPackageResolver(pyenv_sources(tmp_path))
-    expect_location = tmp_path / f"lib/python{major}.{minor}/site-packages"
+    expect_location = (
+        tmp_path / "Lib" / "site-packages"
+        if platform.system() == "Windows"
+        else tmp_path / f"lib/python{major}.{minor}/site-packages"
+    )
```
jherland, line 171: Yet another use case for the `site_packages()` helper:

```suggestion
    expect_location = site_packages(tmp_path)
```

tests/test_local_env.py
```diff
@@ -37,7 +39,32 @@
...
+# that they might point at (and that we should accept)?
+windows_subdirs = [
+    "",
+    "Lib",
+    os.path.join("Lib", "site-packages"),
+]
 
+
+@pytest.mark.skipif(
+    platform.system() != "Windows", reason="Not relevant to Windows virtual environment"
```
jherland, line 53:
```suggestion
    platform.system() != "Windows", reason="Only relevant to Windows virtual environment"
```

tests/test_local_env.py
```diff
@@ -48,6 +75,9 @@ def test_find_package_dirs__various_paths_in_venv(tmp_path, subdir):
     assert set(LocalPackageResolver.find_package_dirs(path)) == expect
 
 
+@pytest.mark.skipif(
+    platform.system() == "Windows", reason="Not relevant to Windows virtual environment"
```
jherland, line 79:
```suggestion
    sys.platform.startswith("win"), reason="Not relevant to Windows virtual environment"
```

tests/test_local_env.py
```diff
@@ -64,6 +94,9 @@ def test_find_package_dirs__various_paths_in_poetry2nix_env(write_tmp_files, sub
...
```
jherland, line 98:
```suggestion
    sys.platform.startswith("win"), reason="Not relevant to Windows virtual environment"
```

tests/test_local_env.py
```diff
@@ -79,6 +112,9 @@ def test_find_package_dirs__various_paths_in_pypackages(write_tmp_files, subdir)
...
```
jherland, line 116:
```suggestion
    sys.platform.startswith("win"), reason="Not relevant to Windows virtual environment"
```

tests/test_real_projects.py
```diff
@@ -183,6 +187,7 @@ def test_real_project(request, project, experiment):
 
     print(f"Testing real project {project.name!r} under {project_dir}")
     print(f"Project description: {project.description}")
+    print(f"Virtual environment of the project {project.name!r} under {venv_dir}")
```
jherland, line 190: The project name is already mentioned two lines above.

```suggestion
    print(f"Virtual environment under {venv_dir}")
```

tests/test_real_projects.py
```diff
@@ -175,6 +178,7 @@ def unpacked_project_dir(self, cache: pytest.Cache) -> Path:
         pytest.param(project, experiment, id=experiment.name)
         for project in ThirdPartyProject.collect()
         for experiment in project.experiments
+        if not platform.system() == "Windows"
```
jherland, line 181:
```suggestion
        if not sys.platform.startswith("win")
```

tests/test_resolver.py
```diff
@@ -127,7 +128,7 @@ def generate_expected_resolved_deps(
 )
 @settings(
     suppress_health_check=[HealthCheck.function_scoped_fixture],
-    deadline=5000,
+    deadline=None,
```
jherland, line 131: Argh 😅 It's a shame that we cannot access the `install_deps` parameter here, and them simply do:

```python
    deadline=None if install_deps else 5000,
```

tests/test_settings.py
```diff
@@ -477,7 +478,12 @@ def test_settings__missing_config_file__uses_defaults_and_warns(tmp_path, caplog
     settings = Settings.config(config_file=missing_file)()
     assert settings.dict() == make_settings_dict()
     assert "Failed to load configuration file:" in caplog.text
-    assert str(missing_file) in caplog.text
+    # On Windows path escape sequences are doubled in repr()
+    assert (
+        repr(str(missing_file))
+        if platform.system() == "Windows"
+        else str(missing_file) in caplog.text
+    )
```
jherland, line 486: It's hard to see that the `... in caplog.text` on the last line also applies to the Windows-side of the if-else condition. What about this?

```suggestion
    missing_file_str = str(missing_file)
    # On Windows path escape sequences are doubled in repr()
    if sys.platform.startswith("win"):
        missing_file_str = repr(missing_file_str)
    assert missing_file_str in caplog.text
```

tests/test_types.py
```diff
@@ -10,19 +11,31 @@
...
+    ),
     "rel_path_cell_line": ((Path("foo"), 3, 4), "foo[3]:4", 322),
     # Cell number is omitted for non-notebooks.
     "no_path_line": (("<stdin>", None, 2), "<stdin>:2", 112),
-    "abs_path_line": ((Path("/foo/bar"), None, 3), "/foo/bar:3", 212),
+    "abs_path_line": (
+        (Path("/foo/bar"), None, 3),
+        os.path.join(os.sep, "foo", "bar:3"),
+        212,
+    ),
```
jherland, line 38: Hmm, I wonder if these tests require so rethinking? I guess absolute paths like `\foo\bar` makes sense on Windows? But are `os.path.join(os.sep, ...)` the appropriate way to create such paths? And should we also test paths with a drive prefix (protected by appropriate `pytest.skipif`)?

jherland (commented):
First and foremost: I'm really impressed with this PR! 🤩 Although I am surprised that people are already using FawltyDeps on Windows with some success, it is obvious that we still have quite a lot of POSIX assumptions in our code, especially under `tests/`, and I'm very happy with how you have tackled this: Your PR systematically fixes everything across our (by now, quite large) codebase.

I've done a first complete pass over the PR, and added (too many, probably 😅) comments. I wasn't sure if I'd get through everything today, so I posted them as I went along. Apologies for not batching them into a single review.

Despite the number of comments, I really don't have any issues with the approach you've taken here. Most of the comments are minor niggles like using `sys.platform` instead of `platform.system` or `Path()` instead of `os.path.join()`.

One thing that is becoming really apparent in this PR is that our `real_projects` tests need some more serious rework to work well on Windows. I think disabling these on Windows is the right way to solve this, for now. Tackling that rewrite is best done in a separate PR.

Up to you, whether you want to clean up this PR by removing the other `real_projects`-related changes, or leave them in. I'm happy either way.

tests/test_cmdline.py
mknorps: No, change removed.

tests/test_cmdline.py
mknorps: Updated in 127473a053bfcf991c9a92f78a3e440f6d84cd0e

tests/test_cmdline.py
mknorps: Updated in 127473a053bfcf991c9a92f78a3e440f6d84cd0e

tests/test_cmdline.py
mknorps: Updated in 127473a053bfcf991c9a92f78a3e440f6d84cd0e

tests/conftest.py
mknorps: We can use `site_packages` function defined in `utils` for that.

tests/test_local_env.py
mknorps: Done in 6c094502c50bb5f789ccb3dd44ceb9fd934c67a2

tests/test_real_projects.py
mknorps: Changes to be included in `skipif`:
```

@pytest.mark.skipif(
    sys.platform.startswith("win"),
    reason="Real projects test are not supported on Windows due to the test environment complications.",
)
@pytest.mark.parametrize(
    "project, experiment",
```

tests/test_resolver.py
mknorps: True :(

tests/test_types.py
mknorps: > I guess absolute paths like \foo\bar makes sense on Windows? 

Yes. Both `\MyDir` and `C:\MyDir` make sense.

> But are os.path.join(os.sep, ...) the appropriate way to create such paths? 

Without `os.sep` we cannot access disc or system-level files, which work both for Windows and Linux systems.

>And should we also test paths with a drive prefix (protected by appropriate `pytest.skipif`)?

That is a good idea. I added it in 6df3f252ce58d29b1664c875438029b55815bc28

mknorps:
> First and foremost: I'm really impressed with this PR! 🤩 Although I am surprised that people are already using FawltyDeps on Windows with some success, it is obvious that we still have quite a lot of POSIX assumptions in our code, especially under `tests/`, and I'm very happy with how you have tackled this: Your PR systematically fixes everything across our (by now, quite large) codebase.
> 
> I've done a first complete pass over the PR, and added (too many, probably 😅) comments. I wasn't sure if I'd get through everything today, so I posted them as I went along. Apologies for not batching them into a single review.
> 
> Despite the number of comments, I really don't have any issues with the approach you've taken here. Most of the comments are minor niggles like using `sys.platform` instead of `platform.system` or `Path()` instead of `os.path.join()`.
> 
> One thing that is becoming really apparent in this PR is that our `real_projects` tests need some more serious rework to work well on Windows. I think disabling these on Windows is the right way to solve this, for now. Tackling that rewrite is best done in a separate PR.
> 
> Up to you, whether you want to clean up this PR by removing the other `real_projects`-related changes, or leave them in. I'm happy either way.

Than you @jherland for your review and insights.

I fixed two main things you asked for - using `Path` instead of `os.path.join` and `sys.platform` instead of `platform.system`.

I would leave the real-project related challenges to another PR. I tested the Python Algorithms project that was failing in the real project tests on a Windows machine and it works.

For Windows users, I think that the most important is that when FawltyDeps looks for a mapping, it checks also virtual envi9ronment available in Windows. That si the most significant improvement in my opinion for a regular user. For a developer, we made it possible with this PR to contribute to FawltyDeps when writing code on Windows thanks to the tests refactor :tada:

tests/test_types.py
jherland: I'm not 100% happy with how the second component of these test vectors (the string representation we expect to be produced for a given `Location` object) now depends on the output of `os.path.join()`, especially when the only thing we really want/expect it to do is the `/` -> `\` transformation. When I look at these test vectors now, it almost looks like we're testing `Path("foo/bar")` against `os.path.join("foo", "bar")`, which is not the point of these `Location` tests...

Another consideration is that we're passing extra ("almost special") characters to `os.path.join()` with statements like `os.path.join(os.sep, "foo", "bar[2]:3")`. The `[2]:3` suffix is not really part of the filename, and I have to _assume_ that `os.path.join()` passes them along unchanged (I would need to look up allowable characters in Windows filenames, and the implementation of `os.path.join()` to know for sure...)

In sum I think we're better off dropping `os.path.join()` in these test vectors, and instead doing something simpler like this in the test below:

```python
def test_location__str(args, string, _):
    assert str(Location(*args)) == string.replace("/", os.sep)
```

jherland (approved):
Still a few comments here and there, but I am happy for this to be merged after you've had a look. Feel free to push back if you feel I'm getting too nitpicky. You can always tell me to fix things myself in a later PR 😅

There's one unresolved comment from the previous round, about the `delete` flag to `NamedTemporaryFile` where I think we need to be careful to get things right. Otherwise, my comments and suggestions are mostly just matters of taste, and I'm happy to get this merged with or without addressing those.

tests/project_helpers.py
jherland: Resolved here, to be addressed in #406.

tests/test_cmdline.py
jherland: Already fixed by using `site_packages()` helper.

tests/test_types.py
mknorps: I see your point. I changed the expected results to strings and used `tests_location__str` function as you proposed. Changes done in 5932b06faafcef4992d4469714805396634efbe9.

fawltydeps/extract_declared_dependencies.py
mknorps: > It would be nice to play around with Python 3.12 on Windows to see if the new delete_on_close flag can somehow solve the issue we're seeing here. Still, we wouldn't be able to use that here any time soon...

Yes, that is why I wil leave this idea for now, as we need to support older versions of Python. It will be definitely a neat solution.

I replaced:
```python
        with NamedTemporaryFile(mode="wt", delete=False) as tmp:
            tmp.write(value)
            tmp.flush()
            for dep in parse_requirements_txt(Path(tmp.name)):
                yield replace(dep, source=source)
```
with:
```python
        # https://github.com/nexB/pip-requirements-parser/pull/19#discussion_r1379279880
        tf = NamedTemporaryFile('wt', delete=False,
                        # we prefer utf8 encoded strings, but ...
                        # - must not change newlines
                        # - must not  change encoding, fallback to system encoding for compatibility
                        newline='', encoding=None)
        tf.write(value)
        tf.close()
        try:
            for dep in parse_requirements_txt(Path(tf.name)):
                yield replace(dep, source=source)
        finally:
            unlink(Path(tf.name))
```
(see 0eb7945ba7fa1f09794561748fe99f49279bc9d3 )

fawltydeps/packages.py
```diff
@@ -345,6 +346,15 @@ def find_package_dirs(cls, path: Path) -> Iterator[Path]:
...
 
+        # Check for packages on Windows
+        if sys.platform.startswith("win"):
+            for site_packages in path.glob(os.path.join("Lib", "site-packages")):
+                if site_packages.is_dir():
+                    yield site_packages
+                    found = True
+            if found:
+                return
+
```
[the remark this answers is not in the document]
mknorps: That is a good nitpick, thank you!
I updated the code accordingly in a5fb06736edf091ec1c657926886d310dda005d2.

fawltydeps/packages.py
jherland: Fixed in https://github.com/tweag/FawltyDeps/commit/a5fb06736edf091ec1c657926886d310dda005d2.

tests/project_helpers.py
mknorps: Removed.

tests/test_dir_traversal.py
```diff
@@ -179,8 +183,8 @@ class DirectoryTraversalVector(Generic[T]):
     DirectoryTraversalVector(
         "symlinks_to_parent__are_not_traversed",
         given=[
-            RelativeSymlink("sub/rel_parent", ".."),
-            AbsoluteSymlink("sub/abs_parent", "."),
+            RelativeSymlink(os.path.join("sub", "rel_parent"), ".."),
+            AbsoluteSymlink(os.path.join("sub", "abs_parent"), "."),
```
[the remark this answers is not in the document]
mknorps: > We'd need something similar for ExpectedTraverseStep and other helpers in order to minimize the diff in this part of the file. Not sure it's worth it...

Especially that we skip all tests that contain Symlinks on Windows due to access rights.
I will change all `os.path.join` to `str(Path())`.

fawltydeps/types.py
```diff
@@ -155,6 +156,12 @@ def __post_init__(self) -> None:
         elif self.path.match("__pypackages__/?.*/lib"):
             return  # also ok
 
+        # Support Windows projects
+        if sys.platform.startswith("win") and self.path.match(
+            os.path.join("Lib", "site-packages")
+        ):
+            return  # also ok
+
```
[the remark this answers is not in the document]
mknorps: Done in 83ea1f2045571be524c51cf1f814bc89cc3d308b.

fawltydeps/types.py
mknorps: Applied in e16b180234eb87b7d44db2dc2e822b7a92d6cc1e.

PR #399, opened 2023-11-22
PyPI Analysis: identify main code directories 

zz1874:
In this PR, we modified the current code of identifying main code directories by the following points:
- return all directories that contains python code files (`.py` or `.ipynb`) instead of the maximum one
  - in case that there are multiple packages/modules in one repository
- exclude all files that contains `test` rather than starts with it
- adjust summary/detailed/json outputs rendering

### DIFF
--- a/PyPI_analysis/main.py
+++ b/PyPI_analysis/main.py
@@ -113,7 +113,7 @@ class Analysis:  # pylint: disable=too-many-instance-attributes
         directories = [path.parts[0] for path in code_paths]
         directory_counts = Counter(directories)
         if directory_counts:
-            return dict(directory_counts)
+            return dict(directory_counts.most_common())
 
     @classmethod
     def create(
--- a/PyPI_analysis/main.py
+++ b/PyPI_analysis/main.py
@@ -173,14 +173,14 @@ class Analysis:  # pylint: disable=too-many-instance-attributes
             if detailed:
                 yield "Code directories: "
                 if self.code_dirs:
-                    for code_dir in self.code_dirs.items():
-                        yield f"  {code_dir[0]}: {code_dir[1]} Python files"
+                    for code_dir, count in self.code_dirs.items():
+                        yield f"  {code_dir}: {count} Python files"
                 else:
                     yield "  There is no main code directory found under the current directory."
             else:
                 if self.code_dirs:
-                    for code_dir in self.code_dirs.items():
-                        yield code_dir[0]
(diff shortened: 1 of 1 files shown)


PyPI_analysis/main.py
```diff
@@ -178,14 +171,16 @@ def print_human_readable(self, out: TextIO, detailed: bool = True) -> None:
...
         def render_code_directory() -> Iterator[str]:
             if detailed:
-                yield "Main code directory: "
-                if self.code_dir:
-                    yield f"  {self.code_dir['path']}: {self.code_dir['counts']} Python files"
+                yield "Code directories: "
+                if self.code_dirs:
+                    for code_dir in self.code_dirs.items():
+                        yield f"  {code_dir[0]}: {code_dir[1]} Python files"
                 else:
```
mknorps, line 178:
```suggestion
                    for code_dir, count in self.code_dirs.items():
                        yield f"  {code_dir}: {count} Python files"
                else:
```

PyPI_analysis/main.py
```diff
@@ -178,14 +171,16 @@ def print_human_readable(self, out: TextIO, detailed: bool = True) -> None:
...
                     yield "  There is no main code directory found under the current directory."
             else:
-                if self.code_dir:
-                    yield self.code_dir["path"]
+                if self.code_dirs:
+                    for code_dir in self.code_dirs.items():
+                        yield code_dir[0]
```
mknorps, line 183:
```suggestion
                    for code_dir, _count in self.code_dirs.items():
                        yield code_dir
```

PyPI_analysis/main.py
```diff
@@ -108,19 +108,12 @@ def code_dir(self) -> Optional[Dict[Path, int]]:
...
-        # Find the directory with the most Python files
         if directory_counts:
-            most_python_files_directory = max(
-                directory_counts, key=directory_counts.get
-            )
-            return {
-                "path": most_python_files_directory,
-                "counts": directory_counts[most_python_files_directory],
-            }
+            return dict(directory_counts)
```
mknorps, line 116: This will sort counter with respect to the highest number of source files found.
```suggestion
            return dict(directory_counts.most_common())
```

mknorps (approved):
I think this looks good - we will have more visibility of all directories in the project :rocket:

I added there few nit-picks (related to iteration over dict) and one more important suggestion - about the sorting of counter object.

Feel free to merge this PR after you address those.

PR #409, opened 2024-01-04
Loosen version constraint on core dependencies and re-pin lock file

### LINKED ISSUE #408: Stop pinning core dependencies

When I install falwtydeps on my project, it uninstalls setuptools and import-libmetadata in order to install older version

```
Installing collected packages: setuptools, pip-requirements-parser, importlib_metadata, fawltydeps
  Attempting uninstall: setuptools
    Found existing installation: setuptools 69.0.2
    Uninstalling setuptools-69.0.2:
      Successfully uninstalled setuptools-69.0.2
  Attempting uninstall: importlib_metadata
    Found existing installation: importlib-metadata 7.0.0
    Uninstalling importlib-metadata-7.0.0:
      Successfully uninstalled importlib-metadata-7.0.0
Successfully installed fawltydeps-0.13.3 importlib_metadata-6.6.0 pip-requirements-parser-32.0.1 setuptools-68.2.2

```

In theory, falwltydeps should work on newer version of its dependencies, so there's no reason to pin these to major versions. 

We should replace `^` with `>=` in pyproject.toml

```
[tool.poetry.dependencies]
python = ">=3.7.2"
importlib_metadata = ">=6.6.0"
isort = [
    # isort 5.12.0 drops support for Python v3.7:
    {version = "^5.10", python = ">=3.8"},
    {version = ">=5.10,<5.12.0", python = "<3.8"},
]
pip-requirements-parser = ">=32.0.1"
pydantic = ">=1.10.4,<3.0.0"
tomli = {version = "^2.0.1", python = "<3.11"}
typing-extensions = {version = ">=4.4.0", python = "<3.8"}
setuptools = ">=68.0.0"
```

I could install falwltydeps in a different virtual environment, or use pipx

jherland:
Fixes #408.

Use `>=` instead of `^` in pyproject.toml to not be too restrictive as
to which version of our dependencies is acceptable when a user installs
FawltyDeps into their project's venv.

The `^` specifier allows for "SemVer compatible" updates to the specified
version, but the exact rules around these are complex, and - it seems -
more restrictive than what we want (at least for most of our deps.

The new `>=` will allow for _any_ newer version of a dependency to be
used. We will need to be somewhat careful (and frequent) in checking that
a new version a dependency does not break the APIs we are using.

See https://python-poetry.org/docs/dependency-specification/ for more
details about the different available specifiers.

To verify that upgraded dependencies do not break FawltyDeps, we upgrade
_all_ our pinned versions to the latest available and modulo two small
nitpicks reported by the latest Mypy version, it seems we're all good to
go! :-)

Commits:
- Apply suggested `pyproject.toml` changes in issue #408
- Re-pin dependencies in lock file
- Appease Mypy after updating pinned version from 1.0.1 to 1.4.1

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,9 +35,13 @@ fawltydeps = "fawltydeps.main:main"
 # These are the main dependencies for fawltydeps at runtime.
 # Do not add anything here that is only needed by CI/tests/linters/developers
 python = ">=3.7.2"
-importlib_metadata = ">=6.6.0"
+importlib_metadata = [
+    # v6.8.0 drops support for Python v3.7:
+    {version = ">=6.6.0", python = ">=3.8"},
+    {version = ">=6.6.0,<6.8.0", python = "<3.8"},
+]
 isort = [
-    # isort 5.12.0 drops support for Python v3.7:
+    # v5.12.0 drops support for Python v3.7:
     {version = "^5.10", python = ">=3.8"},
     {version = ">=5.10,<5.12.0", python = "<3.8"},
 ]
@@ -45,7 +49,11 @@ pip-requirements-parser = ">=32.0.1"
 pydantic = ">=1.10.4,<3.0.0"
 tomli = {version = "^2.0.1", python = "<3.11"}
 typing-extensions = {version = ">=4.4.0", python = "<3.8"}
-setuptools = ">=68.0.0"
+setuptools = [
+    # v68.1.0 drops support for Python v3.7:
+    {version = ">=68.0.0", python = ">=3.8"},
+    {version = ">=68.0.0,<68.1.0", python = "<3.8"},
+]
 
 [tool.poetry.group.nox]
 optional = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,18 +34,18 @@ fawltydeps = "fawltydeps.main:main"
 [tool.poetry.dependencies]
 # These are the main dependencies for fawltydeps at runtime.
 # Do not add anything here that is only needed by CI/tests/linters/developers
-python = "^3.7.2"
-importlib_metadata = "=6.6.0"
+python = ">=3.7.2"
+importlib_metadata = ">=6.6.0"
 isort = [
     # isort 5.12.0 drops support for Python v3.7:
     {version = "^5.10", python = ">=3.8"},
     {version = ">=5.10,<5.12.0", python = "<3.8"},
 ]
-pip-requirements-parser = "^32.0.1"
+pip-requirements-parser = ">=32.0.1"
 pydantic = ">=1.10.4,<3.0.0"
 tomli = {version = "^2.0.1", python = "<3.11"}
-typing-extensions = {version = "^4.4.0", python = "<3.8"}
-setuptools = "^68.0.0"
+typing-extensions = {version = ">=4.4.0", python = "<3.8"}
+setuptools = ">=68.0.0"
 
 [tool.poetry.group.nox]
 optional = true
--- a/fawltydeps/utils.py
+++ b/fawltydeps/utils.py
@@ -42,7 +42,7 @@ def hide_dataclass_fields(instance: object, *field_names: str) -> None:
         raise TypeError(f"{instance!r} is not a dataclass instance")
     remaining_fields = {
         name: value
-        for name, value in instance.__dataclass_fields__.items()  # type: ignore
+        for name, value in instance.__dataclass_fields__.items()
         if name not in field_names
     }
     object.__setattr__(instance, "__dataclass_fields__", remaining_fields)
--- a/poetry.lock
+++ b/poetry.lock
@@ -47,13 +47,13 @@ wrapt = {version = ">=1.11,<2", markers = "python_version < \"3.11\""}
 
 [[package]]
 name = "astroid"
-version = "3.0.0"
+version = "3.0.2"
 description = "An abstract syntax tree for Python with inference support."
 optional = false
 python-versions = ">=3.8.0"
 files = [
-    {file = "astroid-3.0.0-py3-none-any.whl", hash = "sha256:f2510e7fdcd6cfda4ec50014726d4857abf79acfc010084ce8c26091913f1b25"},
-    {file = "astroid-3.0.0.tar.gz", hash = "sha256:1defdbca052635dd29657ea674edfc45e4b5be9cd53630c5b084fcfed94344a8"},
+    {file = "astroid-3.0.2-py3-none-any.whl", hash = "sha256:d6e62862355f60e716164082d6b4b041d38e2a8cf1c7cd953ded5108bac8ff5c"},
+    {file = "astroid-3.0.2.tar.gz", hash = "sha256:4a61cf0a59097c7bb52689b0fd63717cd2a8a14dc9f1eee97b82d814881c8c91"},
 ]
 
 [package.dependencies]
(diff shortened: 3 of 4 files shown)


0x26res:
> Re-pin dependencies in lock file

Did you run `poetry update`? I would have thought this would update setuptools and importlib_metadata in the lock file, but (as far as I can tell) they don't look updated.

jherland:
> > Re-pin dependencies in lock file
> 
> Did you run `poetry update`? I would have thought this would update setuptools and importlib_metadata in the lock file, but (as far as I can tell) they don't look updated.

I ran `poetry lock` (which AFAICS also updates, unless `--no-update` is passed). setuptools does not seem to have been updated, I agree, but importlib_metadata was upgraded from 6.6.0 to 6.7.0.

When I run `poetry update` in my local tree now, it makes no changes to the `poetry.lock` file at all.

I don't know yet why this does not update to the latest versions available on PyPI...

0x26res:
>  setuptools does not seem to have been updated

It could be because another dependency, most probably in the extra deps (lint, format, dev, nox, test) is pinned to an older version of setuptools. I had a look (with pipdeptree), and couldn't find the offender.

In a way it doesn't matter, when resolving dependencies, pip will use the information in the main dependencies section of the pyproject, it won't look at the loc file. The only reason we'd want setuptools to be updated the lock file is if we want to run the test against a newer version of setup tools.

dorranh:
> I ran `poetry lock` (which AFAICS also updates, unless `--no-update` is passed). setuptools does not seem to have been updated, I agree, but importlib_metadata was upgraded from 6.6.0 to 6.7.0.
> 
> When I run `poetry update` in my local tree now, it makes no changes to the `poetry.lock` file at all.
> 
> I don't know yet why this does not update to the latest versions available on PyPI...

@jherland might this be due to how the version bound for `types-setuptools` is currently defined? 
https://github.com/tweag/FawltyDeps/blob/a071f87f08b00d691b7e5c8fc33e3a52b419252e/pyproject.toml#L109

jherland:
@dorranh: changing our `types-setuptools` dependency from `^65.6.0.2` to `>=65.6.0.2` happily upgrades types-setuptools to `69.0.0.0`, but does not affect setuptools at all.

I think the reason setuptools is stuck at `68.0.0` is actually quite a bit simpler: [setuptools v68.1.0 removed support for Python v3.7](https://github.com/pypa/setuptools/blob/main/NEWS.rst#v6810), which means that for as long as FawltyDeps keeps supporting v3.7 Poetry will keep setuptools at <v68.1.0 (aka. ==v68.0.0 as of now).

This fact was not obvious from our `poetry.lock` as it only contains information about the versions of dependencies that Poetry has found to be compatible, and says nothing about why a version of a dependency was found incompatible.

---

FWIW, I think this might point to a bug in Poetry... If I change our `pyproject.toml` like this:

```diff
-setuptools = ">=68.0.0"
+setuptools = [
+    # setuptools 68.1.0 drops support for Python v3.7:
+    {version = ">=68.0.0", python = ">=3.8"},
+    {version = ">=68.0.0,<68.1.0", python = "<3.8"},
+]
 ```

then I shouldn't really have changed anything in Poetry's dependency calculation; in both cases:
- on Python 3.7 setuptools is restricted to `>=68.0.0` (from our dependency specification), and `<68.1.0` due to setuptool's own python requirement.
- on Python >=3.8 setuptools is only restricted to `>=68.0.0` (from our dependency specification).

However, with the above diff applied I _do_ get newer setuptools installed in all venvs that have Python >=3.8.

```
$ for venv in .nox/tests-3-*; do echo "--- $venv:"; "$venv/bin/pip" list | grep setuptools; done
--- .nox/tests-3-10:
setuptools              69.0.3
--- .nox/tests-3-11:
setuptools              69.0.3
--- .nox/tests-3-12:
setuptools              69.0.3
--- .nox/tests-3-7:
setuptools              68.0.0
--- .nox/tests-3-8:
setuptools              69.0.3
--- .nox/tests-3-9:
setuptools              69.0.3
```

FTR, the same argument applied to importlib_metadata (which dropped support for Python 3.7 in v6.8.0). Applying the equivalent "over-specification" convinces Poetry to install the latest version on Python >=3.8:

```
$ for venv in .nox/tests-3-*; do echo "--- $venv:"; "$venv/bin/pip" list | grep importlib-metadata; done
--- .nox/tests-3-10:
importlib-metadata      7.0.1
--- .nox/tests-3-11:
importlib-metadata      7.0.1
--- .nox/tests-3-12:
importlib-metadata      7.0.1
--- .nox/tests-3-7:
importlib-metadata      6.7.0
--- .nox/tests-3-8:
importlib-metadata      7.0.1
--- .nox/tests-3-9:
importlib-metadata      7.0.1
```

So it seems to me that Poetry mistakenly applies its Python 3.7 calculations to all other Python versions unless we jump through hoops to "over-specify" our dependencies. I'm not sure though, am I missing something here?

**Comment by jherland** [comment]
I've created a minimal demo at https://github.com/jherland/poetry_overspecify_deps_demo. AFAICS this exact issue does not exist on the [Poetry issue tracker](https://github.com/python-poetry/poetry/issues) yet?

**Comment by jherland** [comment]
I opened a discussion at https://github.com/orgs/python-poetry/discussions/8857 for now. Let's see if it is promoted into a bug. In the meantime, I guess maybe the best approach for this PR is to "over-specify" our dependencies, to improve test coverage across appplicable dependency versions?

**Comment by jherland** [comment]
> I opened a discussion at https://github.com/orgs/python-poetry/discussions/8857 for now.

Discussion concluded: This is a known shortcoming, unlikely to change in the near future. I guess we need to keep over-specifying our (main) dependencies for now.

**Comment by 0x26res** [comment]
I apologise for sending you in this rabbit hole. I guess it's not a huge deal to over specify the deps for now, given you don't have many of them.

File: pyproject.toml [hunk 1]
```diff
@@ -34,18 +34,26 @@ fawltydeps = "fawltydeps.main:main"
 [tool.poetry.dependencies]
 # These are the main dependencies for fawltydeps at runtime.
 # Do not add anything here that is only needed by CI/tests/linters/developers
-python = "^3.7.2"
-importlib_metadata = "=6.6.0"
+python = ">=3.7.2"
```
mknorps, line 37:
```suggestion
python = "^3.7.2"
```

pyproject.toml
```diff
@@ -34,18 +34,26 @@ fawltydeps = "fawltydeps.main:main"
...
+importlib_metadata = [
+    # v6.8.0 drops support for Python v3.7:
+    {version = ">=6.6.0", python = ">=3.8"},
+    {version = ">=6.6.0,<6.8.0", python = "<3.8"},
```
mknorps, line 41: It would be safer to take a caret requirement here. Major version change may affect our usage.
```suggestion
    {version = "^6.6.0", python = ">=3.8"},
    {version = ">=6.6.0,<6.8.0", python = "<3.8"},
```

pyproject.toml
```diff
@@ -34,18 +34,26 @@ fawltydeps = "fawltydeps.main:main"
...
+]
 isort = [
-    # isort 5.12.0 drops support for Python v3.7:
+    # v5.12.0 drops support for Python v3.7:
     {version = "^5.10", python = ">=3.8"},
     {version = ">=5.10,<5.12.0", python = "<3.8"},
 ]
-pip-requirements-parser = "^32.0.1"
+pip-requirements-parser = ">=32.0.1"
```
mknorps, line 48: Comment: It looks like the authors of this library do not follow semantic versioning pattern `MAJOR.MINOR.PATCH` unless they started with `MAJOR=30`, see https://pypi.org/project/pip-requirements-parser/#history.

mknorps (changes requested):
Thank you Johan for that fix! This is not an obvious thing that Poetry does, and it is good we can at leas work our way around that complication.

My comments are in the spirit of "specify narrow if you can". You did two changes in some places - split requirement version dependening on the Python version and changed caret requirement to ">=". I think the latter change is not necessary. Following poetry documentation on [version specification](https://python-poetry.org/docs/dependency-specification/):
> For instance, if we previously ran poetry add `requests@^2.13.0` and wanted to update the library and ran `poetry update requests`, poetry would update us to version `2.14.0` if it was available, but would not update us to `3.0.0`. If instead we had specified the version string as `^0.1.13`, poetry would update to `0.1.14` but not `0.2.0`. `0.0.x` is not considered compatible with any other version.

Looking also at the examples they gave, caret requirement is allowing change in the second meaningful digit in the version number.

mknorps (approved):
After discussion with @jherland I agree that being more restrictive about the major version may not help us ensure that FawltyDeps will not fail. 

We better create tests that check for the drift of requirements version bounds and a check if FawltyDEps is working on the updated versions. Described in https://github.com/tweag/FawltyDeps/issues/411.

pyproject.toml
jherland: Yeah, and this is a bigger problem (IMHO) with Poetry's affinity for using `^` by default: It assumes that packages are (a) using SemVer, and (b) using it correctly. My impression so far is that a lot of packages fail either (a) or (b), and even then, what constitutes a compatibility-breaking change will differ dependending on _how_ you are using the package, and a breaking change might not even be discovered until it's too late (i.e. after release).

As we discussed previously today, I think we're better off being more liberal/flexible in our dependency version specifications, but rather have automatic tests that will quickly flag problems when a new version of a dependency is released.

PR #414, opened 2024-01-12
Use __future__.annotations instead of stringified type annotations

jherland:
`from __future__ import annotations` became available in Python 3.7, so there is no reason for us not to use it instead of the older "stringified" type annotations (for types that are not yet declared at the time the parser encounters the annotation).

### DIFF
--- a/fawltydeps/main.py
+++ b/fawltydeps/main.py
@@ -10,6 +10,8 @@ minimal computation involved in setting the install_requires and extras_require
 arguments.
 """
 
+from __future__ import annotations
+
 import json
 import logging
 import sys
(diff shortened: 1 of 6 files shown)


mknorps (approved):
Nice!

For the consistency sake, there is one more place we could use annotations. In `dir_traversal.py`, https://github.com/tweag/FawltyDeps/blob/1cc9202f3cc9decf277735f9e43f70497743487f/fawltydeps/dir_traversal.py#L42

Otherwise ready to :rocket:

jherland:
> For the consistency sake, there is one more place we could use annotations. In `dir_traversal.py`,
> https://github.com/tweag/FawltyDeps/blob/1cc9202f3cc9decf277735f9e43f70497743487f/fawltydeps/dir_traversal.py#L42

Nice catch, fixed in [51da946](https://github.com/tweag/FawltyDeps/pull/414/commits/51da946f7dd26b02c4560eb0f047ce462d4045fe).

PR #415, opened 2024-01-18
Fixes for various `--exclude` usability issues

jherland:
Thanks to Maria for actually _testing_ the new `--exclude` option from a user's POV, and notifying me of various griveous issues found. Here we try to tackle the ones that were found and raise in the later stages of #388:

- Running `fawltydeps` with `--exclude` and passing relative paths instead of absolute (our test suite relies too heavily on absolute paths)
- Passing "anchored" exclude patterns to `--exclude` was initially unsupported. Fix that.
- Give the user a warning when the `--exclude` option directly overlaps with a given path (in which case the given path will always win).
- Fix issues where the precedence between `--exclude` patterns and given paths were not followed in practice.

### DIFF
--- a/fawltydeps/traverse_project.py
+++ b/fawltydeps/traverse_project.py
@@ -66,6 +66,25 @@ def find_sources(  # pylint: disable=too-many-branches,too-many-statements
     for pattern in settings.exclude:
         traversal.exclude(pattern)
 
+    requested_paths = {
+        path
+        for path in settings.code | settings.deps | settings.pyenvs
+        if isinstance(path, Path)
+    }
+    defaults = Settings.config(config_file=None)()
+    default_paths = defaults.code | defaults.code | defaults.pyenvs
+    if settings.exclude != defaults.exclude:  # non-default exclude
+        for path in requested_paths:
+            if path in default_paths:
+                continue  # skip checking for conflicts against default paths
+            if traversal.is_excluded(path, is_dir=path.is_dir()):
+                # This path will actually _not_ be excluded in practice
+                # (for files we will handle them directly below, for dirs, they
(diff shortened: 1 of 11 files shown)


tests/test_real_projects.py
```diff
@@ -53,9 +52,9 @@ def verify_requirements(venv_path: Path, requirements: List[str]) -> None:
 def run_fawltydeps_json(
     *args: str, venv_dir: Optional[Path], cwd: Optional[Path] = None
 ) -> JsonData:
-    argv = [sys.executable, "-m", "fawltydeps", f"--config-file={os.devnull}", "--json"]
+    argv = [sys.executable, "-I", "-m", "fawltydeps", "--json"]
```
mknorps, line 55: TIL!

fawltydeps/traverse_project.py
```diff
@@ -61,10 +62,36 @@ def find_sources(  # pylint: disable=too-many-branches,too-many-statements
...
     for pattern in settings.exclude:
-        traversal.exclude(pattern)
+        try:
+            traversal.exclude(pattern)
+        except ExcludeRuleError:  # Anchored pattern needs a base_dir
+            for path in requested_paths:
+                if path.is_dir():
+                    traversal.exclude(pattern, base_dir=path)
+
+    defaults = Settings.config(config_file=None)()
```
mknorps, line 82: Very clever :heart:

mknorps (approved):
This is great! Thank you @jherland !

The changes you did to fix listed issues are very clever :) I like how you expanded the test suite to account for both - relative and absolute paths and how you added the overlaping check. I also tested the changes manually and all works as expected.
This is :rocket:

PR #420, opened 2024-01-29
Allow reading exclude patterns from files (`--exclude-from`)

jherland:
Commits:
- `dir_traversal`: Add `DirectoryTraversal.exclude_from()`
- `Settings`: Add `.exclude_from`, paths containing exclude patterns
- `test_sample_projects`: Run experiments from within the project dir
- `test_sample_projects`: Add `exclude_from` tests
- `dir_traversal`: Prioritize `.exclude` rules over `.exclude_from` rules
- Add `--exclude-from` option to CLI


fawltydeps/traverse_project.py
```diff
@@ -93,6 +94,12 @@ def find_sources(  # pylint: disable=too-many-branches,too-many-statements
                 # of this overlap, so log a warning:
                 logger.warning(f"{path} is both requested and excluded. Will include.")
 
+    for file_with_exclude_patterns in settings.exclude_from:
+        if file_with_exclude_patterns.is_file():
+            traversal.exclude_from(file_with_exclude_patterns)
+        elif file_with_exclude_patterns not in DEFAULT_EXCLUDE_FROM:
+            logger.warning(f"Cannot find {file_with_exclude_patterns}, skipping")
+
```
mknorps, line 102: How will the error handling work if a file contain incorrect pattern?
Like the situation from the loop above (for `exclude` pattern), where some path is in both requested and excluded?

I wonder if we can use the code already availably for handling the ignore patterns to handle contenmts of the files containing those patterns.

mknorps (commented):
I gave it a first loo, and it looks good. I will have to test it iin the CLI stil, and I have one question :)

fawltydeps/traverse_project.py
jherland: Good question. I left this situation unhandled for now, that is: A conflict between an exclude pattern in a file and a given (code/deps/pyenv) path is currently handled _silently_ in favor of the given path (no warning is given).

My reason for this (which I failed to explain in the code) is that I consider there to be a greater level of _user intention_ behind an exclude pattern that is passed directly to FawltyDeps compared to an exclude pattern that comes indirectly via a file (especially a file like `.gitignore` which is also used for other purposes).

In the first case: when the user says `--exclude foo` (or similar via `[tool.fawltydeps]`), and this conflicts with a given `foo` code/deps/pyenv path, then they have (directly) told FD to "look at `foo`" _and_ "please ignore`foo`" in the same invocation. This is nonsense, and we have to assume that there is some confusion somewhere. Thus the warning makes sense here, IMHO.

In the second case, the user is pointing at a file of exclude patterns, and my gut feeling is that the extra level of indirection somewhat "weakens" the user's intention behind those patterns. In other words, when the user's `.gitignore` contains `foo`, but the user has explicitly told us to look at `foo`, I feel that we can reasonably believe that the user actually wants us to look at `foo`, and throwing a warning here could be considered unnecessary noise.

A real-life example exists within FD itself: When we run `nox`, it maintains a collection of virtualenvs under `.nox/`. Obviously we don't want to check these into git, so we have this line in our `.gitignore`:
```
.nox/
```
However, the virtualenvs under `.nox/` are good candidates for the `--pyenv` option, so I could easily see myself running `fawltydeps --pyenv .nox/ --exclude-from .gitignore` in our project. In this case, I would personally feel it a bit noisy for FD to warn me of the `.nox` conflict.

Still, my argument above is based mostly on gut feeling and what I _assume_ the user means in various scenarios. Please feel free to disagree, or consider if there are cases I have overlooked. It would not take much to convince me to throw warnings for all such conflicts, whether they were caused by`--exclude` or via `--exclude-from`. 😅

mknorps: I think I get it now. We are coming from two different understanding what a file provided in `--exclude-from` means.

One approach, that as I feel you took, is to use files that do already something else, like `.gitignore` file and say: "I do not want FD to look at files matching this patterns. They will not be documented in GIT so they do not add to the reproducibility anyway".
In that approach it is reasonable to prioritize _user intent_ that is expressed implicitly in the command line, like 
`fawltydeps --code bar --exclude-from foo `, where `foo` file contains line:
```
bar
```

Another approach, which was more intuitive to me is that a user creates a `.fawltydepsignore` file themselves and intentionally:
```
bar/*
tests
```

In this file, there is a `tests` directory, which the user does not want to check along with other files that should not be provided to FawltyDeps. And now, a user writes:
```
fawltydeps --code tests --ignore-from `.fawltydepsignore`
```
If the `.fawltydepsignore` file is big enough, it may not be obvious that there is a conflict in the user's intentions.
This usage may happen in case where the `exclude` section in `[tools.fawltydeps]` is a bit more complex: you may want to move it to a separate file, just like with the user mapping file. Mapping files and ignore-from files are a bit different in nature, the latter would be moving the configuration of executing FD outside of `pyproject.toml`, which I do not think is a good practice. I can imagine such case anyway.

As a consequence, I would designed the code a bit different.
This, I treat as a basic exclude patter collection code:
```python
    traversal: DirectoryTraversal[AttachedData] = DirectoryTraversal()
    for pattern in settings.exclude:
        try:
            traversal.exclude(pattern)
        except ExcludeRuleError:  # Anchored pattern needs a base_dir
            for path in requested_paths:
                if path.is_dir():
                    traversal.exclude(pattern, base_dir=path)
```
But in case of file as an input we should change ` for pattern in settings.exclude` to ` for pattern in get_patterns_from_file(input_file)`.

jherland:
@mknorps: Thanks to your comments, and having had some time away from this PR, I've realized that this PR is trying to do too much in one go, specifically:

- I agree with you that exclude patterns read from files given by the user should be treated the same as exclude patterns given directly on the command line (see 3e94dfe for beginning to fix this).
- After looking more closely at [ruff's `respect-gitignore` setting](https://docs.astral.sh/ruff/settings/#respect-gitignore), I found two things:
  - (a) The way I handled gitignore files was incomplete: I did not consider global gitignore files (e.g. `~/.config/git/ignore`), nor did I auto-detect/parse `.gitignore` and `.ignore` files found further inside the project. Really, this cannot be solved with a static `.exclude_from` set of paths, instead I think we need to detect/parse these files _while_ we are traversing the project structure.
  - (b) The `respect-gitignore` boolean option is a much more user-friendly way to expose this functionality to the user (rather than manipulating sets of paths). Also, using a simple boolean option frees us up to write a _proper_ implementation for the problems described in (a), rather than somehow trying to combine this into `.exclude_from`.
- Therefore: `--exclude-from` should default to the empty set. This ends up _simplifying_ this PR, but adding/implementing the `respect-gitignore` boolean option will have to be done in a later PR. 🤷🏻‍♂️

fawltydeps/traverse_project.py
jherland: See the comment I just posted below, for how I want to simplify this PR. I think I agree with you on ~all the details.

mknorps: Thank you Johan!

mknorps (approved):
Checked and tested :rocket: 
Thank you @jherland for the contribution :heart_decoration:

Issue #176, opened 2023-02-22
Bug: Dependency to Import mapping on the `qiskit` package does not include the `qiskit` import name


Nour-Mws:
FawltyDeps v0.3.1

When running FawltyDeps (in the customary FawltyDeps env) on the Algorithms project, the lists of unused deps and undeclared imports do not contain the `qiskit` package as it is both used and declared:
Excerpt:
```json
{
          "name": "qiskit",
          "source": {
            "path": "quantum/q_full_adder.py",
            "lineno": 13
          }
}
```

```json
{
      "name": "qiskit",
      "source": {
        "path": "requirements.txt"
      },
      "import_names": [
        "qiskit"
      ],
      "mapping": "IDENTITY"
    },
```
As FawltyDeps defaults to identity mapping in this instance, the import and dependency name get matched.

When I tried installing all declared dependencies of Algorithms (+FawltyDeps) in a new environment, I got `qiskit` in the list of undeclared and unused:
```
These imports appear to be undeclared dependencies:
- 'django'
- 'mpmath'
- 'pytest'
- 'qiskit'
- 'scipy'
- 'seaborn'
These dependencies appear to be unused (i.e. not imported):
- 'keras'
- 'projectq'
- 'qiskit'
- 'texttable'
- 'yulewalker'
```

Looking at the json output, I got:
```json
   {
      "name": "qiskit",
      "references": [
        {
          "name": "qiskit",
          "source": {
            "path": "requirements.txt"
          },
          "import_names": [
            "test",
            "setup"
          ],
          "mapping": "DEPENDENCY_TO_IMPORT"
        }
      ]
    },
```

No `qiskit` to be seen in the `import_names`!

When I tried to import `qiskit` in Python in that venv, however, it worked without problem!

I thought for a second that `qiskit` might be exposed by other packages in the environment, but in that case it would have been matched and it wouldn't have appeared in the undeclared imports.

I tried importing `qiskit` in the FawltyDeps-only environment and Python couldn't find it (as expected).

I haven't dug deeper than that. My next step would be to see how Python's import finder managed to find `qiskit` in the 2nd environment in the first place.

PR #421, opened 2024-02-28
real_projects/python-algorithms: Pin qiskit at <1.0

jherland:
qiskit v1.0 starts to declare the `qiskit` import name (see issue #176 for
details about the weirdness that precedes v1.0). This changes the
behavior of our `real_projects` test of the Algortihms project.

qiskit has also dropped support for Python 3.7 (since v0.44), hence it
is easier for us to keep our tests (which need to work with Python 3.7)
running on the older version, for now.


mknorps (approved):
👍

PR #424, opened 2024-04-16
Add the ruff linter

jherland:
A minimal introduction of `ruff`. The first in a series of 3 PRs.

Commits:
- Mypy: Disallow blanket `type: ignore`
- `pyproject.toml`: Add `ruff` as a lint dependency
- Add minimal `ruff` configuration
- `noxfile.py`: Add `ruff check .` to `lint` target


mknorps (approved):
Thank you for introducing ruff to FawltyDeps. The developer's experience will improve a lot (and CI will be happier as well ;) )

I have one question about the ruff settings, namely:
```
target-version = "py37"
```
It is not ideal that we have to define lower boundary version in two places (second being):
```
[tool.poetry.dependencies]
...
python = ">=3.7.2"
```
I looked into [target-version ](https://docs.astral.sh/ruff/settings/#target-version) option documentation, and I see that it would have been read by default, but from:
```
[project]
requires-python = ">=3.8"
```
Is this a poetry-focused usage of pyproject.toml which disallows that?

Apart from this question, I have no other comments. 
All is clear, works well, and is documented :rocket:

jherland:
> Is this a poetry-focused usage of pyproject.toml which disallows that?

Exactly. Since we're using Poetry, we're currently defining our project metadata in the Poetry-specific `[tool.poetry]` section in our `pyproject.toml`.

> ```
> [project]
> requires-python = ">=3.8"
> ```

This is from the [PEP621](https://peps.python.org/pep-0621/) standard for specifying project metadata, and this standard seems to have more momentum than Poetry's own format, so I'd like to migrate to this at some point. Either we see if Poetry can support the more standard format, or we look around for alternatives to Poetry. But that is a separate issue/effort than this PR. 😄

jherland:
> Either we see if Poetry can support the more standard format, or we look around for alternatives to Poetry. But that is a separate issue/effort than this PR.

Created https://github.com/tweag/FawltyDeps/issues/433 to keep discuss our future with/without Poetry

PR #427, opened 2024-04-17
Use `ruff format` instead of `black`

jherland:
Ruff comes with a built-in formatter (`ruff format`) which serves as a drop-in replacement for `black`. The difference in the formatting are minuscule, and can be seen in the second commit in this PR.

Also remove `isort` as a formatting dependency, since `ruff check` already performs the same checks.

Commits:
- Use ruff's default `line-length`
- Formatting changes introduced by `ruff format .`
- `noxfile.py`: Run `ruff format` instead of `black`
- Remove `black` as a dependency
- Remove `isort` as a formatting dependency
- Update docs to mention `ruff` instead of `black` + `isort`


mknorps (approved):
Tested and approved :rocket: 

image
:arrow_up: always a happy sight :)

PR #428, opened 2024-04-23
Top PyPi packages experiment and execution speed improvement

mknorps:
This PR:
- adds top PyPI packages experiment, 
- reduces constant declarations 
- improves execution speed by taking only the last commit on the main branch from the PyPi repository


PyPI_analysis/experiments/top_1400_pypi_packages/experiment.py
```diff
@@ -0,0 +1,101 @@
...
+        upload_to_aws(save_location, "pypi_analysis/data/" + repo_name)
+
+    try:
+        repository.analysis()
+    except:
+        print(f"Could not analyse repository {repo_name} due to error:\n {e}")
+    else:
+
+        if os.path.exists(os.path.join(results_dir_name, repo_name + ".json")):
+            print("Writing results of FalwtyDeps analysis to disk.")
```
zz1874, line 61: Typo :)
```suggestion
            print("Writing results of FawltyDeps analysis to disk.")
```

PyPI_analysis/experiments/top_1400_pypi_packages/experiment.py
```diff
@@ -0,0 +1,101 @@
...
+            if os.path.exists(save_location):
+                shutil.rmtree(save_location)
+            return
+
+    if save_on_aws:
+        upload_to_aws(save_location, "pypi_analysis/data/" + repo_name)
+
+    try:
+        repository.analysis()
+    except:
```
zz1874, line 56:
```suggestion
    except Exception as e:
```

PyPI_analysis/experiments/top_1400_pypi_packages/experiment.py
```diff
@@ -0,0 +1,101 @@
...
+            #         os.path.join(results_dir_name, repo_name + ".json"),
+            #         bucket,
+            #         os.path.join(
+            #             "pypi_analysis", results_dir_name, repo_name + ".json"
+            #         ),
+            #     )
+            # except ClientError as e:
+            #     print(f"Client error {e} occurs.")
+        else:
+            print(f"Cannot find analysis results of {repo_name}!")
```
zz1874, line 74: As far as I understand, this `else` clause will only be executed if there are no errors in `repository.analysis()` and no `.json` file found? But if there are no errors in `repository.analysis()`, there should be `.json` file found, right?

zz1874 (changes requested):
Great work!
I'm really happy that we can get rid of the heavy JSON results in this repo :+1: . 
Ideally, I think we can combine `biomedical_projects_experiment/experiment.py` and `top_1400_pypi_packages/experiment.py` into one single `experiment.py` in the `experiments/` folder, since most of the lines of them are the same. We can do it in a separate PR.

PyPI_analysis/experiments/top_1400_pypi_packages/experiment.py
mknorps: Yes, but only `json.JSONDecodeError` is being caught in `download_and_analyze.py`.
That is the case for  `s3transfer` library for example:
```
Command failed with return code 1
Cannot find analysis results of s3transfer!
```

PR #432, opened 2024-05-07
Use uv instead of pip to manage virtualenvs

jherland:
`uv` is much faster than `pip` for installing package (it is also faster then `python -m venv` for creating virtualenvs). This PR switches from `venv`/`pip` to `uv` in three separate parts of our project:

1. Instruct Nox to use `uv` instead of `pip` to manage the virtualenv associated with each Nox session (first commit).
2. Use `uv` instead of `venv`/`pip` (when available) to manage virtualenvs for our `sample_project` and `real_projects` tests (next three commits).
3. Use `uv` instead of `venv`/`pip` (when available) to create and populate the temporary virtualenv managed by the `--install-deps` option (next two commits).

The first two points above only concern our tests and has no effect on the FawltyDeps program itself. The last one changes the behavior of the `--install-deps` option to use `uv` when available (otherwise fall back to `venv`/`pip`).

The last commit in this PR adds a small convenience for our users: If they want to ensure that `uv` is available for FawltyDeps to use, they can now install `fawltydeps[uv]`. This `uv` extra will bring in `uv` as a dependency in much the same way that we currently depend on `nox[uv]` in our own developement environment.

Commits:
- Use `uv` instead of `pip` to manage nox virtualenvs
- `tests/project_helpers`: Use `uv` (if available) to prepare virtualenvs
- `.github/workflows/tests.yaml`: Preserve `uv` cache across test runs
- Sidestep issues installing `projectq` with 'uv pip install'
- Prepare to decouple `--install-deps` from `pip`
- `TemporaryAutoInstallResolver`: Use `uv` if available
- `pyproject.toml`: Add `uv` as an "extra" dependency for FawltyDeps

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -31,6 +31,7 @@ jobs:
           path: |
             ~/.cache/pypoetry/virtualenvs
             ~/.cache/pytest
+            ~/.cache/uv
           key: ${{ runner.os }}-poetry-tests-${{ hashFiles('poetry.lock') }}
       - name: Install project
         run: poetry install --no-interaction --sync --with=nox,test
--- a/README.md
+++ b/README.md
@@ -232,14 +232,13 @@ fallback strategy.
 
 #### Mapping by temporarily installing packages
 
-Your local Python environements might not always have all your project's
+Your local Python environments might not always have all your project's
 dependencies installed. Assuming that you don’t want to go through the
 bother of installing packages manually, and you also don't want to rely on
 the inaccurate identity mapping as your fallback strategy, you can use the
-`--install-deps` option. This will `pip install`
-missing dependencies (from [PyPI](https://pypi.org/), by default) into a
-_temporary virtualenv_, and allow FawltyDeps to use this to come up with the
-correct mapping.
+`--install-deps` option. This will automatically install missing dependencies
+(from [PyPI](https://pypi.org/), by default) into a _temporary virtualenv_,
+and allow FawltyDeps to use this to come up with the correct mapping.
 
 Since this is a potentially expensive strategy (e.g. downloading packages from
 PyPI), we have chosen to hide it behind the `--install-deps` command-line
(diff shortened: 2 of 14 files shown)


mknorps:
Thank you @jherland 🫶 

I started checking the PR by looking at all usages of "pip". There are some places where we can consider updating the text or a name:
1. https://github.com/tweag/FawltyDeps/blob/57ebc4c4ebee7ffb811cc478b2ec6cb743514049/README.md#L442:
    "This will use `pip install`" -> "This will use `uv pip install` and fall back to `pip install` if `uv` is not available"
2. https://github.com/tweag/FawltyDeps/blob/57ebc4c4ebee7ffb811cc478b2ec6cb743514049/docs/DesignDoc.md#L192
    Should we also include `uv `instructions? Or should we only apply it in the developer context?
3. https://github.com/tweag/FawltyDeps/blob/57ebc4c4ebee7ffb811cc478b2ec6cb743514049/tests/conftest.py#L33-L34
4. https://github.com/tweag/FawltyDeps/blob/57ebc4c4ebee7ffb811cc478b2ec6cb743514049/tests/project_helpers.py#L178-L179
5. https://github.com/tweag/FawltyDeps/blob/57ebc4c4ebee7ffb811cc478b2ec6cb743514049/tests/test_install_deps.py#L29-L30 does this still hold for `uv`?

mknorps (approved):
Thank you @jherland for handling that ❤️ 

I did a small test before and after installing UV on detect-waste project and the results are awesome!
```
(fawltydeps-py3.11) ➜  detect-waste git:(main) time fawltydeps --install-deps
...
fawltydeps --install-deps  23.11s user 6.87s system 88% cpu 34.040 total
```
vs 🎉 
```
(fawltydeps-py3.11) ➜  detect-waste git:(main) time fawltydeps --install-deps
...
fawltydeps --install-deps  5.21s user 4.54s system 43% cpu 22.191 total
```

The `--install-deps` option will be much more usable now, thank you!

I did not meticulously check all the changes you made, but given that you did not have to change any test of our rich tests suite, I am confident (99% 😄  ) it works without errors.

jherland:
> I started checking the PR by looking at all usages of "pip". There are some places where we can consider updating the text or a name:

Agreed. Look at f2d359c for my rewordings.

>> `monkeypatch.setenv("PIP_NO_INDEX", "True")`
> does this still hold for `uv`?

Actually it does no!t 🙀

Furthermore, although `uv` supports some `UV_*` environment variables that mirror the `PIP_*` equivalents, these settings/variables cannot yet be set via the environment. I found a workaround instead that relies on setting up a (temporary) config file for `uv`. Take a look at 2795898 for the details.

Issue #438, opened 2024-06-12
Test regressions with Pydantic 1.10.16


musicinmybrain:
**Describe the bug**
Two tests fail with the just-released Pydantic 1.10.16, the latest release in the 1.x series.

**To Reproduce**

```
$ gh repo clone tweak/FawltyDeps
$ cd FawltyDeps
$ python3.12 -m venv _e
$ . _e/bin/activate
(_e) $ poetry install --with dev
(_e) $ python -m pytest
```

(tests pass with “default” Pydantic 2.5.3)

```
(_e) $ pip install pydantic==1.10.14
(_e) $ python -m pytest
(_e) $ pip install pydantic==1.10.15
(_e) $ python -m pytest
```

(tests still pass with Pydantic 1.10.14 and 1.10.15)

```
(_e) $ pip install pydantic==1.10.16
(_e) $ python -m pytest
```

```
=========================================================================================== short test summary info ============================================================================================
FAILED tests/test_settings.py::test_settings[env_var_with_wrong_type__raises_SettingsError] - pydantic.env_settings.SettingsError: error parsing env var "fawltydeps_actions"
FAILED tests/test_settings.py::test_settings[env_var_with_invalid_value__raises_SettingsError] - pydantic.env_settings.SettingsError: error parsing env var "fawltydeps_ignore_unused"
=================================================================== 2 failed, 1237 passed, 32 skipped, 53 deselected, 86 warnings in 37.85s ====================================================================
```

(Full output to be pasted in a comment.)

**Expected behavior**
All tests pass – as they do for Pydantic 2.5.3 and 1.10.14.

**Environment**
 - OS name + version: Fedora Linux 40 (`x86_64`)
 - Version of the code: current `main`, fe7787eccac4c6d1a34d7d876c84e11b0d6ab269

**Additional context**

While I was at it, I tried the just-released Pydantic 2.7.4, and I found that all tests pass, so this is exclusively a problem with Pydantic 1.10.16.

It’s not immediately obvious to me what the root cause is.

musicinmybrain:
```
============================= test session starts ==============================
platform linux -- Python 3.12.3, pytest-7.4.4, pluggy-1.2.0
cachedir: /home/ben/.cache/pytest
rootdir: /home/ben/src/forks/FawltyDeps
configfile: pyproject.toml
plugins: hypothesis-6.79.4
collected 1324 items / 53 deselected / 1271 selected

tests/test_analysis.py .                                                 [  0%]
[...]

PR #439, opened 2024-06-13
Work around issue in Pydantic v1.10.16

jherland:
Minor workaround to fix issue #438.

Upstream Pydantic issue: https://github.com/pydantic/pydantic/issues/9656

Commits:
- `test_settings`: Workaround issue in Pydantic v1.10.16
- `noxfile.py`: Rerun `poetry export` if `reqs_from_poetry.txt` is missing

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -59,7 +59,11 @@ def install_groups(
     requirements_txt = Path(session.cache_dir, session.name, "reqs_from_poetry.txt")
     hashfile = requirements_txt.with_suffix(".hash")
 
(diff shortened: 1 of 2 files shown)


jherland:
There is an ongoing PR to fix the upstream issue: https://github.com/pydantic/pydantic/pull/9660

If that ends up being merged and released soon, we can probably drop this PR.

jherland:
Pydantic v1.10.17 does not trigger the same issue, and this workaround is considered unnecessary.

Issue #429, opened 2024-04-23
FawltyDeps fails on s3transfer package


mknorps:
**Describe the bug**
FalwtyDeps fails when run on s3transfer package.

The problem is with the way dependencies are declared in setup.py.
Minimal not-working example of setup.py:
```
from setuptools import setup

setup(
    extras_require={
        'crt1': 'botocore>=1.33.2,<2.0a.0', # FawltyDeps fails
        'crt2': 'botocore[crt]', # FawltyDeps fails
        'crt3': ['botocore'], # FawltyDeps works
    },
)
```

**To Reproduce**
Download s3transfer package and run:
```
fawltydeps
```
from arbitrary environment.

**Expected behavior**
Minimal expectation: FawltyDeps informs of the problem and does not crash.
Ideal outcome: FalwtyDeps handles `setup.py` cases like the above.

**Environment**
 - OS name + version: Ubuntu 22.04 LTS
 - Version of the code:  0.15.0

**Additional context**

```
(fawltydeps-py3.10) ➜  s3transfer git:(master) fawltydeps         
WARNING:fawltydeps.limited_eval:Unable to resolve Call(func=Attribute(value=Attribute(value=Name(id='os', ctx=Load()), attr='path', ctx=Load()), attr='dirname', ctx=Load()), args=[Name(id='__file__', ctx=Load())], keywords=[]) from os.path.dirname(__file__) @ setup.py:7
WARNING:fawltydeps.limited_eval:Failed to parse assignment of 'ROOT': Call(func=Attribute(value=Attribute(value=Name(id='os', ctx=Load()), attr='path', ctx=Load()), attr='dirname', ctx=Load()), args=[Name(id='__file__', ctx=Load())], keywords=[]) from os.path.dirname(__file__) @ setup.py:7
WARNING:fawltydeps.limited_eval:Unable to resolve Call(func=Attribute(value=Name(id='re', ctx=Load()), attr='compile', ctx=Load()), args=[Constant(value='__version__ = [\'"]([0-9.]+)[\'"]')], keywords=[]) from re.compile('__version__ = [\'"]([0-9.]+)[\'"]') @ setup.py:8
[...]

PR #440, opened 2024-06-17
parse_setup_py: Handle string dependency when we expect list of strings

jherland:
As described in issue #429, the s3transfer package comes with setup.py
where the `extras_require` dictionary does not map to a list of strings
(extra requirements), but instead map to a string directly.

When s3transfer is installed with this extra (e.g. `pip install
s3transfer[crt]`), the setuptools machinery handles this variation
gracefully, i.e. it interprets the string as if it was a single-element
list containing the same string.

An experiment reveals that the same interpretation occurs if a string is
passed directly in the `install_requires` argument.

This commit teaches FawltyDeps to do the same.

Also, if parse_one_req() fails to parse a requirement string, catch that
exception (likely an `InvalidRequirement` or some other subclass of
ValueError), and convert it into a DependencyParsingError, which is
handled more gracefully (i.e. print a warning message and continue).

### DIFF
--- a/fawltydeps/extract_declared_dependencies.py
+++ b/fawltydeps/extract_declared_dependencies.py
@@ -9,7 +9,7 @@ import tokenize
 from dataclasses import replace
 from pathlib import Path
 from tempfile import NamedTemporaryFile
-from typing import Callable, Iterable, Iterator, NamedTuple, Optional, Tuple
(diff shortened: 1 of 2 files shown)


mknorps (approved):
Thank you @jherland !

We are getting deeper into the rabbit hole, but at least we are closer to how things works under the Python hood.

I have no comments about the code, all nicely documented and improving the codebase :rocket:

Issue #443, opened 2024-07-16
Migrate away from pkg_resources


jherland:
We are getting this while importing `pkg_resources` from `setuptools`:

```
DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
```

Need to figure out a better provider for parsing requirements strings, e.g. `packaging`. See https://setuptools.pypa.io/en/latest/pkg_resources.html for more details.

setuptools also has a recent vulnerability (CVE-2024-6345) which probably does not affect us, but does remind us to make this change soon.

PR #445, opened 2024-07-23
Replace pkg_resources dependency with packaging.requirement

jherland:
Fixes issue #443.

Getting away from our use of `pkg_resources` turns out to be relatively easy:
The `packaging` package from the [Python Packaging Authority](https://github.com/pypa) is probably as close as we can get to an official and authoritative upstream for this kind of logic, and the `packaging.requirements.Requirement` class seems to encapsulate the equivalent logic the `pkg_resources.Requirement` class that we have used for this purpose until now.

At least, our test suite seems to pass with no further adjustments after this change. :-)

Note that this only concerns the parsing of _single_ requirement strings as they occur in `setup.py` and `pyproject.toml` files.
For parsing full `requirements.txt` files we still rely on [pip-requirements-parser](https://github.com/nexB/pip-requirements-parser/tree/main).
The reason for this is two-fold:
- `packaging.requirements` does not include functionality to parse complete files (and we don't want to have to deal with fiddly details like line continuations in `requirements.txt` files ourselves)
- `pip_requirements_parser` does not include a good API for parsing requirements strings that are _not_ part of a full `requirements.txt` file (requires writing a temp file as seen here: https://github.com/nexB/pip-requirements-parser/pull/19/files)

Commits:
- Add `packaging` as a core dependency
- Replace `pkg_resources.Requirement` with `packaging.requirements.Requirement`
- Remove `setuptools` (and `types-setuptools`) as a FawltyDeps dependency

### DIFF
--- a/fawltydeps/extract_declared_dependencies.py
+++ b/fawltydeps/extract_declared_dependencies.py
@@ -11,8 +11,8 @@ from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Callable, Iterable, Iterator, NamedTuple, Optional, Tuple, Union
 
+from packaging.requirements import Requirement
 from pip_requirements_parser import RequirementsFile  # type: ignore[import]
-from pkg_resources import Requirement
 
 from fawltydeps.limited_eval import CannotResolve, VariableTracker
 from fawltydeps.settings import ParserChoice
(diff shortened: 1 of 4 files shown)


obscurerichard (approved):
👍🏻 :shipit:

PR #456, opened 2024-09-11
Preliminary support for Pixi projects using pixi.toml

jherland:
Parsing `pixi.toml` is not so different from parsing Pixi fields inside `pyproject.toml`. This adds support for Pixi projects that use the `pixi.toml` configuration file.

NOTE: The same Conda caveat still applies: We do not currently differentiate between Conda dependencies and PyPI dependencies, meaning that we assume that a Conda dependency named FOO will map one-to-one to a Python package named FOO. This is certainly not true for Conda dependencies that are not Python packages, and it probably isn't even true for all Conda dependencies that do indeed include Python packages.

Commits:
- `PyEnvSource`: Slightly loosen restrictions on Python environments
- `extract_declared_dependencies.py`: Add support for parsing `pixi.toml`
- Add `pixi.toml` support to the rest of FawltyDeps
- `sample_projects`: Add Pixi project w/default settings
- `test_traverse_project`: Add tests using the Pixi sample projects
- `extract_declared_dependencies`: Handle exceptions due to invalid toml


mknorps (approved):
🚀 Thank you @jherland for this PR and thorough checks and tests on this new functionality.

PR #457, opened 2024-09-16
Initial support for environment.yml files 

jherland:
This adds support for Conda projects that use environment.yml for declaring dependencies.

NOTE: The same Conda caveat as for Pixi still applies: We do not currently differentiate between Conda dependencies and PyPI dependencies, meaning that we assume that a Conda dependency named FOO will map one-to-one to a Python package named FOO. This is certainly not true for Conda dependencies that are not Python packages, and it probably isn't even true for all Conda dependencies that do indeed include Python packages.

Commits:
- Restructure code for extracting declared dependencies
- Add PyYAML dependency
- Initial support for parsing `environment.yml` files
- Improve error handling when deps file is passed to mismatched parser
- Add `environment.yml` support to the rest of FawltyDeps
- `sample_projects`: Add example Conda project
- `test_traverse_project`: Add tests using the Conda sample project


fawltydeps/extract_deps.py
mknorps, line 1: Looks lighter that way!
What do you think about the following code organizing pattern:
```
└── extract_deps
    ├── __init__.py
    ├── pixi_parser.py
    ├── pyproject_parser.py
    ├── requirements_parser.py
    └── setup_parser.py
```
The content of the current `extract_deps.py` would be in `__init__.py`.
We would improve the logical structure of the package with purpose segregation.

jherland: Love this idea. And finally implemented in the latest rebase. 😄

mknorps (changes requested):
Thank you @jherland , this is a great improvement on both, the code readability side and on the features side.
environment.yml was a declaration file asked by the users to be supported :heart: 

:heavy_check_mark:  I checked the changes, looked if there are no places left with the old code structure references.

:stop_sign:  Then I ran the newest version on a randomly chosen [project](https://github.com/SciTools/cartopy) that uses [environment.yaml](https://github.com/SciTools/cartopy/blob/main/environment.yml) files to declare dependencies. Parsed names are trailed with `>` ehn there was some version specification of a package. To recreate, run in Cartopy project:
``` 
fawltydeps environment.yml --list-deps
```
The output I got is:
```
cython>
matplotlib-base>
numpy>
owslib>
packaging>
pillow>
pre-commit
pydata-sphinx-theme
pykdtree
pyproj>
pyshp>
pytest
pytest-mpl
pytest-xdist
ruff
scipy>
setuptools_scm
shapely>
sphinx
sphinx-gallery
```
Could you check if this is something we will not support for now or a parsing bug?

jherland:
> 🛑 Then I ran the newest version on a randomly chosen [project](https://github.com/SciTools/cartopy) that uses [environment.yaml](https://github.com/SciTools/cartopy/blob/main/environment.yml) files to declare dependencies. Parsed names are trailed with `>` when there was some version specification of a package.

Thanks for finding this!

I had overlooked the proper parsing of the package name from the version + build info in a Conda dependency specifier, and I obviously also had no test case to catch this.

Commit 4e75de3 should fix it: I added the `environment.yaml` file from cartopy as a new test case, and I fixed the parser with a relatively straightforward regex, stolen from Conda's own dependency parsing code.

With this added, I also needed to properly handle the case when a package name is missing or malformed. I added a couple of test cases for this, and made sure that we raise and handle the appropriate exceptions, so that we can print an error message about the faulty dependency and then keep on parsing the rest of the dependencies.

mknorps (approved):
Thank you @jherland.
Looks great now :rocket:

Issue #465, opened 2025-01-28
top-level venv doesn't work


vanschelven:
**Describe the bug**

When the virtualenv is the same as the project-dir, fawltydeps doesn't work

**To Reproduce**

When the venv is external, everything is fine:
```
klaas@pop-os:/tmp$ git clone git@github.com:simplejson/simplejson.git
Cloning into 'simplejson'...
[..]
klaas@pop-os:/tmp$ python -m venv externalvenv
klaas@pop-os:/tmp$ . externalvenv/bin/activate
(externalvenv) klaas@pop-os:/tmp$ pip install fawltydeps
Collecting fawltydeps
  Using cached fawltydeps-0.18.0-py3-none-any.whl (64 kB)
[..]
(externalvenv) klaas@pop-os:/tmp$ cd simplejson/
(externalvenv) klaas@pop-os:/tmp/simplejson$ fawltydeps 
[..]
WARNING:fawltydeps.limited_eval:Failed to parse assignment of 'cmdclass': Call(func=Name(id='dict', ctx=Load()), args=[], keywords=[keyword(arg='test', value=Name(id='TestCommand', ctx=Load()))]) from dict(test=TestCommand) @ setup.py:100

These imports appear to be undeclared dependencies:
- 'setuptools'

For a more verbose report re-run with the `--detailed` option.
```

Approach with "top level venv", however: 

```
klaas@pop-os:/tmp/simplejson$ python -m venv .
klaas@pop-os:/tmp/simplejson$ . bin/activate
(simplejson) klaas@pop-os:/tmp/simplejson$ pip install fawltydeps
Collecting fawltydeps
  Using cached fawltydeps-0.18.0-py3-none-any.whl (64 kB)
[..]
(simplejson) klaas@pop-os:/tmp/simplejson$ fawltydeps 

No undeclared or unused dependencies detected.
```

Being explicit doesn't help:

```
(simplejson) klaas@pop-os:/tmp/simplejson$ fawltydeps --pyenv .

No undeclared or unused dependencies detected.
```

**Expected behavior**

I expect this to work equally well as when using a venv outside of the project-dir, or a deeply nested venv

**Environment**

 - Linux (pop-os 22.04)
 - Version of the code: FawltyDeps v0.18.0

jherland:
Thanks for reporting! To be honest this is the first time I've seen top-level venv, and I was not aware that keeping your source tree "inside" a virtualenv, like this, was even an option.

[...]

PR #468, opened 2025-01-28
Work around top-level venv special case

jherland:
This should solve issues #465.

Add special handling for the case were we have found a Python environment rooted at the same directory where we have been asked to look for code and/or deps.

Commits:
- `tests/sample_projects`: Add `top_level_venv` for issue #465
- `traverse_project`: Better handle overlap between pyenv and code/deps dir


obscurerichard (approved):
👍🏻 :shipit: 

Nice unit test for this too...

PR #470, opened 2025-02-18
Suggest package names for undeclared imports, when possible

jherland:
Based on @layus' idea in https://gist.github.com/layus/bbb1a4c699f3d48ff18f08ec76486558.

Commits:
- Provide package suggestions for undeclared imports
- Various fixups and code movements to make tests/linters pass
- Reorganize fundamental `Package` types


layus (approved):
Seem good to me. I am surprised by the amount of extra work you had to do to adapt my patch. And this would also mandate some changes in the text output to display the information. But not needed to land this.

jherland:
> Seem good to me. I am surprised by the amount of extra work you had to do to adapt my patch. And this would also mandate some changes in the text output to display the information. But not needed to land this.

Yes, the initial refactoring was more than I expected, too, but I found a better approach in 4fdb5ec and bbfc00c that avoids much unnecessary code movement. The overall diff is already quite a bit smaller. 😄 

BTW I'm still refactoring and working on the text output, so I keep the PR in draft mode for now. 😉

layus (approved):
Much cleaner indeed! Thanks again :-)

tests/test_cmdline.py
```diff
@@ -235,12 +241,12 @@ def test_list_imports__from_dir_with_some_excluded__prints_imports_from_unexclud
         {
             "file1.py": """\
                 from my_pathlib import Path
-                import pandas, scipy
+                import my_pandas, scipy
```
layus, line 244: I am a bit curious about this mass-rename. Why was it needed for numpy and pandas but not scipy for example ?

layus (approved):
Still okay to me.
I like that there are a good number of tests for the feature.
One question above, but nothing blocking a merge.

tests/test_cmdline.py
jherland: The Python environment in which the GitHub CI is running is not as clean and pristine as the custom Python environments we set up with Nox when running tests locally. The GitHub CI environment sometimes happen to include pandas and numpy, but scipy has not been a problem (yet). Or, to quote the commit message:

>  The requests module is accidentally available in the Python environment in which these tests are run on GitHub CI. Use some non-existing module names instead of commonly used modules in our examples, to help avoid CI behaving differently from local test runs.

I'm not sure if we can exactly know/predict what is going to be present in the GitHub CI environment, but in any case (and in addition to the workaround in this PR), the proper fix is to _decouple_ ourselves from this environment and establish our own where we have full control.

Looking at `.github/workflows/tests.yaml`, I see we are running `nox` with `--no-venv --no-install` which disables the isolation that we have by default when running locally. I suspect this was done in order for the GHA caching to work as intended.

Anyway, we really should replace this (and all our Poetry use) with `uv`, the result of which should be a significant speedup in how we create/populate virtualenvs in CI, and we should maybe no longer have to rely on GHA caching at all.

PR #471, opened 2025-02-19
Drop Python v3.8 support

robmoss2k:
`isort` recently dropped support for Python v3.8. `FawltyDeps`' dependency on `isort="^5.10"` breaks projects that use `isort="^6.0.0"`. As Python v3.8 is no longer supported, and according to the Boundaries section of CodeDesign, support should be dropped and the `isort` requirement should be updated to `>=5.10`.

Note that the _only_ breaking change in `isort 6.0.0` is dropping Python v3.8 support.


robmoss2k:
Fixes #472.

jherland:
Thanks again for the work put into this PR. We've come around to the idea of dropping Python v3.8 support, and I hope to get this PR merged soon. We might want to make another, smaller release before we drop v3.8 support, but I'll get back to that if and when it happens.

A couple of initial observations before I look closer at this PR:

The second commit that you added earlier today ("Update dependencies") unfortunately rewrites the entire `pyproject.toml` file (did you use some kind of auto-formatter?). For one, it loses all the authors except one, and it's also hard to see what - if any - intentional changes are there in addition to the reformatting. Please drop or redo that commit.

Also, please rebase on current `main` to fix the conflict mentioned by GitHub here.

robmoss2k:
That second commit shouldn't have happened - I'll back that one out, sorry. I have a tool that upgrades the `pyproject.toml` file to get rid of the deprecated fields (mainly the `tool.poetry` section) but it's completely specific to my setup.

PR #475, opened 2025-02-25
Follow-up after removing support for Python v3.8

jherland:
Various cleanups and modernizations made possible by #474.


obscurerichard (approved):
👍🏻 :shipit: 

This all looks pretty normal to me for the types of changes that would be good to have in the wake of a version upgrade... and the tests are passing.

PR #485, opened 2025-04-28
CI: call actionlint to check workflow files

smelc:
## Context

As part of me learning [CodeQL](https://codeql.github.com/), I'm looking at how to run CodeQL Python security rules on FawltyDeps, in its CI. When doing so, I noticed that FawltyDeps' CI had no quality checks, so I'm proposing to add one.

## How to trust this PR

* [actionlint](https://github.com/rhysd/actionlint) is a well-established checker
* There was only one bash error found by actionlint, which I fixed in commit `CI: fix shell warning found by actionlint`, well done team :clap: 
* There was a warning about the `checkout` action version being too old. I fixed it, I think it's not controversial
* There is a warning about the `setup-python` version being too old. I didn't fix it, as I think the changelog should be inspected to understand what updating means and this seemed outside the scope of this PR. So this check is ignored with actionlint's `-ignore` flag.

Finally, you can look at a run of the augmented CI on this branch (where I had added a rule to have it trigger): https://github.com/smelc/FawltyDeps/commits/smelc/ci-add-actionlint/ and https://github.com/smelc/FawltyDeps/actions/runs/14703920052/job/41259362025

### DIFF
--- a/.github/workflows/actionlint.yaml
+++ b/.github/workflows/actionlint.yaml
@@ -15,5 +15,7 @@ jobs:
         run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
         shell: bash
       - name: Check workflow files
-        run: ${{ steps.get_actionlint.outputs.executable }} -color
+        run: |
+          ${{ steps.get_actionlint.outputs.executable }} -color \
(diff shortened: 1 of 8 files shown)


layus (approved):
Looks good to me. This code is not controversial, so it's just a matter of deciding if the feature is needed for this repo.

PR #487, opened 2025-05-07
CI: run CodeQL

smelc:
> [!WARNING]
>
> The second commit needs to be dropped before merging

## Context

[CodeQL](https://codeql.github.com/) is a static analyzer provided by GitHub. Depending on the queries it is called with, it can act both as a code smell detector, as well as a checker for security vulnerabilities via dataflow analyses and taint tracking.

This PR adds CodeQL checks to this repository, to increase our users' confidence that FawltyDeps is secure. I have configured it to run Python's _quality_ rules (good practices) as well as the _security-extended_ (all security rules).

## How to trust this PR

Look at the security warnings found on my fork (where this pipeline has been merged into `main`): https://github.com/smelc/FawltyDeps/security/code-scanning

image

* Observe the findings make sense
* Observe that FawltyDeps is indeed secure, comparing it with a more PoC concept like https://github.com/smelc/nix-security-tracker/security/code-scanning

### Require CodeQL checks to be provided by every PR (doesn't block PRs)(optional)

This is in branch protection:

image

## Make CodeQL block PRs on alerts (optional)

This is not per branch, but repo-wise. This is to block PRs that create new alerts (existing alerts are not considered). Given that this codebase is very secure right now, and by design FawltyDeps doesn't read user inputs interactively (no `input`, no URLs provided by the user), I advice to set this quite high.

To see an example of a PR that cannot be merged because of a new vulnerability, see https://github.com/smelc/nix-security-tracker/pull/3

image

## How vulnerabilities look

Vulnerabilities can be found both on https://github.com/smelc/FawltyDeps/security/code-scanning (see that you can filter by PR) and they appear on a PR's _Files_ tab, like this (from [this PR](https://github.com/smelc/nix-security-tracker/pull/3/files)):

image

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -2,6 +2,9 @@ name: "CodeQL Analysis"
 
 on:
   workflow_call:
+  push:
+    branches: ["smelc/run-code-ql-security-rules"]
+  pull_request:
 
 jobs:
   codeql:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,31 @@
+name: "CodeQL Analysis"
+
+on:
+  workflow_call:
+
+jobs:
+  codeql:
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v3
+      with:
+        languages: "python"
+        source-root: "fawltydeps"
+        queries: security-and-quality
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v3
+      with:
+        category: "/language:python"


jherland (approved):
Happy to merge this as-is (without the 2nd commit), in order to get CodeQL runs  started for FawltyDeps. Once it's running, we can always re-evaluate its usefulness later, but before it's merged, its hard to tell exactly how well it will work for us.

I'm not exactly sure how much real value it will add, given that we're already running with quite an extensive set of `ruff` checks. As far as I gather from your screenshots, it only reports two new problems: the use of `except KeyError: pass` in [`parse_dynamic_pyproject_contents()`](https://github.com/tweag/FawltyDeps/blob/291e2d27446b7f2c9b868b143bce3c1a91f2ef9b/fawltydeps/extract_deps/pyproject_toml_parser.py#L189). For these cases I would actually _disagree_ with CodeQL's complaint, and label them false positives.

Unfortunately, unlike most other linters, it appears there is no standard way to annotate these cases (e.g. with a `# noqa` comment) to silence CodeQL. In fact this remains an open issue with CodeQL, and from the discussion at https://github.com/github/codeql/issues/11427, I'm not getting the impression that the maintainers see this as a high-priority issue. Rather, I'm getting some bad vibes in the direction that CodeQL is as much about locking projects into the Microsoft ecosystem. as it is about improving code quality.

I also briefly looked at running CodeQL locally (in a Nox session) so that we can add this to our growing collection of easy-to-run linters, but it seems not very straightforward at all to set up the environment needed to get the same query running locally, especially not as part of a poetry/uv/pip-managed virtualenv.

In summary, I'm slightly skeptical to this tool's usefulness, but I'm willing to give it a try 😄. 

🚀

smelc:
@jherland> let me highlight, no pressure from me to merge this :wink:

I'm going to look if we can disable the query you disagree with. CodeQL's GH action integration is fine-grained, maybe we can do that. And I'll get back to you here.

jherland:
Yeah, I'm pretty sure it can be easily ignored/disabled in the GitHub UI, at least that's what I gather from https://github.com/github/codeql/issues/11427. It's just strange that it cannot be done _at_ the site of the apparent violation, i.e. in the source itself.

Anyway, I'm happy for you to just drop the second commit from this PR, and then I can maybe manage to get it merged before I go on PTO next week. 😄

smelc:
Here is how the new commit affects the _Security_ tab in my fork: 

image

(the alerts are marked are _Closed_ I presume because they used to exist on the branch's HEAD, but were ultimately removed)

jherland:
Thanks, much appreciated. But I'm not sure how to merge your PR without you first dropping the last commit? I cannot push to your fork, and GitHub does not allow me to merge a partial PR...

PR #488, opened 2025-05-08
RFC/Follow-up to actionlint CI action

jherland:
First, thanks to @smelc for [adding the actionlint action in the first place](https://github.com/tweag/FawltyDeps/pull/485). Although part of this PR appears to undo his work, that is the opposite of what I actually want to accomplish here:

You see, I have this weird hangup on being able to easily run all (or at least most) CI checks locally. Not only does it make it easier to reproduce CI failures in a local environment, but it also decouples us somewhat from the specifics of the hosting/CI platform that we happen to run on, making it easier to potentially migrate away from GitHub in the future for whatever reason.

This PR uses the [actionlint-py](https://pypi.org/project/actionlint-py/) shim to provide the actionlint executable inside our local development environment, as well as adding it to our `lint` Nox action (i.e. te be run together with our other linters via `nox -s lint`).

Is this maybe one step too far? After all, the files we are linting here are very much GitHub-specific, and you could claim that if the local development environment is to remain independent  of GitHub, then it's stupid to add an action whose 
raison d'être very much depends on Github... 🤔


smelc (approved):
> I have this weird hangup on being able to easily run all (or at least most) CI checks locally

Agreed that running the CI locally can boost developer productivity :+1: 

I tested running `actionlint` locally with `nox -s lint` and confirms that it works and reports errors :+1: 

:warning: I didn't check the changes related to installing `python` and `poetry`

jherland:
My objective with the last commit on this PR is to decrease overhead in CI somewhat by using fewer actions:
- replacing the explicit caching action with setup-python's built-in Poetry cache
- Using `pipx install poetry` as prescribed on https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages rather than the separate [Gr1N/setup-poetry](https://github.com/Gr1N/setup-poetry) action.

Does it work? I think so. Let's look at a the relatively quick action, whose runtime should be dominated by the setup/caching overhead: **format (3.11, ubuntu-latest)**

- Run from the `main` branch (cache found) took 25s: [Link](https://github.com/tweag/FawltyDeps/actions/runs/14860575031/job/41724134791)
  Screenshot_20250509_113250

- Run from this PR (cache not found) took 18s: [Link](https://github.com/tweag/FawltyDeps/actions/runs/14912783893/job/41891026699)
  Screenshot_20250509_112338

- Run from this PR (cache found) took 15s: [Link](https://github.com/tweag/FawltyDeps/actions/runs/14925875989/job/41930519514)
  Screenshot_20250509_113105

This is one sample, not statistically significant, and there is huge variation between runs (and platforms!), but quickly browsing a few builds before/after this change suggests that `actions/setup-python` + `Gr1N/setup-poetry` + `actions/cache` is spending ~5s more per run than `pipx install poetry` + `actions/setup-python` w/caching.

obscurerichard (approved):
👍🏻 :shipit:

PR #489, opened 2025-05-21
Follow-up to #484: Add a --base-dir option

### LINKED ISSUE #483: Consider adding a base_dir flag

I use fawltydeps in a weird way, because I want to check only some files in a project folder. For example:

```
fawltydeps project_root/ --code project_root/foo/bar.py --code project_root/foo/some/where/else.py
```

In that setup, it is unclear what fawltydeps should do according to the docs at https://github.com/tweag/FawltyDeps/blob/main/docs/usage.md. The best I could understand is that project_root is both --code and --deps. As --code is overriden, then project_root is only --deps. (which is what I want, I want dependencies from my project to be resolved).

What is missing though is a way to tell fawltydeps that the base path of project_root/foo/bar.py and project_root/some/where/else.py is `project_root/foo` so as to resolve relative imports correctly.

I created a patch adding a new `--base_path` option and pass it in traverse_project as the second argument to validate_code_source.

https://github.com/tweag/FawltyDeps/blob/034d6054b169a1a96ccf14f12e20cb4d6045e3e5/fawltydeps/traverse_project.py#L106-L108

I wonder if the same could be achieved with changing current directory first or not.

**Additional context**

@jherland mentioned that passing a path on top of --code arguments has some side effects.

Please, blame bazel for the weird python project layout.

jherland:
This was intended as an update to #484, but since I don't have push access to that branch in @layus' fork, I'm opening a new PR here instead.

This version is rebased on top of origin/main, and adds three more commits implementing the changes I suggested in https://github.com/tweag/FawltyDeps/pull/484#issuecomment-2858157022:
- Rename the internal `basepaths` name to `search_paths`, to reduce (my own) confusion.
- Rename the `--base-path` option to `--base-dir` to make it clear that is only accepts directories.
- Make sure the `--base-dir` option -- when given -- also applies during directory traversal, as I believe this is what an average user would expect. See commit 44cc39cb3d5b86ef73a8147411e53db5710830d3 for a fuller explanation, including another `sample_projects` test case.

Closes #483.


fawltydeps/cli_parser.py
```diff
@@ -204,6 +204,16 @@ def populate_parser_paths_options(parser: argparse._ActionsContainer) -> None:
...
         ),
     )
+    parser.add_argument(
+        "--base-dir",
+        type=Path,
+        metavar="BASE_DIR",
+        help=(
+            "Directory used as base for determining 1st-party imports, i.e."
+            " where your project expects to find its own packages/modules."
+            " Defaults to any directory given to --code (or search_paths)."
```
layus, line 214: The subtlety is that `--base-dir` applies to all --code/search_paths, while the default is for each --code/search_path to be its own base_dir (when it is a dir).


```suggestion
            "Directory used as base for determining 1st-party imports, i.e."
            " where your project expects to find its own packages/modules."
            " If not provided, each directory given to --code (or search_paths)"
            " acts as the base directory for all its children; files will default"
            " to use the current working directory instead."
```

fawltydeps/settings.py
```diff
@@ -236,17 +237,17 @@ def create(cls, cmdline_args: argparse.Namespace) -> Settings:
...
+            if search_paths not in (code_paths, deps_paths, pyenv_paths):
                 msg = (
-                    "All four path specifications (code, deps, pyenvs, and base)"
-                    f"have been used. Use at most 3. basepaths={base_paths}, "
-                    f"code_paths={code_paths}, deps_paths={deps_paths}, "
-                    f"pyenv_paths={pyenv_paths}"
+                    "All four path specifications (code, deps, pyenvs, and"
+                    "search_paths) have been used. Use at most 3."
+                    f"search_paths={search_paths}, code_paths={code_paths}, "
+                    f"deps_paths={deps_paths}, pyenv_paths={pyenv_paths}"
```
layus, line 250: Not to be addressed as part of this PR, but this limitation confuses me.
I think a warning would be better here, saying that `search_paths` will be ignored because code_paths, deps_paths and pyenv_paths are already populated.
Or just an expanded explanation about search_paths being used as a default for all these values.
Not quite sure what rn.

tests/test_sample_projects.py
```diff
@@ -87,6 +93,7 @@ def build_settings(self, cache: pytest.Cache) -> Settings:
             pyenvs = {Path(path) for path in self.pyenvs}
         return Settings(
             actions={Action.REPORT_UNDECLARED, Action.REPORT_UNUSED},
+            base_dir=None if self.base_dir is None else Path(self.base_dir),
```
layus, line 96: I miss `Path <$> self.base_dir` ;-)

layus (approved):
search_paths is an amazing name, it helps a lot to understand what happens.
I did not do the full search to see if you missed a rename somewhere in comments. I will trust the test suite.
I struggled a bit with the wording, and did my best to improve it in user-facing documentation.
None of my comments are hard blockers though.

Thank you for improving quite a lot on the initial PoC :-)

fawltydeps/cli_parser.py
jherland: Yeah, I think this subtlety is quite hard to convey in a terse help message on the command line. I'm almost tempted to make leave as it is (short and incomplete), and rather document it more thoroughly in a [FAQ](https://tweag.github.io/FawltyDeps/FAQ/) called "What's the deal with `--base-dir`" or something. 😅

I'll have to think about this a little...

fawltydeps/settings.py
jherland: Yeah, this code is a bit clumsy. I think I agree with you that a warning is better here. The fix is relatively quick, so I'll just add  it to this PR.

tests/sample_projects/base_dir_testbed/subproject/lib/foo.py
```diff
@@ -0,0 +1,5 @@
+import foo
+
+
```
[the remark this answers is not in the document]
jherland: It's tangential to the purpose of this test, but I think you are right. There's no need to mix that isort issue into this test: we actually do not need an `import` statement inside `lib/foo.py` at all, given that what we're testing is that `lib/foo.py` _itself_ is found to be a first-party import. 👍🏻

fawltydeps/settings.py
jherland: Fixed in [220e802](https://github.com/tweag/FawltyDeps/pull/489/commits/220e8023e0e956a51f53ea559fcdb71ca597d84c).

fawltydeps/cli_parser.py
jherland: I added more documentation in [4b17802](https://github.com/tweag/FawltyDeps/pull/489/commits/4b17802e8af9227f12dca9b36f0241c6272a7ccd), opting to keep the CLI help text a little bit shorter.

PR #495, opened 2025-06-11
environment.yml: Ignore editable self-dependencies

### LINKED ISSUE #494: environment.yml with pip and -e .
**Describe the bug**
pip with -e ./ in an environment.yml file causes:

```
ERROR:fawltydeps.extract_deps.environment_yml_parser:Failed to parse Pip dependencies in environment.yml: Expected package name at the start of dependency specifier
    -e ./
    ^

```

**To Reproduce**

```
name: myenv
channels:
  - conda-forge
dependencies:
  # Core Python and conda packages
  - python=3.12
  - pip
  - setuptools

  # Pip packages
  - pip:
    - -e ./
```

**Expected behavior**
Either just ignore or not throw ERROR

**Environment**
- using micromamba in linux runner for CI

jherland:
Issue #494 highlights the use of "-e ./" (i.e. an editable
self-dependency) in environment.yml files.

We already work around these kinds of self-dependencies in
pyproject.toml and pixi.toml files, and our requirements.txt parser
automatically ignores any lines with pip options, hence environment.yml
is the only(?) remaining dependency file format where we're not yet
handling this.

Fix that, by recognizing this case before we emit an unnecessary error
message about it. In either case we continue parsing the remaining
lines, so the actual behavior (except for the error message) does not
need to change.

Fixes #494.

### DIFF
--- a/fawltydeps/extract_deps/environment_yml_parser.py
+++ b/fawltydeps/extract_deps/environment_yml_parser.py
@@ -64,7 +64,15 @@ def parse_environment_yml_deps(
                 yield parse_item(dep_item, source)
             except ValueError as e:
                 # InvalidCondaRequirement or packaging.requirements.InvalidRequirement
-                logger.error(f"{error_msg} {e}")
+                # Skip editable self-dependencies silently:
+                if dep_item.removesuffix("/") in {
+                    "-e .",
+                    "--editable .",
+                    "--editable=.",
+                }:
+                    pass
+                else:
+                    logger.error(f"{error_msg} {e}")
         elif isinstance(dep_item, dict) and len(dep_item) == 1 and "pip" in dep_item:
             pip_deps = dep_item.get("pip")
             yield from parse_environment_yml_deps(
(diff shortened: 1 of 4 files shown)


fawltydeps/extract_deps/environment_yml_parser.py
```diff
@@ -64,7 +64,15 @@ def parse_environment_yml_deps(
                 yield parse_item(dep_item, source)
             except ValueError as e:
                 # InvalidCondaRequirement or packaging.requirements.InvalidRequirement
-                logger.error(f"{error_msg} {e}")
+                # Skip editable self-dependencies silently:
+                if dep_item.removesuffix("/") in {
+                    "-e .",
+                    "--editable .",
+                    "--editable=.",
```
layus, line 71: Is `.` the only value to skip here? Could there be other values that need to be skipped, like `nested/lib` or such ?

layus (approved):
The change makes sense, the code too. Ready to ship as-is.

fawltydeps/extract_deps/environment_yml_parser.py
jherland: For now, I think we only want to work around the special case of direct (editable) self-dependency.

AFAICS, this is by far the most common case of this pattern, and I'm not 100% sure if a nested path dependency (editable or not) can be safely ignored...

jherland: ...like, in your `nested/liib` example: Presumably this dependency declaration would make `import lib` available in your Python environment, and I suspect it _should_ be treated by FD as a 3rd-party import?

